PR #3, opened 2023-07-06
use_snake_case_for_server_api

juli1:

The server API uses `camelCase` for all attributes when it should returns `snake_case`. The serde serialization for the node is using `camelCase`, which is what we expect in rules for JavaScript code. We need to have the conversion in the server to use `snake_case`.


juli1:
@raczosala please tell me if this fixes the issue with the server.

PR #14, opened 2023-07-08
Timeout fix wihout barrier

juli1:

Same problem than [this PR](https://github.com/DataDog/datadog-static-analyzer/pull/13) but without a barrier. The real problem is that we added more complexity rather than doing the correct fix.

In the fix in [this PR](https://github.com/DataDog/datadog-static-analyzer/pull/13), we add a barrier to synchronize the threads. The real issue is that the created thread was [notifying the condition variable](https://github.com/DataDog/datadog-static-analyzer/blob/ffce75f6a1c387b7a0e6f27089eb4463b3febb37/kernel/src/analysis/javascript.rs#L213) before the main thread could [wait on it](https://github.com/DataDog/datadog-static-analyzer/blob/ffce75f6a1c387b7a0e6f27089eb4463b3febb37/kernel/src/analysis/javascript.rs#L226). The fix added a barrier but the real issue is that the mutex of the condition variable should have been used.

## What is your solution?

The solution is to lock the main mutex before even starting the thread ([see here](https://github.com/DataDog/datadog-static-analyzer/pull/14/files#diff-3ef73571d7322cf3fd57c413d4333c816bb235110e2a1474377e3c122c03c8abR175-R177)). This way, the thread will be blocked when trying to [acquire the mutex in the thread](https://github.com/DataDog/datadog-static-analyzer/pull/14/files#diff-3ef73571d7322cf3fd57c413d4333c816bb235110e2a1474377e3c122c03c8abR212).

It will guarantee that [waiting on the condition variable](https://github.com/DataDog/datadog-static-analyzer/pull/14/files#diff-3ef73571d7322cf3fd57c413d4333c816bb235110e2a1474377e3c122c03c8abR221) occurs before [notifying the condition variable](https://github.com/DataDog/datadog-static-analyzer/pull/14/files#diff-3ef73571d7322cf3fd57c413d4333c816bb235110e2a1474377e3c122c03c8abR214).

1. To ensure better quality, all commits are also tested on Windows, MacOS and Linux. This would ensure all our changes work across all platforms this analyzer will support.
2. One bug has been fixed about filtering paths for Windows.

### DIFF
--- a/bins/src/bin/datadog-static-analyzer.rs
+++ b/bins/src/bin/datadog-static-analyzer.rs
@@ -188,7 +188,7 @@ fn main() -> Result<()> {
 
     let languages = get_languages_for_rules(&rules);
 
-    let files_to_analyze = get_files(directory_to_analyze.as_str(), ignore_paths.clone().as_ref())
+    let files_to_analyze = get_files(directory_to_analyze.as_str(), ignore_paths.clone())
         .expect("unable to get the list of files to analyze");
 
     print_configuration(
--- a/bins/src/bin/datadog-static-analyzer.rs
+++ b/bins/src/bin/datadog-static-analyzer.rs
@@ -211,7 +211,7 @@ fn main() -> Result<()> {
 
     // we always keep one thread free and some room for the management threads that monitor
     // the rule execution.
-    let ideal_threads = ((num_cpus::get() as f32 - 1.0) * 0.80) as usize;
+    let ideal_threads = ((num_cpus::get() as f32 - 1.0) * 0.90) as usize;
     let num_threads = if ideal_threads == 0 { 1 } else { ideal_threads };
     rayon::ThreadPoolBuilder::new()
         .num_threads(num_threads)
--- a/bins/src/bin/datadog-static-analyzer.rs
+++ b/bins/src/bin/datadog-static-analyzer.rs
@@ -188,7 +188,7 @@ fn main() -> Result<()> {
 
     let languages = get_languages_for_rules(&rules);
 
-    let files_to_analyze = get_files(directory_to_analyze.as_str(), &ignore_paths.clone())
+    let files_to_analyze = get_files(directory_to_analyze.as_str(), &ignore_paths)
         .expect("unable to get the list of files to analyze");
 
     print_configuration(
--- a/bins/src/bin/datadog-static-analyzer.rs
+++ b/bins/src/bin/datadog-static-analyzer.rs
@@ -188,7 +188,7 @@ fn main() -> Result<()> {
 
     let languages = get_languages_for_rules(&rules);
 
-    let files_to_analyze = get_files(directory_to_analyze.as_str(), ignore_paths.clone())
+    let files_to_analyze = get_files(directory_to_analyze.as_str(), &ignore_paths.clone())
         .expect("unable to get the list of files to analyze");
 
     print_configuration(
--- a/cli/src/file_utils.rs
+++ b/cli/src/file_utils.rs
@@ -24,21 +24,11 @@ fn get_extensions_for_language(language: &Language) -> Option<Vec<String>> {
 
 // get the files to analyze from the directory. This function walks the directory
 // to analyze recursively and gets all the files.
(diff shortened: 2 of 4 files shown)


cli/src/file_utils.rs
```diff
@@ -42,9 +42,15 @@ pub fn get_files(directory: &str, paths_to_ignore: &[String]) -> Result<Vec<Path
 
         // check if the path should be ignored by a glob or not.
         for path_to_ignore in paths_to_ignore {
-            // we build the expanded glob to make sure it will match with the full path.
-            let expanded_glob = format!("{}/{}", directory, path_to_ignore);
-            if glob_match(expanded_glob.as_str(), entry.display().to_string().as_str()) {
+            let path_buf = entry.to_path_buf();
```
robertohuertasm, line 45: Why not introducing a short-circuit if `should_include` is already `false` in ln.41:
```rust
 if !should_include {
    continue;
}
```

robertohuertasm, line 45: We could probably move this line out the loop to avoid multiple evaluation.

cli/src/file_utils.rs
```diff
@@ -42,9 +42,15 @@ pub fn get_files(directory: &str, paths_to_ignore: &[String]) -> Result<Vec<Path
...
+
+            let relative_path = path_buf
+                .as_path()
+                .strip_prefix(directory)
+                .unwrap()
+                .to_str()
+                .expect("should get the path");
```
robertohuertasm, line 52: What about this? Removing the unwraps and using a result instead to avoid panics.
```suggestion
            let relative_path = path_buf
                .strip_prefix(directory)
                .ok()
                .and_then(|p| p.to_str())
                .ok_or_else(|| anyhow::Error::msg("should get the path"))?;
```

bins/src/bin/datadog-static-analyzer.rs
```diff
@@ -188,7 +188,7 @@ fn main() -> Result<()> {
 
     let languages = get_languages_for_rules(&rules);
 
-    let files_to_analyze = get_files(directory_to_analyze.as_str(), ignore_paths.clone().as_ref())
+    let files_to_analyze = get_files(directory_to_analyze.as_str(), &ignore_paths.clone())
```
robertohuertasm, line 191: I think you can avoid the `clone`.

robertohuertasm (commented):
Some comments for your consideration.

kernel/src/analysis/javascript.rs
```diff
@@ -215,12 +216,6 @@ pub fn execute_rule(
 
     yield_now();
```
robertohuertasm, line 217: Is this really needed?

robertohuertasm: Also, not related to this line but to line 243 (cannot put a comment there), why the explicit `drop`? The `JoinHandle` is going to be dropped anyway.

cli/src/file_utils.rs
juli1: The reason is because I want to favor simplicity/readability over performance, especially for this part that is here only in the beginning.

kernel/src/analysis/javascript.rs
juli1: This was something that was related to the previous code. Removing it, thank you for flagging it :)

bins/src/bin/datadog-static-analyzer.rs
juli1: Removed, thank you for flagging it :)

cli/src/file_utils.rs
robertohuertasm: In my opinion, adding the short-circuit here doesn't hurt readability giving that it's just 3 lines just after `should_include` definition. On the contrary, it makes clear the intent. At least for me, not having that short-circuit make the code a little bit less intuitive. I had to inspect the whole block to understand the motivation because I found odd not having the short-circuit in there in the first place. 

That being said, it's just a suggestion and some of these things are subjective. I added the comment because of the reasons exposed before (totally subjective) and for performance (small in this case).

dastrong (approved):

PR #15, opened 2023-07-08
chore(clippy): apply idiomatic changes

robertohuertasm:

There are some changes that could be done to improve readability in the code, avoid allocations or simplify the code.

## What is your solution?

By following some of the Clippy recommendations I decided to apply some refactors to the codebase. I'll try to explain the rationale of each change in the code itself by leaving some messages as a mode of auto-review.

No problem if some change is deemed as subjective and you prefer to dismiss it.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,6 +4,7 @@ members = [
     "bins",
     "cli",
     "kernel",
+    "server"
 ]
 
 [profile.release]
--- a/kernel/build.rs
+++ b/kernel/build.rs
@@ -6,11 +6,12 @@ where
     F: FnMut(&mut Command) -> &mut Command,
 {
     let mut command = Command::new(name);
-    println!("Running {:?}", command);
+    println!("Running {command:?}");
     let configured = configure(&mut command);
-    if configured.status().is_err() {
-        panic!("failed to execute {:?}", configured);
-    }
+    assert!(
+        configured.status().is_ok(),
+        "failed to execute {configured:?}"
+    );
 }
 
 fn main() {
--- a/kernel/src/analysis/analyze.rs
+++ b/kernel/src/analysis/analyze.rs
@@ -47,70 +47,72 @@ pub fn analyze(
 ) -> Vec<RuleResult> {
     let lines_to_ignore = get_lines_to_ignore(code, language);
 
-    match get_tree(code, language) {
-        Some(tree) => rules
-            .into_iter()
-            .map(|rule| {
-                let invalid_query_result = RuleResult {
-                    rule_name: rule.name.clone(),
-                    filename: filename.to_string(),
-                    violations: vec![],
-                    errors: vec![ERROR_INVALID_QUERY.to_string()],
-                    execution_error: None,
-                    execution_time_ms: 0,
-                    output: None,
-                };
-
-                if let Some(tree_sitter_query) = &rule.tree_sitter_query {
-                    let query_try = get_query(tree_sitter_query.as_str(), &rule.language);
-
-                    match query_try {
-                        Ok(query) => {
-                            let nodes =
-                                get_query_nodes(&tree, &query, filename, code, &HashMap::new());
-
-                            if nodes.is_empty() {
-                                RuleResult {
-                                    rule_name: rule.name.clone(),
-                                    filename: filename.to_string(),
-                                    violations: vec![],
-                                    errors: vec![],
-                                    execution_error: None,
-                                    execution_time_ms: 0,
-                                    output: None,
-                                }
-                            } else {
-                                let mut rule_result = execute_rule(
-                                    rule,
-                                    nodes,
-                                    filename.to_string(),
-                                    analysis_option.clone(),
-                                );
-
-                                // filter violations that have been ignored
-                                rule_result.violations = rule_result
-                                    .violations
-                                    .iter()
-                                    .cloned()
-                                    .filter(|v| !lines_to_ignore.contains(&v.start.line))
-                                    .collect();
-                                rule_result
-                            }
-                        }
-                        Err(_) => invalid_query_result,
-                    }
-                } else {
-                    invalid_query_result
-                }
-            })
-            .collect(),
-        None => {
+    get_tree(code, language).map_or_else(
+        || {
             if analysis_option.use_debug {
-                eprintln!("error when parsing source file {}", filename);
+                eprintln!("error when parsing source file {filename}");
             }
             vec![]
-        }
-    }
+        },
+        |tree| {
+            rules
+                .into_iter()
+                .map(|rule| {
+                    let invalid_query_result = RuleResult {
+                        rule_name: rule.name.clone(),
+                        filename: filename.to_string(),
+                        violations: vec![],
+                        errors: vec![ERROR_INVALID_QUERY.to_string()],
+                        execution_error: None,
+                        execution_time_ms: 0,
+                        output: None,
+                    };
+
+                    if let Some(tree_sitter_query) = &rule.tree_sitter_query {
+                        let query_try = get_query(tree_sitter_query.as_str(), &rule.language);
+
+                        match query_try {
+                            Ok(query) => {
+                                let nodes =
+                                    get_query_nodes(&tree, &query, filename, code, &HashMap::new());
+
+                                if nodes.is_empty() {
+                                    RuleResult {
+                                        rule_name: rule.name.clone(),
+                                        filename: filename.to_string(),
+                                        violations: vec![],
+                                        errors: vec![],
+                                        execution_error: None,
+                                        execution_time_ms: 0,
+                                        output: None,
+                                    }
+                                } else {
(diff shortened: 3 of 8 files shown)


juli1:
Thanks for sending a pull request!

First, as for the previous pull request, is it possible to follow the template and explain the problem?

Second, this PR changes a lot of files with some changes being subjective (e.g. using `println!("Running {:?}", command);` or `println!("Running {command:?}");`). Our objective is to have pull request that focuses on a clear problem at hand and avoid too many small changes that are subjective and not a core issue.

In short, is it possible to explain the problem this PR solves (or the feature it brings) and focuses on this?

Thanks!

Cargo.toml
```diff
@@ -4,6 +4,7 @@ members = [
     "bins",
     "cli",
     "kernel",
+    "server"
```
robertohuertasm, line 7: Explicitly adding `server` to the workspace. It's already included via path dependency but for clarity's sake I think it was worth it to introduce it here.

kernel/build.rs
```diff
@@ -6,11 +6,12 @@ where
     F: FnMut(&mut Command) -> &mut Command,
 {
     let mut command = Command::new(name);
-    println!("Running {:?}", command);
+    println!("Running {command:?}");
```
robertohuertasm, line 9: See [clippy rule](https://rust-lang.github.io/rust-clippy/master/index.html#/uninlined_format_args). Readability change. Applies to all the changes like this.

kernel/build.rs
```diff
@@ -6,11 +6,12 @@ where
...
     let configured = configure(&mut command);
-    if configured.status().is_err() {
-        panic!("failed to execute {:?}", configured);
-    }
+    assert!(
```
robertohuertasm, line 11: Readability change. See this [clippy lint](https://rust-lang.github.io/rust-clippy/master/index.html#/manual_assert). Applies to all if-panic in the codebase.

kernel/src/analysis/analyze.rs
```diff
@@ -47,70 +47,72 @@ pub fn analyze(
...
-                let invalid_query_result = RuleResult {
-                    rule_name: rule.name.clone(),
-                    filename: filename.to_string(),
-                    violations: vec![],
-                    errors: vec![ERROR_INVALID_QUERY.to_string()],
-                    execution_error: None,
-                    execution_time_ms: 0,
-                    output: None,
-                };
+    get_tree(code, language).map_or_else(
```
robertohuertasm, line 50: This one is possibly the most subjective one although it's also based on some clippy lints like [this](https://rust-lang.github.io/rust-clippy/master/index.html#/option_if_let_else) and [this one](https://rust-lang.github.io/rust-clippy/master/index.html#/option_if_let_else).

kernel/src/analysis/javascript.rs
```diff
@@ -133,7 +133,7 @@ lazy_static! {
     }
         "#,
         );
-        let mut rt = JsRuntimeForSnapshot::new(Default::default(), Default::default());
+        let mut rt = JsRuntimeForSnapshot::new(RuntimeOptions::default(), Default::default());
```
robertohuertasm, line 136: Readability change. See this [clippy lint](https://rust-lang.github.io/rust-clippy/master/index.html#/default_trait_access).

kernel/src/analysis/javascript.rs
```diff
@@ -418,10 +419,12 @@ res
                     rule.name, filename, e
                 );
             }
-            let error_message = match e.to_string().find("at rule_code") {
-                Some(pos) => e.to_string()[..pos].to_string(),
-                None => e.to_string(),
-            };
+
+            let err_str = e.to_string();
+            let error_message = err_str
```
robertohuertasm, line 424: Readability and conciseness change. See this [clippy lint](https://rust-lang.github.io/rust-clippy/master/index.html#/option_if_let_else).

kernel/src/analysis/tree_sitter.rs
```diff
@@ -1,10 +1,10 @@
 use crate::model::analysis::{MatchNode, MatchNodeContext, TreeSitterNode};
 use crate::model::common::{Language, Position};
-use anyhow::{anyhow, Result};
+use anyhow::Result;
 use std::collections::HashMap;
 use tree_sitter::QueryCursor;
 
-fn get_tree_sitter_language(language: &Language) -> Option<tree_sitter::Language> {
+fn get_tree_sitter_language(language: &Language) -> tree_sitter::Language {
```
robertohuertasm, line 7: Here, there's no need to return the `Option` at all, given that the code is always going to return a `Some` variant. This is introducing unnecessary complexity.

robertohuertasm: More context [here](https://rust-lang.github.io/rust-clippy/master/index.html#/unnecessary_wraps)

kernel/src/analysis/tree_sitter.rs
```diff
@@ -129,12 +124,14 @@ pub fn map_node(node: tree_sitter::Node) -> Option<TreeSitterNode> {
                 line: u32::try_from(cursor.node().range().end_point.row + 1).unwrap(),
                 col: u32::try_from(cursor.node().range().end_point.column + 1).unwrap(),
             },
-            field_name: cursor.field_name().map(|v| v.to_string()),
+            field_name: cursor.field_name().map(ToString::to_string),
```
robertohuertasm, line 127: Typical [refactor](https://rust-lang.github.io/rust-clippy/master/index.html#/redundant_closure_for_method_calls) to avoid unnecessary closures.

kernel/src/analysis/tree_sitter.rs
```diff
@@ -129,12 +124,14 @@ pub fn map_node(node: tree_sitter::Node) -> Option<TreeSitterNode> {
...
             children,
         };
 
         Some(ts_node)
     }
+
+    let mut ts_cursor = node.walk();
```
robertohuertasm, line 134: Move this line after the internal function for readability's sake. See this [clippy lint](https://rust-lang.github.io/rust-clippy/master/index.html#/items_after_statements).

kernel/src/analysis/tree_sitter.rs
```diff
@@ -84,7 +81,7 @@ pub fn get_query_nodes(
                 filename: filename.to_string(),
                 variables: variables.clone(),
             },
-        })
+        });
```
robertohuertasm, line 84: Added the semicolon given that we're not returning. See [this](https://rust-lang.github.io/rust-clippy/master/index.html#/semicolon_if_nothing_returned).

kernel/src/model/rule.rs
```diff
@@ -162,7 +162,7 @@ impl Rule {
         }
         let description = self
             .decode_description()
-            .unwrap_or(Some("invalid description".to_string()));
+            .unwrap_or_else(|_| Some("invalid description".to_string()));
```
robertohuertasm, line 165: Refactor to avoid allocation in case the unwrap goes well. See this [clippy lint](https://rust-lang.github.io/rust-clippy/master/index.html#/or_fun_call).

kernel/src/model/common.rs
```diff
@@ -2,7 +2,7 @@ use derive_builder::Builder;
 use serde::{Deserialize, Serialize};
 use std::fmt;
 
-#[derive(Deserialize, Debug, Serialize, PartialEq)]
+#[derive(Deserialize, Debug, Serialize, Eq, PartialEq)]
```
robertohuertasm, line 5: Added `Eq` trait as per this [clippy lint](https://rust-lang.github.io/rust-clippy/master/index.html#/derive_partial_eq_without_eq).

kernel/src/model/common.rs
```diff
@@ -11,10 +11,10 @@ pub enum OutputFormat {
 impl fmt::Display for OutputFormat {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         let s = match self {
-            OutputFormat::Json => "JSON",
-            OutputFormat::Sarif => "SARIF",
+            Self::Json => "JSON",
```
robertohuertasm, line 14: It's usually a good practice to use `Self` in these scenarios. See  [this](https://rust-lang.github.io/rust-clippy/master/index.html#/use_self).

robertohuertasm:
@juli1 I've commented all the changes so you can judge on each of them. There's a rationale attached to each of them most of the time based on readability and common practices.

kernel/src/analysis/analyze.rs
juli1: I agree with you, this one does not really make the code any better. I would tend to not include this change as its quite hard to read. An alternative would be to change the mapping code and divide it into functions to make it more digestible.

juli1 (approved):
Thanks for taking the time to add explanation to the changes! This looks good to me. One caveat on the change in `analyze` as it makes the code slightly harder to read but we can ship this and refactor later.

PR #16, opened 2023-07-08
chore(js): use separate file for js

robertohuertasm:

I saw there was a huge chunk of JS code in the `javascript.rs` file. Having it embedded makes it difficult to maintain it and it's error prone to do modifications on it if needed.

## What is your solution?

Use [include_str! macro](https://doc.rust-lang.org/std/macro.include_str.html) to include this code from a file at **compilation time**.

This basically allows the contributors to edit that file in the IDE and have syntax support, catching potential errors.


kernel/src/analysis/javascript.rs
```diff
@@ -17,122 +17,7 @@ const JAVASCRIPT_EXECUTION_TIMEOUT_MS: u64 = 5000;
...
-        return code.substring(startChar, endChar);
-    };
-
-    // helper function getCodeForNode
-    function getCodeForNode(node, code) {
-        return getCode(node.start, node.end, code);
-    }
-        "#,
-        );
+        let code: FastString = FastString::from_static(include_str!("./js/stella.js"));
```
juli1, line 20: Does this work also for Windows? Are the paths with `/` supported on windows or will the build fail?

juli1 (commented):
One pending question. We may want to ship [this change](https://github.com/DataDog/datadog-static-analyzer/pull/14) first that will enable and test the build on Windows. If the build passes, no reason for not shipping it.

kernel/src/analysis/javascript.rs
juli1: Can you rebase this on the current `main`? Test for windows will then execute and we can test that the change works for Windows.

robertohuertasm: Rebased

robertohuertasm: Good point about the path. I'm so used to work with Unix systems these days that I haven't thought of that. Will test it.

robertohuertasm: According to [this](https://users.rust-lang.org/t/why-does-include-str-act-differently-on-different-oss/43314) it should work across OSs but I'll try to double check that anyways.

robertohuertasm: Double-checked! It works in Windows!

juli1: Nice! Thank you for checking!

juli1 (approved):

PR #17, opened 2023-07-09
feat(version): add version information to some binaries

robertohuertasm:

1. There are two different places in the the `kernel` crate that were exposing a VERSION const:
   1. [utils.rs](https://github.com/robertohuertasm/datadog-static-analyzer/blob/ffce75f6a1c387b7a0e6f27089eb4463b3febb37/kernel/src/utils.rs#L4)
   2. [constants.rs](https://github.com/robertohuertasm/datadog-static-analyzer/blob/ffce75f6a1c387b7a0e6f27089eb4463b3febb37/kernel/src/constants.rs#L1)
 1. Some crates don't expose the version, so it may be difficult to determine if we have the latest build.

## What is your solution?

1. Remove the code in [utils](https://github.com/robertohuertasm/datadog-static-analyzer/blob/ffce75f6a1c387b7a0e6f27089eb4463b3febb37/kernel/src/utils.rs#L4) as it seems that it's not being used at all.
2. Add a `--version` flag for some CLIs so that consumers can understand which version are they using.

## Alternatives considered

I thought of using the `kernel` Cargo.toml version to expose the version as a constant but then @juli1 pointed out the current release system which uses the commit_sha as the version, so I rollback that change.

### DIFF
--- a/kernel/src/constants.rs
+++ b/kernel/src/constants.rs
@@ -1,1 +1,2 @@
-pub const VERSION: &str = "development";
+// pub const VERSION: &str = "development";
+pub const VERSION: &str = env!("CARGO_PKG_VERSION");
--- a/kernel/src/constants.rs
+++ b/kernel/src/constants.rs
@@ -1,2 +1,1 @@
-// pub const VERSION: &str = "development";
-pub const VERSION: &str = env!("CARGO_PKG_VERSION");
+pub const VERSION: &str = "development";
--- a/bins/src/bin/datadog-export-rulesets.rs
+++ b/bins/src/bin/datadog-export-rulesets.rs
@@ -1,4 +1,5 @@
 use cli::datadog_utils::get_ruleset;
+use kernel::constants::VERSION;
 use kernel::model::ruleset::RuleSet;
 
 use getopts::Options;
@@ -19,6 +20,7 @@ fn main() {
     opts.optopt("o", "output", "output file", "rulesets.json");
     opts.optmulti("r", "ruleset", "ruleset to fetch", "python-security");
     opts.optflag("h", "help", "print this help");
+    opts.optflag("v", "version", "shows the version");
 
     let matches = match opts.parse(&args[1..]) {
         Ok(m) => m,
@@ -26,6 +28,11 @@ fn main() {
             panic!("error when parsing arguments: {}", f)
         }
     };
+
+    if matches.opt_present("v") {
+        println!("{}", VERSION);
+        exit(1);
+    }
 
     if matches.opt_present("h") {
         print_usage(&program, opts);
--- a/bins/src/bin/datadog-static-analyzer.rs
+++ b/bins/src/bin/datadog-static-analyzer.rs
@@ -5,6 +5,7 @@ use cli::model::config_file::ConfigFile;
 use cli::rule_utils::{get_languages_for_rules, get_rulesets_from_file};
 use cli::sarif_utils::generate_sarif_report;
 use kernel::analysis::analyze::analyze;
+use kernel::constants::VERSION;
 use kernel::model::analysis::AnalysisOptions;
 use kernel::model::common::Language;
 use kernel::model::rule::{Rule, RuleInternal};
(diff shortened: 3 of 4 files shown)


robertohuertasm:
btw, would you consider a PR moving from `getops` to [clap](https://docs.rs/clap/latest/clap/)? Now that they have incorporated the work of [structopt](https://docs.rs/structopt/latest/structopt/) I feel that it offers a better tooling for CLIs.

robertohuertasm:
btw, would you consider a PR moving from `getops` to [clap](https://docs.rs/clap/latest/clap/)? Now that they have incorporated the work of [structopt](https://docs.rs/structopt/latest/structopt/) I feel that it offers a better tooling for CLIs.

juli1:
Thanks for submitting a pull request

1. Is it possible to follow the pull request template, explain the problem at hand and the solution? Thanks!
2. This break our internal and and [public build](https://github.com/DataDog/datadog-static-analyzer/blob/main/.github/workflows/release.yml). 

I want to understand the clear value of this. I am not opposed to a change but I want to understand (1) why we are doing a change and (2) how it provides value to what we already have.

kernel/src/constants.rs
```diff
@@ -1 +1,2 @@
-pub const VERSION: &str = "development";
+// pub const VERSION: &str = "development";
```
juli1, line 1: This would break our release mechanism. See [comment here](https://github.com/DataDog/datadog-static-analyzer/pull/17#issuecomment-1627700809).

juli1 (changes requested):
See comment on internal and public builds.

kernel/src/constants.rs
robertohuertasm: Oh, I see. I didn't pay attention to that. I thought the idea was to use the kernel version in Cargo.toml but probably it doesn't make sense at this stage given that the crate is not public in crates.io or something like that. I'll revert that change.

robertohuertasm:
@juli1, I modified the PR's description according to the template as suggested. Hope it's ok now.

robertohuertasm:
Regarding [this comment](https://github.com/DataDog/datadog-static-analyzer/pull/17#issuecomment-1627664191) I maybe can open an issue for further discussion, but in a nutshell:
- [clap](https://docs.rs/clap/latest/clap/) is probably one of the better crates for building CLIs in the rust ecosystem (even the [Rust CLI Working Group example](https://rust-cli.github.io/book/tutorial/cli-args.html#parsing-cli-arguments-with-clap) is using it)
- It's a well-known crate, so it's easier for future contributors to work in the codebase
- It has better ergonomics than `getops` for what I can see
- Offers some features like `version` and `help` out of the box

That being said, it would not be addressing any concerning issue right now. It would be just a Quality-of-Life kind of change which could probably entice further contributors.

kernel/src/constants.rs
juli1: Yes, this is not really documented for now and we took the freedom to have our own system as this tool is published on GitHub release and not on crates.io. We are using the Git `sha` at build time so that we can clearly know what version the user is running. It helps a lot for troubleshooting. We have a similar system for the server, the `/version` endpoint returns the sha of the running code.

kernel/src/constants.rs
```diff
@@ -1 +1,2 @@
...
+pub const VERSION: &str = env!("CARGO_PKG_VERSION");
```
juli1, line 2: We should put it back to `pub const VERSION: &str = "development";` as it is what the build system is expecting to inject the `sha` at build time.

juli1 (changes requested):
Thank you so much for sending this PR. I think this is good because it will provide the version in the command line. One comment on the declaration of the constant to make sure we keep the existing pattern and we should be good to go.

kernel/src/constants.rs
robertohuertasm: Oh, sorry, it seems I forgot to push that change! My bad, I thought it was already reverted. Let me fix it. Thanks for catching it!

robertohuertasm: Already pushed the change. It should be fine now.

juli1 (approved):

PR #18, opened 2023-07-09
feat(workspace): use workspace dependencies

robertohuertasm:

1. The repo has different crates that share some dependencies. Now, those dependencies are declared in each Cargo.toml, along with their versions. This can make difficult to upgrade the version of the dependencies in order to keep all of them in sync.
1. We have a dependency on `serde_derive`.

## What is your solution?

1. The solution is to use [Cargo Workspace Depencencies](https://doc.rust-lang.org/cargo/reference/workspaces.html#the-dependencies-table).
1. Use serde's `derive` feature.

Nor 1 or 2 are inherently a problem. This are suggestions to basically reduce the maintainability cost (1) and reduce the number of third party dependencies (2)

Other projects using similar approach:

- [Cargo](https://github.com/rust-lang/cargo/blob/master/Cargo.toml#L13)
- [Rust Analyzer](https://github.com/rust-lang/rust-analyzer/blob/master/Cargo.toml#L47)
- [Deno](https://github.com/gliheng/deno/blob/main/Cargo.toml#L43)
- [Vercel](https://github.com/vercel/next.js/blob/canary/Cargo.toml#L27)
- [Solana](https://github.com/solana-labs/solana/blob/master/Cargo.toml#L126)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,3 +8,11 @@ members = [
 
 [profile.release]
 lto = true
+
+[workspace.dependencies]
+anyhow = "1"
+base64 = "0.21.2"
+serde = { version = "1", features = ["derive"] }
+serde_json = "1"
+derive_builder = "0.12"
+serde-sarif = "0.4"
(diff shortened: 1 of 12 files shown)


juli1:
Thank you for the pull request! Is it possible to follow the pull request template, explain the problem at hand and the solution? Thanks!

robertohuertasm:
Sure, no problem. I edited the description according to the template as suggested. Apologies for the inconveniences.

juli1:
> Sure, no problem. I edited the description according to the template as suggested. Apologies for the inconveniences.

Thank you so much! It explains the changes and it also helps to educate the team on the rust ecosystem as this is still new to us.

It looks good to me. @edrevo, do you have any comment about this change?

juli1 (approved):

edrevo (approved):

PR #19, opened 2023-07-10
Ensure the server is using snake case in response

juli1:

The server was still using some `camelCase` in the server response. Because of recent renaming, this needs to be changed.

## What is your solution?

Copy the data model that will ensure we do not use `camelCase` and ensure we use `snake_case`.

We also bump the version to be released internally.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -27,8 +27,8 @@ trigger_internal_image:
   variables:
     IMAGE_VERSION: current
     IMAGE_NAME: datadog-static-analyzer
-    RELEASE_TAG: 2023070601
-    BUILD_TAG: 2023070601
+    RELEASE_TAG: 2023071001
+    BUILD_TAG: 2023071001
     TMPL_SRC_IMAGE: v${CI_PIPELINE_ID}-${CI_COMMIT_SHORT_SHA}
     RELEASE_STAGING: "true"
     RELEASE_PROD: "true"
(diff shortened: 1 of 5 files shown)


dastrong (commented):
Why not just use [#[serde(rename_all = "...")]](https://serde.rs/variant-attrs.html#rename_all) instead of doing all this?

juli1:
> Why not just use [#[serde(rename_all = "...")]](https://serde.rs/variant-attrs.html#rename_all) instead of doing all this?

Where would you use this directive?

dastrong:
> > Why not just use [#[serde(rename_all = "...")]](https://serde.rs/variant-attrs.html#rename_all) instead of doing all this?
> 
> Where would you use this directive?



> > Why not just use [#[serde(rename_all = "...")]](https://serde.rs/variant-attrs.html#rename_all) instead of doing all this?
> 
> Where would you use this directive?

I don't a specific spot, but adding this whole helper to do the conversion seems like overkill. 

You had two `#[serde(rename = "executionError")]` and `#[serde(rename = "executionTimeMs")]` camelCase values. Why is everything else needed.

juli1:
> You had two `#[serde(rename = "executionError")]` and `#[serde(rename = "executionTimeMs")]` camelCase values. Why is everything else needed.

I do not understand the comment. First, the `rename` were removed, why adding more then?

The helper functions were added because we need to serialize/deserialize in `snake_case`. In the `kernel`, we need to serialize/deserialize because this is how the `kernel` works (for serialization/deserialization of JavaScript code that is using `camelCase`).

I am not sure what is the recommendation here but if there is something I am missing, please comment to the code, more than happy to have a look and see how we can improve.

dastrong (approved):

PR #23, opened 2023-07-12
Add SARIF validation tests + fix level

juli1:

1. We want to make sure that our SARIF output is compliant with the SARIF standard.
2. We want to also surface the correct `enum` value for level

## What is your solution?

1. Add test with [valico](https://github.com/s-panferov/valico) to validate that the output from the analyzer is compliant with the SARIF schema.
2. Fix the enum value being surface and map `Notice` to `note`

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -63,5 +63,14 @@ curl -X POST \
         ]
      }' \
      http://localhost:8000/analyze
-
 ```
+
+
+curl -X POST \
+-H "Content-Type: application/json" \
+--data '{
+"file_encoding": "utf-8",
+"language": "PYTHON",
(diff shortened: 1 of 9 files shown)


DEVELOPMENT.md
```diff
@@ -63,5 +63,14 @@ curl -X POST \
...
+
+
+curl -X POST \
+-H "Content-Type: application/json" \
+--data '{
+"file_encoding": "utf-8",
+"language": "PYTHON",
+"code": "ZGVmIGZvbyhhcmcxKToKICAgIHBhc3M="
+}' \
+http://localhost:8000/get-treesitter-ast
```
dastrong, line 76: Add a header and place in a shell block

dastrong (approved):

DEVELOPMENT.md
juli1: done

PR #24, opened 2023-07-12
fix(windows): use shell in actions

robertohuertasm:

As I was testing the server binary to see how we could work with it in our extension, I realized that the Windows binary was not sending the version.

## What is your solution?

Not sure if this is going to solve it as it seems that the substitution is taking place, given that when I do a request to the `version` endpoint I don't get `development` but an empty string.

In any case, it's a good practice to use the `bash` shell so typical Unix commands can get executed in Windows. See [here](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsshell) for more information. 

I'm submitting this under the assumption that `sed` will work as expected when executed in a bash shell.

Screenshot of the server not returning anything:

image

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -41,6 +41,7 @@ jobs:
           target: ${{ matrix.target }}
         if: startsWith(matrix.os, 'ubuntu')
       - name: Set the version
+        shell: bash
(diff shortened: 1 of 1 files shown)


juli1 (approved):

robertohuertasm:
@juli1, for the sake of completeness. I tested the [latest release](https://github.com/DataDog/datadog-static-analyzer/releases/tag/0.0.3) and it worked:

image

juli1:
Thank you so much for your help!

PR #27, opened 2023-07-12
moar languages

juli1:

We need to add support for more languages: C#, Dockerfile, Go, Java, JSON, YAML

## What is your solution?

Add the necessary support in our build system

### DIFF
--- a/kernel/build.rs
+++ b/kernel/build.rs
@@ -31,12 +31,22 @@ fn main() {
             .map(|x| dir.join(x))
             .collect();
 
-        cc::Build::new()
-            .cpp(tree_sitter_project.cpp)
-            .include(dir)
-            .files(files)
-            .warnings(false)
-            .compile(tree_sitter_project.name.as_str());
+        if tree_sitter_project.cpp {
+            cc::Build::new()
+                .cpp(tree_sitter_project.cpp)
+                .flag_if_supported("-lstdc++")
+                .include(dir)
+                .files(files)
+                .warnings(false)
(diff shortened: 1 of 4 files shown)


kernel/build.rs
```diff
@@ -19,28 +19,96 @@ fn main() {
         name: String,       // the directory where we clone the project
         repository: String, // the repository to clone
         build_dir: PathBuf, // the directory we use to build the tree-sitter project
+        files: Vec<String>,
+        cpp: bool,
```
dastrong, line 23: What's cpp?

kernel/src/model/common.rs
```diff
@@ -41,10 +53,16 @@ static ALL_LANGUAGES: &[Language] = &[
...
+            Self::Csharp => "C#",
+            Self::Dockerfile => "Dockerfile",
+            Self::Go => "Go",
+            Self::Java => "java",
             Self::JavaScript => "javascript",
+            Self::Json => "json",
             Self::Python => "python",
             Self::Rust => "rust",
             Self::TypeScript => "typescript",
+            Self::Yaml => "yaml",
```
dastrong, line 65: Some have capitals and other don't. Was there a reason for that?

dastrong (approved):

kernel/build.rs
juli1: It's C++ (shortcut for `CPlusPlus`)

kernel/src/model/common.rs
juli1: fixed!

PR #28, opened 2023-07-12
add support for static files

juli1:

We want to support serving static files for the server. We did it before but we need to specify the path for each file. As we will serve all the WASM files, we want to just specify one directory that will contain all the static files.

## What is your solution?

Add a new flag `-s` or `--static` that provides a single directory that contains all static files to serve. If a static file is requested and is in the directory, we return it.

For extra safety, we ensure that the file does not contain `..` or start with `.` to prevent any malicious users to walk through the file hierarchy.

### DIFF
--- a/bins/src/bin/datadog-static-analyzer-server.rs
+++ b/bins/src/bin/datadog-static-analyzer-server.rs
@@ -58,24 +58,16 @@ async fn serve_static(
     server_configuration: &State<ServerConfiguration>,
     name: &str,
 ) -> Option<NamedFile> {
-    server_configuration.static_directory.as_ref()?;
-
-    if name.contains("..") {
+    if server_configuration.static_directory.is_none()
+        || name.contains("..")
+        || name.starts_with('.')
+    {
         return None;
     }
 
-    if name.starts_with('.') {
-        return None;
-    }
+    let s = server_configuration.static_directory.as_ref().unwrap();
 
-    let full_path = Path::new(
-        server_configuration
-            .static_directory
-            .clone()
-            .unwrap()
-            .as_str(),
-    )
-    .join(Path::new(name));
+    let full_path = Path::new(s).join(name);
     NamedFile::open(full_path).await.ok()
 }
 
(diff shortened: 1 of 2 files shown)


bins/src/bin/datadog-static-analyzer-server.rs
```diff
@@ -58,13 +58,25 @@ async fn serve_static(
...
+
+    let full_path = Path::new(
+        server_configuration
+            .static_directory
+            .clone()
+            .unwrap()
+            .as_str(),
+    )
+    .join(Path::new(name));
+    NamedFile::open(full_path).await.ok()
```
robertohuertasm, line 79: What about applying this instead of the current code:
```rust
if server_configuration.static_directory.is_none()
    || name.contains("..")
    || name.starts_with('.')
{
    return None;
}

let s = server_configuration.static_directory.as_ref().unwrap();

let full_path = Path::new(s).join(name);
NamedFile::open(full_path).await.ok()
```

bins/src/bin/datadog-static-analyzer-server.rs
```diff
@@ -58,13 +58,25 @@ async fn serve_static(
     server_configuration: &State<ServerConfiguration>,
     name: &str,
 ) -> Option<NamedFile> {
-    let name_optional = server_configuration.files_to_serve.get(name);
+    server_configuration.static_directory.as_ref()?;
```
robertohuertasm, line 61: This feels weird and hard to read. I understand you're doing it to check that `static_directory` is `Some`.

I would probably use a simple if and collide the 3 conditions (see suggestion below with the full code).

bins/src/bin/datadog-static-analyzer-server.rs
```diff
@@ -58,13 +58,25 @@ async fn serve_static(
...
+    }
+
+    let full_path = Path::new(
+        server_configuration
+            .static_directory
+            .clone()
+            .unwrap()
+            .as_str(),
+    )
+    .join(Path::new(name));
```
robertohuertasm, line 78: No need to use `Path::new` here. The `AsRef` trait will make it work for a `&str`.

bins/src/bin/datadog-static-analyzer-server.rs
```diff
@@ -58,13 +58,25 @@ async fn serve_static(
...
     }
+
+    if name.starts_with('.') {
+        return None;
+    }
+
+    let full_path = Path::new(
+        server_configuration
+            .static_directory
+            .clone()
```
robertohuertasm, line 74: No need to clone if we use `.as_ref()` here.

bins/src/bin/datadog-static-analyzer-server.rs
juli1: I have nothing against this suggestion. Applied it!

dastrong (approved):

PR #31, opened 2023-07-24
Support unnamed nodes capture

juli1:

We want to be able to capture unnamed nodes from the tree-sitter queries.

## What is your solution?

Our solution is to capture unnamed node in the initial called to `map_node`. Doing so allows us to capture unnamed node. For all the children of the node captures, we only capture named nodes.

Added a test from the JIRA ticket to make sure this is correctly captured/handled.

### DIFF
--- a/kernel/src/analysis/tree_sitter.rs
+++ b/kernel/src/analysis/tree_sitter.rs
@@ -102,14 +102,13 @@ pub fn get_query_nodes(
 // map a node from the tree-sitter representation into our own internal representation
 // this is the representation that is passed to the JavaScript layer and how we represent
 // or expose the node to the end-user.
-//
-// If this is NOT a named node, we do not return anything.
 pub fn map_node(node: tree_sitter::Node) -> Option<TreeSitterNode> {
     fn map_node_internal(
         cursor: &mut tree_sitter::TreeCursor,
         only_named_node: bool,
     ) -> Option<TreeSitterNode> {
-        // we do not map space, parenthesis and other non-named nodes.
+        // we do not map space, parenthesis and other non-named nodes if there
+        // when `only_named_node` is true (which is `true` for children only).
         if only_named_node && !cursor.node().is_named() {
             return None;
         }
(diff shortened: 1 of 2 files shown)


kernel/src/analysis/tree_sitter.rs
```diff
@@ -105,17 +105,21 @@ pub fn get_query_nodes(
 //
 // If this is NOT a named node, we do not return anything.
```
dastrong, line 106: You wanna update this comment since this is no longer true

kernel/src/analysis/tree_sitter.rs
```diff
@@ -105,17 +105,21 @@ pub fn get_query_nodes(
...
 pub fn map_node(node: tree_sitter::Node) -> Option<TreeSitterNode> {
-    fn map_node_internal(cursor: &mut tree_sitter::TreeCursor) -> Option<TreeSitterNode> {
+    fn map_node_internal(
+        cursor: &mut tree_sitter::TreeCursor,
+        only_named_node: bool,
+    ) -> Option<TreeSitterNode> {
         // we do not map space, parenthesis and other non-named nodes.
```
dastrong, line 112:
```suggestion
        // we do not map space and parenthesis nodes
        // we do capture non-named nodes
```

kernel/src/analysis/tree_sitter.rs
```diff
@@ -145,7 +149,10 @@ pub fn map_node(node: tree_sitter::Node) -> Option<TreeSitterNode> {
     }
 
     let mut ts_cursor = node.walk();
-    map_node_internal(&mut ts_cursor)
+
+    // Initially, we do not capture only named node to allow capturing unnamed node from
+    // the tree-sitter query.
```
dastrong, line 154:
```suggestion
    // Initially, we capture both un/named nodes to allow capturing unnamed node from
    // the tree-sitter query.
```

dastrong (changes requested):

dastrong (approved):

PR #53, opened 2023-08-09
add SHA to SARIF report

juli1:

We want to add the SHA related to a violation in the SARIF report.

## What is your solution?

1. We are adding a `-g` option
2. When the `-g` option is specific, we add the `SHA` for a violation by running `git blame` on it

This has been tested against the repository https://github.com/juli1/rosie-tests (see the section on integration tests below).

Since it's hard to have unit tests for such a feature, we added a script for doing integration tests. This integration test script is found in `misc/integration-test.sh`. It is added in our GitHub actions.

This scripts
 - Get the repository https://github.com/juli1/rosie-tests
 - Run the new stella with the `-g` option
 - Check that the SHA and the CATEGORY tags are in the report

### DIFF
--- a/.github/workflows/integration-tests.yaml
+++ b/.github/workflows/integration-tests.yaml
@@ -1,5 +1,5 @@
 on: push
-name: Check Code
+name: Run Integration tests
 jobs:
   integration_tests:
     runs-on: ubuntu-latest
--- a/.github/workflows/integration-tests.yaml
+++ b/.github/workflows/integration-tests.yaml
@@ -1,0 +1,20 @@
+on: push
+name: Check Code
+jobs:
+  integration_tests:
+    runs-on: ubuntu-latest
+    env:
+      DD_API_KEY: ${{ secrets.DD_API_KEY }}
+      DD_APP_KEY: ${{ secrets.DD_APP_KEY }}
+      DD_SITE: ${{ secrets.DD_SITE }}
+    steps:
+      - uses: actions/checkout@v1
+      - name: Install Stable + Rustfmt + Clippy
+        uses: actions-rs/toolchain@v1
+        with:
+            toolchain: stable
+            components: clippy
+            override: true
+            default: true
+      - name: Run Integration tests
+        run: ./misc/integration-test.sh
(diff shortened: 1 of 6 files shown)


dastrong (commented):
Document the integration tests

------

Do we really need a flag `-g` to add this info? If a `.git` folder exists, add the commit info. If there's no `.git` folder, don't add.

------

We're calling blame on the starting line of the violation. What about a violation where that's reported for the surrounded element of where the change was made?

For example say I removed the default statement and had the following code. For an violation, that I introduced, the analyzer would attach the commit SHA from whoever added/changed the `switch(num) {` line.
```java
class Foo {
    public String getValue(int num) {
        switch (num) {            // violation starts here
            case 1:
                return "One";
            case 2:
                return "Two";
            case 3:
                return "Three";   
        }                                    // violation ends here
    }
}
```

------

An SHA isn't specific to Datadog like the categories are. I would prefer to place the SHA somewhere more general to allow us to integrate with other SARIF tools better in the future. Inside the result object, I propose using the partialFingerprints, so if we decide in the future to add other information we can easily (_we'd only use commitSha for now_)
```json
"partialFingerprints": {
  "commitSha": "1d02cdf954371c7ac4cf70dd4d9d42c299114935",
  "email": "danielstrong.ds@gmail.com",
  "author": "Daniel Strong",
  "date": "2023-07-26T19:32:18Z",
  "commitMessage": "Update password.py"
}
```

juli1:
> Document the integration tests


> Do we really need a flag `-g` to add this info? If a `.git` folder exists, add the commit info. If there's no `.git` folder, don't add.

We do. Maybe you do not want to do it or add the information. The user should decide.


> We're calling blame on the starting line of the violation. What about a violation where that's reported for the surrounded element of where the change was made?
> 
> For example say I removed the default statement and had the following code. For an violation, that I introduced, the analyzer would attach the commit SHA from whoever added/changed the `switch(num) {` line.
> 
> ```java
> class Foo {
>     public String getValue(int num) {
>         switch (num) {            // violation starts here
>             case 1:
>                 return "One";
>             case 2:
>                 return "Two";
>             case 3:
>                 return "Three";   
>         }                                    // violation ends here
>     }
> }
> ```

Can you explain the issue?

> 
> An SHA isn't specific to Datadog like the categories are. I would prefer to place the SHA somewhere more general to allow us to integrate with other SARIF tools better in the future. Inside the result object, I propose using the partialFingerprints, so if we decide in the future to add other information we can easily (_we'd only use commitSha for now_)
> 
> ```json
> "partialFingerprints": {
>   "commitSha": "1d02cdf954371c7ac4cf70dd4d9d42c299114935",
>   "email": "danielstrong.ds@gmail.com",
>   "author": "Daniel Strong",
>   "date": "2023-07-26T19:32:18Z",
>   "commitMessage": "Update password.py"
> }
> ```

Gonna add, thank you.

dastrong (approved):

PR #76, opened 2023-09-03
Add progress bar in the CLI

juli1:

We want to add a progress bar in the command line.

## What is your solution?

Use the [indicatif](https://docs.rs/indicatif/latest/indicatif/) library to show the progress. We do not use colors as it could sometimes not render well in logs.

## Testing

Screenshot 2023-09-03 at 4 23 34 PM

 - Fixed a bug to make sure we report issues all all languages, not only the latest one.
 - Also enable the progress bar only when not in debug mode. Otherwise, it puts too much information on the terminal.

### DIFF
--- a/cli/src/file_utils.rs
+++ b/cli/src/file_utils.rs
@@ -9,6 +9,7 @@ use walkdir::WalkDir;
 
 static FILE_EXTENSIONS_PER_LANGUAGE_LIST: &[(Language, &[&str])] = &[
     (Language::JavaScript, &["js", "jsx"]),
+    (Language::Dockerfile, &["docker", "dockerfile"]),
     (Language::Python, &["py", "py3"]),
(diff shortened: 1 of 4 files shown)


dastrong (approved):

cli/src/file_utils.rs
```diff
@@ -9,6 +9,7 @@ use walkdir::WalkDir;
 
 static FILE_EXTENSIONS_PER_LANGUAGE_LIST: &[(Language, &[&str])] = &[
     (Language::JavaScript, &["js", "jsx"]),
+    (Language::Dockerfile, &["docker", "dockerfile"]),
```
dastrong, line 12:
```suggestion
```

You don't mention this in the description nor test for it. See comments on Slack

PR #84, opened 2023-09-08
Add --staging flag to fetch rules from staging

juli1:

Since we will be using staging to write rules, we need to be able to easily test the rules in staging.

## What is your solution?

We are introduced the flag `--staging` that will force the analyzer and other binaries to use the staging endpoint.


dastrong (commented):
Since this overrides the `DD_SITE` variable, I'd prefer using `force_staging` instead to help indicate that

dastrong (approved):

PR #96, opened 2023-09-25
Add integration tests for production rules

juli1:

We want to make sure that all rules released in production are working in both the CLI and the server.

## What is your solution?

The solution showed in this PR is for the **CLI ONLY**.

A python script (`test-production-rules.py`) test production rules and verify that the number of expected annotations match the number of violations reported by the CLI.

We will do a follow-up to test the server.

### DIFF
--- a/cli/src/file_utils.rs
+++ b/cli/src/file_utils.rs
@@ -9,6 +9,7 @@ use walkdir::WalkDir;
 
 static FILE_EXTENSIONS_PER_LANGUAGE_LIST: &[(Language, &[&str])] = &[
     (Language::JavaScript, &["js", "jsx"]),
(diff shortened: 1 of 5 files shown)


dastrong (approved):

cli/src/file_utils.rs
```diff
@@ -9,6 +9,7 @@ use walkdir::WalkDir;
 
 static FILE_EXTENSIONS_PER_LANGUAGE_LIST: &[(Language, &[&str])] = &[
     (Language::JavaScript, &["js", "jsx"]),
+    (Language::Java, &["java"]),
```
dastrong, line 12: @juli1 You're missing the `.jar` and `.class` file extensions

PR #99, opened 2023-09-27
Update time for integration tests

juli1:

Update time for integration tests, run before to make sure it executes when we wake up.

### DIFF
--- a/.github/workflows/integration-tests.yaml
+++ b/.github/workflows/integration-tests.yaml
@@ -1,7 +1,7 @@
 on:
   push:
   schedule:
-    # run every day at 19m UTC
+    # run every day at 9am UTC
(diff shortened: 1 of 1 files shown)


.github/workflows/integration-tests.yaml
```diff
@@ -1,8 +1,8 @@
 on:
   push:
   schedule:
-    # run every day at 10am UTC (5am EST)
-    - cron:  '0 11 * * *'
+    # run every day at 19m UTC
+    - cron:  '0 9 * * *'
```
dastrong, line 5:
```suggestion
    # run every day at 9:00am UTC (5:00am EST)
    - cron:  '0 9 * * *'
```

PR #104, opened 2023-09-28
Support Kotlin and Swift

juli1:

We want to make sure we can support Kotlin and Swift.

## What is your solution?

1. Adding support for the language
2. Add test to make sure the languages parse

The Swift grammar does not build out of the box. We need to use the rust package instead of using the sources.

### DIFF
--- a/kernel/src/analysis/tree_sitter.rs
+++ b/kernel/src/analysis/tree_sitter.rs
@@ -3,6 +3,7 @@ use crate::model::common::{Language, Position};
 use anyhow::Result;
 use std::collections::HashMap;
 use tree_sitter::QueryCursor;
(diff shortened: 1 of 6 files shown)


kernel/src/analysis/tree_sitter.rs
```diff
@@ -3,6 +3,7 @@ use crate::model::common::{Language, Position};
 use anyhow::Result;
 use std::collections::HashMap;
 use tree_sitter::QueryCursor;
+use tree_sitter_swift::language as swift_language;
```
dastrong, line 6: Could you leave a brief comment here explaining why Swift is done differently

dastrong (approved):

PR #109, opened 2023-10-05
feat(server): added new yaml parser endpoint

robertohuertasm:

This PR is adding a new endpoint to the server that will parse a YAML file and return it as JSON. This will avoid having to use 3rd party libraries in the extensions, particularly in VS Code, which has JSON support out of the box.

Aside from that, the PR adds two more tools which may be useful for the developers:

- [Cargo Make](https://sagiegurari.github.io/cargo-make/): Cross OS task runner. Helpful to execute some recurring tasks, like starting the server in watch mode so it restarts when a file is changed, different clippy combinations...
- HTTP file: can be used along with some extensions ([REST Client]() in VS Code, for example) to easily test the server endpoints.

## What is your solution?

Example of REST Client:
image

I added some notes to the `Tools` section in the readme.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,4 +16,5 @@ serde = { version = "1", features = ["derive"] }
 serde_json = "1"
 derive_builder = "0.12"
 serde-sarif = "0.4"
+serde_yaml = "0.9"
 sha2 = "0.10.7"
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -1,0 +1,12 @@
+[tasks.server]
+workspace = false
+command = "cargo"
+watch = { watch = ["./bins", "./kernel", "./cli", "./server", "Cargo.toml"] }
+args = ["run", "--bin", "datadog-static-analyzer-server"]
+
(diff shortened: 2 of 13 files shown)


juli1:
Why parsing YAML using the rust parser while we can add YAML support for the static analyzer? We could instead integrate YAML as a language, which would then allow us to check YAML files in the static analyzer. There is already [a tree-sitter grammar](https://github.com/ikatyang/tree-sitter-yaml) for it, it should be straightforward to integrate.

robertohuertasm:
@juli1 apologies because I think I didn't give enough context. The basic idea was to provide a simple endpoint that would help us parse the `static-analysis.datadog.yml` to get the rulesets. That's it. We were not thinking about a more complex scenario. We don't want to analyze the YAML file but to just transform it into JSON. We can take it offline if you prefer and discuss the alternatives.

robertohuertasm:
After some offline discussion we're going to handle this from the client, so I'm closing this.

PR #114, opened 2023-10-10
Add YAML support

juli1:

Add YAML support. This is required to handle `yarn.lock` files and other configuration files.

## What is your solution?

Add support for YAML and add a test


dastrong (approved):
> Add YAML support. This is required to handle `yarn.lock` files and other configuration files.

Approved it for the YAML logic portion. However, I don't see us processing `.lock` files like the description mentions

juli1:
> > Add YAML support. This is required to handle `yarn.lock` files and other configuration files.
> 
> Approved it for the YAML logic portion. However, I don't see us processing `.lock` files like the description mentions?

This will be done later in another program. We first add support for YAML and we will later add the lockfile logic later.

PR #119, opened 2023-10-27
Add subdirectory option

juli1:
## What problem are you trying to solve?

We want to be able to scan a sub-directory of a repository.

## What is your solution?

We are providing a `--subdirectory` option (short option: `-u`). This will scan only the files in this subdirectory. The git metadata are still read from the directory.

If running `datadog-static-analyzer -i /path/to/code -o plop.json -u plop -g`

- The Git metadata are read from `/path/to/code/.git`
- Only code in `/path/to/code/plop` is scanned 

## What the reviewer should know

- Bump internal version numbers
- Bump version number to release a new CLI

## Testing

Screenshot 2023-10-27 at 7 45 26 PM

PR #121, opened 2023-10-29
fix(shutdown): correctly handle rocket errors

### LINKED ISSUE #120: Process never exits when fails to bind the port in v0.1.5 [Server]
**Describe the bug**
Binary v0.1.5 does not exit when it fails to bind the port. This is making the process to hang when started. 

**To Reproduce**
- Run a process in port 9090
- Run the binary with -p 9090
- We will get an error in the terminal but the process never exits

**Expected behavior**
The process should exit with a particular code. Before it was `101`.

**Screenshots**
Here you can see that 0.1.5 hangs.

image

**Desktop (please complete the following information):**
 - OS: MacOS Ventura 13.6

robertohuertasm:
Fixes #120 

## What is your solution?
If we're not using the `keep alive` functionality, we will just wait for the Rocket thread to finish.

If not, we will await for the first occurrence. Either the Rocket thread finishes (for instance, due to a panic...) or the keep alive functionality notifies the shutdown and causes the `shutdown handle` to finish.

## Other considerations
I took the time to fix some clippy issues. It was impossible to even commit without doing `--no-verify`. I'm afraid potential contributors may be facing the same issue.  Is that expected or are you not getting that at all when trying to commit to the repo?

IDE-1744

### DIFF
--- a/bins/Cargo.toml
+++ b/bins/Cargo.toml
@@ -18,13 +18,18 @@ name = "datadog-static-analyzer-server"
 [dev-dependencies.cargo-husky]
 version = "1"
 default-features = false # Disable features which are enabled by default
-features = ["precommit-hook", "run-cargo-test", "run-cargo-clippy", "run-cargo-fmt"]
+features = [
+  "precommit-hook",
+  "run-cargo-test",
+  "run-cargo-clippy",
+  "run-cargo-fmt",
+]
 
 [dependencies]
 # local
-cli = {path = "../cli"}
-kernel = {path = "../kernel" }
-server = {path = "../server"}
+cli = { path = "../cli" }
+kernel = { path = "../kernel" }
+server = { path = "../server" }
 # workspace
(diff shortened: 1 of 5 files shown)


juli1 (approved):
Thanks for your contribution! I never had any issue with clippy in the past. It should be working normally.

robertohuertasm:
Thanks @juli1! Not sure if you were planning to create a release soon, but I'd really appreciate having this already internally in order to leverage this behavior in VS Code.

juli1:
> Thanks @juli1! Not sure if you were planning to create a release soon, but I'd really appreciate having this already internally in order to leverage this behavior in VS Code.

New release should come tomorrow once [this](https://github.com/DataDog/datadog-static-analyzer/pull/119) is merged.

robertohuertasm:
Saw v0.1.6 already! Thank you for that! 🙌🏻

PR #124, opened 2023-10-31
feat(shutdown): add new shutdown endpoint under a flag

robertohuertasm:

In the IDE we have to deal with some scenarios where the server must be restarted. This basically means that we have to kill the server and restart it again. The most prominent use case for this scenario is when we receive a new binary update.

The main problem here is that we may need to restart the server from an IDE instance that didn't spawn the process, so we need to use several strategies to share the process id amongst instances which are error prone or not reliable. Specifically, in VS Code, we have some constraints around that. 

Aside from that, we have to use different approaches for each of the OS we support in some of the cases, so with this proposed solution we think we would greatly simplify our problem.

- It would allow for all the IDEs to use it, without having to replicate the logic of killing the process from random instances.
- It would reduce the dependencies on 3rd party libraries to help us mitigate the multiple OS support.
- We would be able to reliably shutdown the server the moment we want.

## What is your solution?

We have added a new flag `-e` that when present will enable a new `shutdown` endpoint. By doing this, the user will decide whether they need this behaviour or not.

The endpoint will accept both the verbs `GET` and `POST` and will return a 403 (Forbidden) if the flag `-e` is not present. Other wise:

- GET: will get a 204 if the server can be shutdown. This is handy in some cases where we want to understand if the server is killable or not (i.e. showing a command to stop it or not...)
- POST: will return a 200 and shutdown the server.

## Other comments

I fixed some clippy warnings that were avoiding me to commit properly because of the pre-commit hook.

IDE-1896

### DIFF
--- a/bins/src/bin/datadog-static-analyzer-server.rs
+++ b/bins/src/bin/datadog-static-analyzer-server.rs
@@ -4,7 +4,7 @@ use lazy_static::lazy_static;
 use rocket::fairing::{Fairing, Info, Kind};
 use rocket::fs::NamedFile;
 use rocket::futures::FutureExt;
-use rocket::http::Header;
+use rocket::http::{Header, Status};
 use rocket::serde::json::{json, Json, Value};
 use rocket::{Build, Ignite, Request as RocketRequest, Response, Rocket, Shutdown, State};
 use server::model::analysis_request::AnalysisRequest;
@@ -18,8 +18,6 @@ use std::sync::Mutex;
 use std::time::{Duration, SystemTime, UNIX_EPOCH};
 use std::{env, process, thread};
 
-pub struct CORS;
-
 /// get the current timestamp
 fn get_current_timestamp_ms() -> u128 {
     SystemTime::now()
@@ -27,6 +25,8 @@ fn get_current_timestamp_ms() -> u128 {
         .unwrap()
         .as_millis()
 }
+
+pub struct CORS;
 
 // Adding CORS for the server.
 // See https://stackoverflow.com/questions/62412361/how-to-set-up-cors-or-options-for-rocket-rs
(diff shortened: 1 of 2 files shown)


bins/src/bin/datadog-static-analyzer-server.rs
```diff
@@ -51,6 +51,25 @@ impl Fairing for CORS {
...
 }
 
+#[rocket::get("/shutdown")]
+fn shutdown_get(server_configuration: &State<ServerConfiguration>) -> Status {
+    if server_configuration.is_shutdown_enabled {
+        Status::NoContent
+    } else {
+        Status::Forbidden
+    }
+}
```
juli1, line 61: Please add a documentation to explain what this endpoint does and the goal.

bins/src/bin/datadog-static-analyzer-server.rs
```diff
@@ -51,6 +51,25 @@ impl Fairing for CORS {
...
+#[rocket::post("/shutdown")]
+fn shutdown_post(server_configuration: &State<ServerConfiguration>, shutdown: Shutdown) -> Status {
+    if server_configuration.is_shutdown_enabled {
+        shutdown.notify();
+        Status::Ok
+    } else {
+        Status::Forbidden
+    }
+}
+
```
juli1, line 72: Similar, please add a comment to explain the logic as this is not documented elsewhere.

bins/src/bin/datadog-static-analyzer-server.rs
```diff
@@ -51,6 +51,25 @@ impl Fairing for CORS {
     }
 }
 
+#[rocket::get("/shutdown")]
+fn shutdown_get(server_configuration: &State<ServerConfiguration>) -> Status {
+    if server_configuration.is_shutdown_enabled {
+        Status::NoContent
```
juli1, line 57: Why not `200` or `Ok`? After all, if we can shutdown, it's okay?

juli1 (approved):
Fix and ship: please add comments on the two methods.

bins/src/bin/datadog-static-analyzer-server.rs
robertohuertasm: We could have used 200, sure. But in this case, a 204 is also a success code that indicates that no content is expected from that endpoint. I mean, I don't have a strong opinion here aside than what I already stated, so if you prefer returning a 200 code, I'd be fine with it. 

I now realise that I used a 200 for the post version, which doesn't return any content either. I would unify both codes with 204, or 200 if you prefer that for some other reason.

juli1 (approved):

PR #125, opened 2023-10-31
feat(headers): add custom headers with server information

robertohuertasm:

While integrating the static analyzer with VS Code we found it would be useful to get some extra information about the server in each request so we can optimize the amount of request we do to the server itself.

For instance, if a particular instance of VS Code is pinging to a server that has been updated by another instance, we may have the case where some instances are not aware that they're pinging now to a different version of the server.

Now, we have abused some of the VS Code APIs to broadcast this to the rest of the instances, so they can react accordingly and clear their cached information (supported languages...).

Although this works, we cannot guarantee that these messages are properly broadcasted in a timely manner. In the end, we're using an API that was not conceived for that. 

So the idea for us would be to get rid of this mechanism and pivot to another strategy where we would check if we're querying the right version of the server and react accordingly per instance if we're not.

Unfortunately, this it's also not optimal as we would have to do 2 requests always. One to check the version, another one being the actual request.

## What is your solution?

Just by adding some custom headers to all the responses we could get that kind of information for free. So the solution is create a middleware that would add some custom headers:

- `X-static-analyzer-server-version`:  version of the server
- `X-static-analyzer-server-revision`: revision of the server
- `X-static-analyzer-server-shutdown-enabled`: "true" or "false", will be also useful for us. 

## Comments

This PR is based on #124 

image

IDE-1900

### DIFF
--- a/bins/src/bin/datadog-static-analyzer-server.rs
+++ b/bins/src/bin/datadog-static-analyzer-server.rs
@@ -51,6 +51,38 @@ impl Fairing for CORS {
     }
 }
 
+pub struct CustomHeaders;
+
+#[rocket::async_trait]
+impl Fairing for CustomHeaders {
+    fn info(&self) -> Info {
+        Info {
+            name: "Custom Headers",
+            kind: Kind::Response,
+        }
+    }
+
+    async fn on_response<'r>(&self, request: &'r RocketRequest<'_>, response: &mut Response<'r>) {
+        let state = request.guard::<&State<ServerConfiguration>>().await;
+
+        if let rocket::outcome::Outcome::Success(server_configuration) = state {
+            response.set_header(Header::new(
+                "X-static-analyzer-server-shutdown-enabled",
+                server_configuration.is_shutdown_enabled.to_string(),
+            ));
+        }
+
+        response.set_header(Header::new(
+            "X-static-analyzer-server-version",
+            get_version(),
+        ));
+        response.set_header(Header::new(
+            "X-static-analyzer-server-revision",
+            get_revision(),
+        ));
+    }
+}
+
 /// The shutdown endpoint, when a GET request is received, will return a 204 code if the shutdown mechanism is enabled.
 /// It will return a 403 code otherwise.
 ///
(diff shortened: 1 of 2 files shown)


bins/src/bin/datadog-static-analyzer-server.rs
```diff
@@ -51,6 +51,38 @@ impl Fairing for CORS {
...
+            kind: Kind::Response,
+        }
+    }
+
+    async fn on_response<'r>(&self, request: &'r RocketRequest<'_>, response: &mut Response<'r>) {
+        let state = request.guard::<&State<ServerConfiguration>>().await;
+
+        if let rocket::outcome::Outcome::Success(server_configuration) = state {
+            response.set_header(Header::new(
+                "X-static-analyzer-server-shutdown-enabled",
```
juli1, line 70: Put the flag as a constant in `constants.rs` as `SERVER_HEADER_SHUTDOWN_ENABLED` so all headers (either, request or response) are in one place.

bins/src/bin/datadog-static-analyzer-server.rs
```diff
@@ -51,6 +51,38 @@ impl Fairing for CORS {
...
+                server_configuration.is_shutdown_enabled.to_string(),
+            ));
+        }
+
+        response.set_header(Header::new(
+            "X-static-analyzer-server-version",
```
juli1, line 76: Similar to the comment above (put the flag as a constant in `constants.rs`)

bins/src/bin/datadog-static-analyzer-server.rs
```diff
@@ -51,6 +51,38 @@ impl Fairing for CORS {
...
+            get_version(),
+        ));
+        response.set_header(Header::new(
+            "X-static-analyzer-server-revision",
```
juli1, line 80: Similar to the comment above (put the flag as a constant in `constants.rs`)

juli1 (changes requested):

bins/src/bin/datadog-static-analyzer-server.rs
robertohuertasm: Added them to `server/constants.rs`.

juli1 (approved):

PR #127, opened 2023-10-31
feat(server): extend the keepalive behavior to all requests 

robertohuertasm:

Now only the ping requests are extending the life of the server if keep alive is on. No other requests matter.

## What is your solution?

Added a new fairing/middleware so that if the keep alive functionality is on, all the requests will slide the keep alive counter.

## Notes

Built on top of #126 

IDE-1896


robertohuertasm:
@juli1 this PR was built on top of #126 which has not been merged yet. Just in case you didn't notice it.

PR #128, opened 2023-10-31
refactor(server): split the code in several files

robertohuertasm:

This is a PR for your consideration. As we continue evolving the server it may be easy to have the different parts ofthe code separated so users can easily locate the different pieces.

For instance, now we have a file called `fairings.rs` and another one called `endpoints.rs`. This makes easy to decide how to organize the code.

## What is your solution?

Splitted the code in several files according to their functionality

This is just a suggestion. I can see that you may prefer to have everything in one unique file. No problems if we have to close this.

Builds on top of #126


juli1 (approved):
Sorry for getting back on this right now as other project items have been taking my time. The changes look good and I think it promotes good engineering practices (smaller files, separation of concerns, etc).

Thanks for sending this PR and thank you for your patience!

PR #129, opened 2023-11-02
feat(server): add new version guard validation

robertohuertasm:

In some binary update scenarios we may have VS Code instances which may not be notified of a server restart. This could cause some requests being sent to the new server running with the expectation of it being the old version. As this instances may have cached some information from the previous server which may not be relevant anymore, they should ping the server for the version all the time in order to control this. 

A way to mitigate this, is the solution proposed below.

See related [JIRA ticket](https://datadoghq.atlassian.net/browse/IDE-1913) for more information.

## What is your solution?

Added a new guard which will check the existence of the `X-Static-Analyzer-Server-Expected-Version`. If it's not there, the request will continue as expected. If it's there, it will check that the version matches the one from the server and will return a `BadRequest` code if they don't match.

Note that this behavior is totally opt-in. If the header is not sent, no validation will happen.

IDE-1913

### DIFF
--- a/bins/src/bin/datadog-static-analyzer-server.rs
+++ b/bins/src/bin/datadog-static-analyzer-server.rs
@@ -4,11 +4,12 @@ use rocket::fairing::{Fairing, Info, Kind};
 use rocket::fs::NamedFile;
 use rocket::futures::FutureExt;
 use rocket::http::{Header, Status};
+use rocket::request::{FromRequest, Outcome};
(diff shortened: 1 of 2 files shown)


robertohuertasm:
For reference, the discussion about if this PR is pertinent of not is held in the [JIRA ticket IDE-1913](https://datadoghq.atlassian.net/browse/IDE-1913).

robertohuertasm:
Discarded for the moment.

PR #134, opened 2023-11-13
chore(server): add launch debug configuration for vscode

robertohuertasm:

While working in the static analysis integration for VS Code I find useful to debug the server to get insights on some of the errors.

## What is your solution?

I added a launch configuration for vs code to start a debug session for the server.

I know this works only for VS Code and maybe the owners of the repo don't want to commit IDE related files, but since VS Code is widely used I pushed this for your consideration.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,26 @@
+{
+  // Use IntelliSense to learn about possible attributes.
+  // Hover to view descriptions of existing attributes.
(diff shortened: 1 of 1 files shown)


robertohuertasm:
@juli1 not sure why the mac check build is failing... something to do with ssl. I'm pretty sure my change doesn't have to do with that, so it may be something related to GitHub runners configuration.

juli1 (approved):

PR #145, opened 2023-12-07
Move the v8 snapshot creation to build time

abhirag:
As of now we create the v8 snapshot lazily during runtime, moving the snapshot creation to build time should improve startup time
## What is your solution?
Move the snapshot creation to build time, the v8 snapshot bytes are then included in the binary
## Alternatives considered
Alternative is the current approach where we create the v8 runtime snapshot lazily during runtime
The tradeoff is between better startup times and a tad fatter binary. I can try and profile to get some concrete numbers if the tradeoff seems unacceptable. The approach I put forward is how Deno's cli seems to be handling snapshotting too. Also I had to fix some clippy lints ([manual_retain](https://rust-lang.github.io/rust-clippy/master/index.html#/manual_retain), [iter_overeager_cloned](https://rust-lang.github.io/rust-clippy/master/index.html#/iter_overeager_cloned))


juli1:
Sorry for the late reply, just looked at the project and noticed the PR! We will get back to you shortly on this!

kernel/build.rs
```diff
@@ -1,6 +1,27 @@
...
+            snapshot_path,
+            startup_snapshot: None,
+            skip_op_registration: false,
+            extensions: vec![deno_extension::init_ops_and_esm()],
+            compression_cb: None,
+            with_runtime_cb: None,
+        },
+    );
+}
+
```
juli1, line 24: Please put a comment explaining what it does. Especially that it build the snapshot that will be later used by a static variable. We may also put a reference to the cargo book about the environment variable, explaining where to put [references to `OUT_DIR`](https://doc.rust-lang.org/cargo/reference/environment-variables.html)

kernel/src/analysis/javascript.rs
```diff
@@ -3,25 +3,17 @@ use crate::model::analysis::{
...
 
-lazy_static! {
-    static ref STARTUP_DATA: Vec<u8> = {
-        let code: FastString = FastString::from_static(include_str!("./js/stella.js"));
-        let mut rt = JsRuntimeForSnapshot::new(RuntimeOptions::default());
-        rt.execute_script("common_js", code).unwrap();
-        rt.snapshot().to_vec()
-    };
-}
+static STARTUP_DATA: &[u8] = include_bytes!(concat!(env!("OUT_DIR"), "/DENO_SNAPSHOT.bin"));
```
juli1, line 16: Put a comment that explains this file is built at compile-time in `build.rs`

juli1 (changes requested):

juli1 (approved):

abhirag:
@juli1 Thanks for your feedback, added the required comments

PR #155, opened 2023-12-20
Workaround x86 bug

juli1:

The server is crashing, which is likely due to a dependency upgrade. It seems that [this bug](https://github.com/denoland/deno_core/issues/217) is still hitting us, this issue occurred in the past.

## What is your solution?

Try fix from [this PR](https://github.com/denoland/deno/pull/20495)


egfx-notifications:
Just in case this helps, it seems to be fixed in current deno versions, at least I do not get the segfaults anymore with them. There has been a fix for deno in https://github.com/denoland/deno/pull/20495, but it is in the cli, not the deno_core component. As a consumer of deno_core you may need to apply the same fix to your code for now?

juli1:
> Just in case this helps, it seems to be fixed in current deno versions, at least I do not get the segfaults anymore with them. There has been a fix for deno in [denoland/deno#20495](https://github.com/denoland/deno/pull/20495), but it is in the cli, not the deno_core component. As a consumer of deno_core you may need to apply the same fix to your code for now?

It actually helps a lot! So we just need to call `deno_core::JsRuntime::init_platform(None);` once on the main thread before calling any JS code?

egfx-notifications:
> > Just in case this helps, it seems to be fixed in current deno versions, at least I do not get the segfaults anymore with them. There has been a fix for deno in [denoland/deno#20495](https://github.com/denoland/deno/pull/20495), but it is in the cli, not the deno_core component. As a consumer of deno_core you may need to apply the same fix to your code for now?
> 
> It actually helps a lot! So we just need to call `deno_core::JsRuntime::init_platform(None);` once on the main thread before calling any JS code?

That's how I understand the fix in deno. May be worth a try.

juli1:
> That's how I understand the fix in deno. May be worth a try.

Alright, trying in this PR then :)

PR #167, opened 2024-01-02
[STAL-994] Pin tree-sitter grammars

jasonforal:
During our build stage, we currently pull tree-sitter grammars via git, which defaults to the latest commit on the repo's default branch. Thus, upstream changes to these repos could silently break our rules.

## What is your solution?
Pin the tree-sitter grammar git repos to a specific commit hash

## Alternatives considered
* git submodules were considered

* If a nonexistent git hash is provided, the build will panic


juli1 (approved):

juli1:
Please make sure the pipeline passes before merging

PR #168, opened 2024-01-02
Adding support for C

abhirag:
Adding support for C

Started testing using some rules listed [here](https://github.com/abhirag/static-analyzer-c-rules/blob/main/README.md) but  abandoning those rules and basing my testing on rules derived from [CWE](https://cwe.mitre.org/index.html) and CERT C coding standard


abhirag:
@juli1 / @jasonforal  I was trying to write some C rules to test for common memory safety issues and keep tripping up on identifying calls in the same local scope. For example when trying to identify double-free I would need to identify two `free` calls in the same local scope. Do you have any pointers on constructing such queries?

PR #170, opened 2024-01-05
[STAL-1039] fix: square bracket filename encoding

dastrong:

A filename that has `[` or `]` in it will fail the SARIF schema validation.

## What is your solution?

Encode these two characters.

## Alternatives considered

None.

This change is important to have as it's needed to dogfood.


juli1 (approved):

juli1:
@dastrong please update the gitlab version

dastrong:
@jasonforal added `#` and `%` as well per your comment.

PR #179, opened 2024-01-16
[STAL-1102] Build tree-sitter queries once per rule

jasonforal:
* Improve runtime performance by building tree-sitter queries once per rule

## What is your solution?
* Have `InternalRule` own a `tree_sitter::Query`

* The least-invasive way to annotate that `rule` is valid for the thread's lifetime is to create a thread scope. While this isn't a cost-free abstraction, the overhead isn't significant. A future refactor to how we handle concurrency will address this (among other things) in a cleaner fashion.


kernel/src/model/rule.rs
```diff
@@ -204,13 +205,14 @@ impl Rule {
...
 
-        let tree_sitter_query_code = String::from_utf8(
+        let tree_sitter_query = String::from_utf8(
             general_purpose::STANDARD.decode(
                 self.tree_sitter_query_base64
                     .as_ref()
                     .ok_or_else(|| anyhow!("tree sitter query is empty"))?,
             )?,
         )?;
+        let tree_sitter_query = get_query(&tree_sitter_query, &self.language)?;
```
juli1, line 215: We should show/log an error message to the user when we do not get the query.

juli1 (changes requested):

PR #204, opened 2024-02-09
Jacobotb/prototype/include exclude

jacobotb:


jacobotb:
It makes sense to think of a trie for this, but I think it gets complicated because the `only` and `ignore` fields can not only contain path prefixes, but also glob patterns, and because we have to operate at several levels:

As I keep thinking of it, I keep going between "it would work" and "it wouldn't work". Right now I have an idea about it, so I'll see if I can prototype it tomorrow.

PR #205, opened 2024-02-09
Add Vectorscan bindings and library

jasonforal:
Adds support for Vectorscan.

  * CI and multi-platform/arch build support will come in a later PR. Vectorscan is a beast to build, and until I can figure out a way to make it play nicely within our Cargo workspace, I am excluding it from being compiled by default. This lets us merge into main without having to completely refactor CI.
  * I am not sure if there is a better way to _include_ crates in a Cargo workspace but _exclude_ them from being compiled by default other than setting `default-members` in Cargo to every crate except the ones you want to ignore. This workaround has no side effects and will be reverted eventually.
  * As internally discussed, we will use build-time repo fetching instead of using git submodules

### DIFF
--- a/crates/vectorscan-sys/build.rs
+++ b/crates/vectorscan-sys/build.rs
@@ -30,6 +30,21 @@ impl Dependency {
     }
 }
 
+/// At a high level, this build script builds [Vectorscan](https://github.com/VectorCamp/vectorscan), a fork
+/// of [Hyperscan](https://github.com/intel/hyperscan), a high-performance multiple regex matching library.
+///
+/// The freshly-built `hs` library is then announced to rustc's linker for use in the compilation process.
+///
+/// This script can also build [Chimera](https://intel.github.io/hyperscan/dev-reference/chimera.html), which
+/// is an officially-supported regex engine that functions as a hybrid between Hyperscan and PCRE,
+/// with full support for PCRE syntax. While disabled by default, Chimera can be activated with feature `chimera`.
+///
+/// Lastly, `bindgen` is used to automatically create low-level bindings between Hyperscan's C ABI and Rust.
+/// These bindings are pre-generated at `src/bindings_hs.rs` and `src/bindings_ch.rs` for convenience (so
+/// that library consumers don't need to install multiple dependencies just to generate the required bindings).
+///
+/// However, if compiled with the `generate-bindings` feature, this crate will freshly generate bindings
+/// and save them to its `OUT_DIR`.
 fn main() {
     println!("cargo:rerun-if-changed=build.rs");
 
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -1,5 +1,11 @@
 [workspace]
 members = ["crates/*"]
+default-members = [
+    "crates/bins",
+    "crates/cli",
+    "crates/static-analysis-kernel",
+    "crates/static-analysis-server",
+]
```
juli1, line 8: This will be removed later once we figure out the build issues?

crates/vectorscan-sys/build.rs
```diff
@@ -0,0 +1,173 @@
...
+                    &dependency.source_path.to_string_lossy(),
+                    "--strip-components=1"
+                ])
+                .status()
+                .is_ok_and(|exit_status| exit_status.success()));
+        }
+    }
+
+    match env::var("CARGO_CFG_TARGET_OS").unwrap().as_str() {
+        "windows" => panic!("Vectorscan cannot be compiled for os: windows"),
```
juli1, line 65: Is this temporary or permanent? We need to make it work for WIndows (and have it available in the IDE, etc)

crates/vectorscan-sys/build.rs
```diff
@@ -0,0 +1,173 @@
...
+                .join("vendored-sources")
+                .join(pretty_name.as_ref()),
+            tarball_path: Path::new(env!("CARGO_MANIFEST_DIR"))
+                .join("vendor")
+                .join(tarball_name.as_ref()),
+        }
+    }
+}
+
+fn main() {
```
juli1, line 31: If possible, have a high-level comment of the commands being done.

crates/vectorscan-sys/build.rs
```diff
@@ -0,0 +1,173 @@
...
+        _ => println!("cargo:rustc-link-lib=stdc++"),
+    }
+
+    let mut hs_cmake = cmake::Config::new(hs_dep.source_path);
+    hs_cmake
+        // Override Vectorscan's defaults
+        .define("BUILD_EXAMPLES", "OFF")
+        .define("BUILD_BENCHMARKS", "OFF");
+
+    #[cfg(feature = "chimera")]
```
juli1, line 77: What is the `chimera` feature?

juli1 (commented):
This is very encouraging work! Some questions/comments as we are building this feature

1. Can we document (e.g. more comments in the code and `build.rs` files) the build system
2. Should we document more the bindings and the API for vectorscan, especially the end-user API?
3. I do not see a lot of tests and we should probably add more unit tests. Tests may not be easy to write for the low-level code and we should rather have more tests on higher-level code/functions
4. Can we fetch a git repository for `vectorscan` and `pcre` instead of putting blobs in the repo?

juli1 (approved):

PR #213, opened 2024-02-14
Running without static-analysis.datadog.yml file

juli1:

We want to be able to run the analyzer with the default rules, regardless if there is a `static-analysis.datadog.yml` or not.

## What is your solution?

Load all the default rules at startup-time by using the default ruleset endpoint and run the tool on all of them.
Also display a disclaimer if no file is passed and invite the user to read our documentation.

## Testing

 - [x] Add integration tests
 - [x] Add unit tests

### DIFF
--- a/crates/bins/src/bin/datadog-static-analyzer.rs
+++ b/crates/bins/src/bin/datadog-static-analyzer.rs
@@ -363,7 +363,9 @@ fn main() -> Result<()> {
 
     for language in &languages {
         let files_for_language = filter_files_for_language(&files_to_analyze, language);
-
+        if files_for_language.is_empty() {
+            continue;
+        }
         // we only use the progress bar when the debug mode is not active, otherwise, it puts
         // too much information on the screen.
         let progress_bar = if !configuration.use_debug {
--- a/crates/bins/src/bin/datadog-static-analyzer.rs
+++ b/crates/bins/src/bin/datadog-static-analyzer.rs
@@ -361,7 +361,6 @@ fn main() -> Result<()> {
     let files_filtered_by_size = filter_files_by_size(&files_to_analyze, &configuration);
 
     for language in &languages {
-
         let files_for_language = filter_files_for_language(&files_filtered_by_size, language);
 
         if files_for_language.is_empty() {
--- a/crates/bins/src/bin/datadog-static-analyzer.rs
+++ b/crates/bins/src/bin/datadog-static-analyzer.rs
@@ -1,5 +1,5 @@
 use cli::config_file::read_config_file;
-use cli::datadog_utils::get_rules_from_rulesets;
+use cli::datadog_utils::{get_all_default_rulesets, get_rules_from_rulesets};
 use cli::file_utils::{
     are_subdirectories_safe, filter_files_for_language, get_files, read_files_from_gitignore,
 };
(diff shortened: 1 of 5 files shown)


crates/cli/src/datadog_utils.rs
```diff
@@ -4,11 +4,21 @@ use anyhow::{anyhow, Result};
 use kernel::model::rule::Rule;
 use kernel::model::ruleset::RuleSet;
 
-use crate::model::datadog_api::ApiResponse;
+use crate::model::datadog_api::{ApiResponse, ApiResponseDefaultRuleset};
 
 const STAGING_DATADOG_SITE: &str = "datad0g.com";
 const DEFAULT_DATADOG_SITE: &str = "datadoghq.com";
 
+const DEFAULT_RULESETS_LANGAGES: &[&str] = &[
```
robertohuertasm, line 12: typo `LANGAGES`

crates/bins/src/bin/datadog-static-analyzer.rs
```diff
@@ -251,18 +251,28 @@ fn main() -> Result<()> {
         // if there is no config file, we must read the rules from a file.
         // Otherwise, we exit.
```
robertohuertasm, line 252: We don't exit now, so the comment may be Invalid

crates/bins/src/bin/datadog-static-analyzer.rs
```diff
@@ -251,18 +251,28 @@ fn main() -> Result<()> {
...
+                " - Datadog documentation: https://docs.datadoghq.com/code_analysis/static_analysis"
+            );
+            println!(" - Static analyzer repository on GitHub: https://github.com/DataDog/datadog-static-analyzer");
+            let rulesets_from_api =
+                get_all_default_rulesets(use_staging).expect("cannot get default rules");
+            let rules_from_api: Vec<Rule> = rulesets_from_api
+                .iter()
+                .flat_map(|v| v.rules.clone())
+                .collect();
+            rules.extend(rules_from_api);
```
robertohuertasm, line 266: This could probably be done like `rules.extend(rulesets_from_api.into_iter().flat_map(|v| v.rules));` thus avoiding cloning and collection.

crates/bins/src/bin/datadog-static-analyzer.rs
```diff
@@ -251,18 +251,28 @@ fn main() -> Result<()> {
...
+        } else {
+            let rulesets_from_file = get_rulesets_from_file(rules_file.clone().unwrap().as_str());
+            let rules_from_file: Vec<Rule> = rulesets_from_file
+                .context("cannot read ruleset")?
+                .iter()
+                .flat_map(|v| v.rules.clone())
+                .collect();
+            rules.extend(rules_from_file);
```
robertohuertasm, line 274: Similar to the previous comment, this could be: 
```rs
 rules.extend(
    rulesets_from_file
        .context("cannot read ruleset")?
        .into_iter()
        .flat_map(|v| v.rules),
);
```

crates/cli/src/datadog_utils.rs
```diff
@@ -4,11 +4,21 @@ use anyhow::{anyhow, Result};
...
+    "CSHARP",
+    "DOCKERFILE",
+    "GO",
+    "JAVA",
+    "JAVASCRIPT",
```
robertohuertasm, line 17: Is Typescript ommitted on purpose?

crates/bins/src/bin/datadog-static-analyzer.rs
juli1: updated, thank you!

crates/cli/src/datadog_utils.rs
juli1: Good point! Added!

crates/cli/src/datadog_utils.rs
juli1: Thank you, fixed!

crates/bins/src/bin/datadog-static-analyzer.rs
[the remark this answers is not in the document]
juli1: not really because we already put a warning when rules are pulled from the API

PR #228, opened 2024-02-23
[STAL-1418] Violation fingerprints

juli1:

We want to add a unique fingerprint to each violation in the sarif file. 

## What is your solution?

This is reported in the `partialFingerprint` in the SARIF file. This hash contains the content of the file and some other meta information to attempt to make it unique.

## Testing and Performance

Ran the tool on a large repo and disable the generation of fingerprints. Run on 1630 files for Go and YAML. The objective is to see the performance impact.

To disable fingerprinting, we just added `let fingerprint_option = None;` in the `sarif_utils.rs`

Then, built each version with `cargo build -r`

### Without fingerprints

```
Found 1500 violation(s) in 1630 file(s) using 105 rule(s) within 12 sec(s)
./target/release/datadog-static-analyzer --directory  --format sarif --output  57.55s user 5.52s system 435% cpu 14.486 total
```

### With fingerprints

```
Found 1500 violation(s) in 1630 file(s) using 105 rule(s) within 12 sec(s)
./target/release/datadog-static-analyzer --directory  --format sarif --output  56.25s user 5.35s system 447% cpu 13.756 total
```

### DIFF
--- a/crates/cli/src/file_utils.rs
+++ b/crates/cli/src/file_utils.rs
@@ -318,11 +318,12 @@ pub fn filter_files_by_diff_aware_info(
 ///  SHA2(<file-location-in-repository> - <characters-in-directory> - <content-of-code-line> - <number of characters in line>)
 pub fn get_fingerprint_for_violation(
     violation: &Violation,
-    repository_root: &str,
-    file: &str,
+    repository_root: &Path,
+    file: &Path,
     use_debug: bool,
 ) -> Option<String> {
-    let path = Path::new(repository_root).join(file);
+    let path = repository_root.join(file);
+    let filename = file.to_str().unwrap_or("");
     if !path.exists() || !path.is_file() {
         return None;
     }
@@ -337,8 +338,8 @@ pub fn get_fingerprint_for_violation(
                     .collect::<String>();
                 let hash_content = format!(
                     "{}-{}-{}-{}",
-                    file,
-                    file.len(),
+                    filename,
+                    filename.len(),
                     line_content_stripped,
                     line_content_stripped.len()
                 );
@@ -410,8 +411,8 @@ mod tests {
         let directory_string = d.into_os_string().into_string().unwrap();
         let fingerprint = get_fingerprint_for_violation(
             &violation,
-            &directory_string,
-            "resources/test/gitignore/test1",
+            Path::new(directory_string.as_str()),
+            Path::new("resources/test/gitignore/test1"),
             false,
         );
         assert!(!fingerprint.is_none());
@@ -422,8 +423,8 @@ mod tests {
 
         let fingerprint_unknown_file = get_fingerprint_for_violation(
             &violation,
-            &directory_string,
-            "path/does/not/exists",
+            Path::new(directory_string.as_str()),
+            Path::new("path/does/not/exists"),
             false,
         );
         assert!(fingerprint_unknown_file.is_none());
--- a/crates/cli/src/file_utils.rs
+++ b/crates/cli/src/file_utils.rs
@@ -1,14 +1,19 @@
-use crate::model::cli_configuration::CliConfiguration;
-use crate::model::config_file::PathConfig;
-use crate::model::datadog_api::DiffAwareData;
-use anyhow::Result;
-use glob_match::glob_match;
-use kernel::model::common::Language;
 use std::collections::HashSet;
 use std::fs;
 use std::fs::read_to_string;
 use std::path::{Path, PathBuf};
+
+use anyhow::Result;
+use glob_match::glob_match;
+use sha2::{Digest, Sha256};
 use walkdir::WalkDir;
+
+use kernel::model::common::Language;
+use kernel::model::violation::Violation;
+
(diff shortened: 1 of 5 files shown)


crates/cli/src/file_utils.rs
```diff
@@ -308,16 +313,65 @@ pub fn filter_files_by_diff_aware_info(
...
+    match read_to_string(&path) {
+        Ok(file_content) => match file_content.lines().nth(line - 1) {
+            Some(line_content) => {
+                let line_content_stripped = line_content.replace(' ', "");
+                let hash_content = format!(
+                    "{}-{}-{}-{}",
+                    file,
+                    file.len(),
+                    line_content_stripped,
+                    line_content_stripped.len()
```
dastrong, line 340: Why do we need the number of characters here and for the line above?

crates/cli/src/file_utils.rs
```diff
@@ -308,16 +313,65 @@ pub fn filter_files_by_diff_aware_info(
         .collect();
 }
 
+/// Generate a fingerprint for a violation that will uniquely identify the violation. The fingerprint is calculated
+/// as is
+///  SHA2(<file-location-in-repository> - <characters-in-directory> - <content-of-code-line> - <number of characters in line>)
```
dastrong, line 318: Why do we not use the rule ID?

juli1: Because the violation does not contain the rule id.

crates/cli/src/file_utils.rs
juli1: Just add more entropy for the hash.

crates/cli/src/file_utils.rs
juli1: no, it's passed to the function (see the function definition).

crates/cli/src/file_utils.rs
```diff
@@ -308,16 +313,65 @@ pub fn filter_files_by_diff_aware_info(
...
+pub fn get_fingerprint_for_violation(
```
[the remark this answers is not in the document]
juli1: I thought about this, especially for performance.

The problem is that `RuleResult` contains multiple violations. You want to distinguish between different violations, not have the fingerprint at the rule result level.  One workaround would be to pass the file content to the function and I think it does not separate concerns correctly.

When doing the benchmark on 1,000 of violations, you see there is no impact. The file content is already in your buffer-cache in your OS so you hit the memory and do not do any I/O (the content stays in memory).

I prefer maintainable code over performance (performance is cheap, software maintenance is expensive), which is why I prefer to stick with this unless we have a huge performance degradation.

crates/cli/src/file_utils.rs
juli1: So let's say you have the *exact* same line of code in two different files. In _some_ way, the filename turns out to have the _same_ hash. This length would just add another piece of info to try to distinguish between both.

I know it's borderline adding a tinfoil hat but that the reasoning behind it (e.g. add more stuff that could add more entropy). Happy to drop it if necessary.

crates/cli/src/file_utils.rs
```diff
@@ -308,16 +313,65 @@ pub fn filter_files_by_diff_aware_info(
...
+    let path = Path::new(repository_root).join(file);
+    if !path.exists() || !path.is_file() {
+        return None;
+    }
+    let line = violation.start.line as usize;
+
+    match read_to_string(&path) {
+        Ok(file_content) => match file_content.lines().nth(line - 1) {
+            Some(line_content) => {
+                let line_content_stripped = line_content.replace(' ', "");
```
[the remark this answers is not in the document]
juli1: For the former, that should not be an issue. The goal is if you add a space for style (e.g. your IDE changes the style), the violation is not new.

And yes, you are correct, fixing this.

crates/cli/src/file_utils.rs
juli1: yeah, this is why it's kind of a tinfoil hat justification.

crates/cli/src/file_utils.rs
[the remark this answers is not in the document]
juli1: yes, that would be even better then. Updating.

crates/cli/src/sarif/sarif_utils.rs
```diff
@@ -342,10 +345,26 @@ fn generate_results(
...
+                );
+
+                let partial_fingerprints: BTreeMap<String, String> =
+                    match (sha_option, fingerprint_option) {
+                        (Some(sha), Some(fp)) => BTreeMap::from([
+                            ("SHA".to_string(), sha),
+                            ("FINGERPRINT".to_string(), fp),
+                        ]),
+                        (None, Some(fp)) => {
+                            BTreeMap::from([("DATADOG_FINGERPRINT".to_string(), fp)])
```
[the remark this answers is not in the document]
juli1: Good catch, thank you! Fixed!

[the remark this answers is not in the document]
juli1: Added constants.

PR #235, opened 2024-02-27
Fix Stella console

juli1:

When in a rule, we do `console.log(var)`, it logs nothing, which is a bad user experience. We should log a string.

## What is your solution?

Force the `console.log` to log a string by prefixing the entry in the logs with `""`.

## Alternatives considered

Adding another dependency by using the `deno` console. I prefer not to introduce another dependency for now, this should do the work, especially since output is for debug only.

### DIFF
--- a/crates/static-analysis-kernel/src/analysis/javascript.rs
+++ b/crates/static-analysis-kernel/src/analysis/javascript.rs
@@ -526,6 +526,7 @@ def foo(arg1):
 
     #[test]
     fn test_execute_with_console() {
+        // Test for a string
         let q = r#"
 (function_definition
     name: (identifier) @name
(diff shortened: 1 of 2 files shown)


juli1:
> I imagined we would something a bit more explicit and type-aware to cast, instead of implicit:
> 
> ```js
> if Array.isArray(foo) {
>    // either JSON stringify or map + join on recursive function call
> } else if typeof foo === "object" {
>    return JSON.stringify(foo);
> } else if ... {
>    // at some point we can just do an explicit cast
>    return String(foo);
> }
> ```
> 
> Because if you do implicit (or even explicit e.g. `String(obj)`), you'll get the infamous `[object Object]`
> 
> Additionally, `Symbol` will throw a `TypeError` if we try to cast it implicitly, but not if it's cast explicitly.

If you really want to have this, I can do it, yes. I was just trying to solve the simple issue but I can go the extra mile.

PR #238, opened 2024-03-01
Separate the configuration file structs from the parsing functions.

jacobotb:
Future additions to the configuration file format will require extra processing to make the information contained in it usable by the program.

## What is your solution?
Separate all the serialization stuff into the src/config_file.rs file.


juli1:
I am not sure I understand the purpose of the changes. Is it possible to explain clearly the problem we are trying to solve and what is the objective?

PR #247, opened 2024-03-05
Add initial secrets-core crate

jasonforal:
# secrets-core
Adds the `secrets-core` crate. Some modules within (detailed below) are currently being refactored, and are not currently in this PR.

## Architecture/Overview
`secrets-core`: unopinionated and generic crate. It accepts arbitrary bytes, runs a chain of (arbitrary) matchers against those bytes, categorizes bytes into capture groups, runs (arbitrary) validators against those capture groups, and outputs the results, along with capture locations.
* Sample matchers: Hyperscan, Unix glob, Shannon Entropy, Edit Distance
* Sample validators: HTTP Request, Postgres URI, RSA Private Key
* Currently tied to a file source as the input source, but addition of other sources (e.g. Git commit messages, Git blobs) would be straightforward with this architecture.
* Under-the-hood, abstractions were designed assuming vectorized detectors (e.g. Hyperscan, Aho-Corasick), but it is not a requirement, and a lot of this complexity is hidden behind more traditional "single scan -> single pattern match" paradigm.
* Specific Rust implementation details (lifetimes, allocations, etc.) were chosen assuming (but not requiring) multi-threading.
* Matchers and Validators are modular, and are good candidates for future community open-source contribution.

Future crates:
* `secrets`: A consumer of `secrets-core` that will provide all of our application-specific logic: file I/O, YAML rule parsing, SARIF output, Rulesets.
* `secrets-cli`: a standalone wrapper of `secrets` for a binary that only runs the scanner (enabling integrations like Git hooks). Not immediately planned, but listed here for completeness.

# What the reviewer should know
* This code is still "inert" in the sense that it will not compile by default in CI. It can still be built/tested/linted by passing `package` to cargo: e.g. `cargo build --package=secrets-core`
  * The motivation for this is to delay the refactor of the CI build, especially given there are no integration tests yet.
* Four modules are currently in a state of refactor, and are not included in this PR but _will be_ a part of secrets-core:
  * `engine`: The entrypoint for any consumer of the library. It's a lightweight orchestrator that takes a configuration and spawns workers to execute that configuration.
  * `evaluator`: A stateful executor that executes a rule's match logic.
  * `worker`: A lightweight, semi-stateful orchestrator that handles the glue/pass-off between matcher and validator.
  * `glob`: A glob matcher for file paths.
* The HTTP validator workflow lacks test coverage. This and further tests around validation will come in a separate PR. I need to think through the right mix of unit/integration tests.
* In a separate PR, the library API will expose hooks for library consumers, e.g. a function to create a generic HTTP validator with a default config, etc. This is evolving in parallel with my implementation of the `secrets` crate.

# Addendum

Happy to keep the PR open if you'd like to wait for the aforementioned modules to land.

### DIFF
--- a/crates/secrets-core/src/lib.rs
+++ b/crates/secrets-core/src/lib.rs
@@ -2,6 +2,11 @@
 // This product includes software developed at Datadog (https://www.datadoghq.com/).
 // Copyright 2024 Datadog, Inc.
 
+// NOTE: Because this crate still has some scaffolding, these ignores are added
+// to (temporarily) silence clippy while iterating on the `pub` interface to expose.
+#![allow(unused_imports)]
+#![allow(dead_code)]
+
 mod capture;
 mod common;
 mod location;


crates/secrets-core/src/capture.rs
```diff
@@ -0,0 +1,313 @@
...
+        // invariant in debug mode.
+        debug_assert!(name_slots
+            .as_ref()
+            .map(|name_slots| name_slots.len() == capture_slots.len())
+            .unwrap_or(true));
+        Self {
+            name_slots,
+            capture_slots,
+        }
+    }
```
juli1, line 35: Just curious, why not using [deriver_builder](https://crates.io/crates/derive_builder) for this?

crates/secrets-core/src/common.rs
```diff
@@ -0,0 +1,62 @@
...
+}
+
+impl ByteSpan {
+    /// Creates a ByteSpan given a start and end index.
+    pub fn new(start_index: usize, end_index: usize) -> Self {
+        Self {
+            start_index: start_index as u32,
+            end_index: end_index as u32,
+        }
+    }
```
juli1, line 21: Same question about the builder

crates/secrets-core/src/location.rs
```diff
@@ -0,0 +1,553 @@
...
+use bstr::{BStr, ByteSlice};
+use std::cell::{Cell, RefCell};
+use std::num::NonZeroU32;
+
+/// A one-based point representing a line and column in a string.
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash)]
+pub(crate) struct Point {
+    pub line: NonZeroU32,
+    pub col: NonZeroU32,
+}
```
juli1, line 15: Why not reusing the `Position` from the [common model](https://github.com/DataDog/datadog-static-analyzer/blob/main/crates/static-analysis-kernel/src/model/common.rs#L96)?

juli1 (approved):

PR #250, opened 2024-03-06
Replace glob_match with globset

jacobotb:
The `glob_match` crate has a bug with a pattern like `**/*test*.py` , where a file like `abc/test/def/test.py` does not match.

Or even `t/test.py` does not match.

See https://github.com/devongovett/glob-match/issues/9 for issue.

## What is your solution?
The `globset` crate does not have that bug and is pretty fast too.

`globset` has a separate compilation step, so I do it during configuration parsing.

### DIFF
--- a/crates/bins/src/bin/datadog-static-analyzer.rs
+++ b/crates/bins/src/bin/datadog-static-analyzer.rs
@@ -55,20 +55,10 @@ fn print_configuration(configuration: &CliConfiguration) {
     let ignore_paths_str = if configuration.path_config.ignore.is_empty() {
         "no ignore path".to_string()
     } else {
-        configuration
-            .path_config
-            .ignore
-            .iter()
-            .map(|p| p.clone().into())
-            .collect::<Vec<String>>()
-            .join(",")
+        configuration.path_config.ignore.join(",")
     };
     let only_paths_str = match &configuration.path_config.only {
-        Some(x) => x
-            .iter()
-            .map(|p| p.clone().into())
-            .collect::<Vec<String>>()
-            .join(","),
+        Some(x) => x.join(","),
         None => "all paths".to_string(),
     };
 
--- a/crates/bins/src/bin/datadog-static-analyzer.rs
+++ b/crates/bins/src/bin/datadog-static-analyzer.rs
@@ -55,10 +55,20 @@ fn print_configuration(configuration: &CliConfiguration) {
     let ignore_paths_str = if configuration.path_config.ignore.is_empty() {
         "no ignore path".to_string()
     } else {
-        configuration.path_config.ignore.join(",")
+        configuration
+            .path_config
+            .ignore
+            .iter()
+            .map(|p| p.clone().into())
+            .collect::<Vec<String>>()
+            .join(",")
     };
     let only_paths_str = match &configuration.path_config.only {
-        Some(x) => x.join(","),
+        Some(x) => x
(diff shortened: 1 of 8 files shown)


crates/bins/src/bin/datadog-static-analyzer.rs
```diff
@@ -55,10 +55,20 @@ fn print_configuration(configuration: &CliConfiguration) {
...
+            .collect::<Vec<String>>()
+            .join(",")
     };
     let only_paths_str = match &configuration.path_config.only {
-        Some(x) => x.join(","),
+        Some(x) => x
+            .iter()
+            .map(|p| p.clone().into())
+            .collect::<Vec<String>>()
+            .join(","),
```
juli1, line 71: Curious why this change? Seems more complicated to read?

crates/bins/src/bin/datadog-static-analyzer.rs
```diff
@@ -55,10 +55,20 @@ fn print_configuration(configuration: &CliConfiguration) {
...
         "no ignore path".to_string()
     } else {
-        configuration.path_config.ignore.join(",")
+        configuration
+            .path_config
+            .ignore
+            .iter()
+            .map(|p| p.clone().into())
+            .collect::<Vec<String>>()
+            .join(",")
```
juli1, line 64: Curious as to why this change? Seems more complicated to read

crates/cli/src/model/config_file.rs
```diff
@@ -1,19 +1,28 @@
...
 use serde;
 use serde::{Deserialize, Serialize};
 
+#[derive(Deserialize, Serialize, Debug, Default, Clone)]
+#[serde(from = "String", into = "String")]
+pub struct PathPattern {
+    pub glob: Option<GlobMatcher>,
+    pub prefix: PathBuf,
+}
+
```
juli1, line 17: Can we document the use of this `struct`? A comment to explain what is it used for?

crates/bins/src/bin/datadog-static-analyzer.rs
jacobotb: Because that was the way I knew how to coerce a vector of PathPattern into a vector of String. Now I used some Rust 🪄 to solve it.

crates/cli/src/model/config_file.rs
jacobotb: Done.

crates/cli/src/file_utils.rs
```diff
@@ -154,18 +153,19 @@ pub fn get_files(
...
-        return true;
-    }
-    return Path::new(path).starts_with(Path::new(pattern));
+fn matches_pattern(pattern: &PathPattern, path: &str) -> bool {
+    pattern
+        .glob
+        .as_ref()
+        .map(|g| g.is_match(path))
+        .unwrap_or(false)
+        || Path::new(path).starts_with(&pattern.prefix)
```
juli1, line 162: Can we implement this as part of the `PathPattern` struct?

```
impl PathPattern {
   fn match(path: &str) {
      ....
   }
}

This way, we made sure we tie/attach this function to each type? This occurred to me after reading the clarification about the `PathPattern`, it seems this function would be part of the "type".

juli1 (approved):

juli1:
Approved and agreed on a follow up to make the function as part of a type.

PR #255, opened 2024-03-07
[STAL-1496] add option to fail if there is any violation

juli1:

The problem is reported by @yoannmoinet: we want to have an option to exit a non-zero code if there is at least one violation. 

## What is your solution?

We offer the option for different severity.

- `--fail-on-any-violation=<list-of-severities>` fail if there is at least one violation with a severity from the list of `<list-of-severities>`

## Testing

Added unit tests for the counting as well as an integration tests.

See below for manual testing.

Incorrect severity

```
datadog-static-analyzer --directory /Users/julien.delange/git/rosie-tests --output /tmp/plop.plop --fail-on-any-violation=none,warning.notice,error        14:02:18
thread 'main' panicked at crates/bins/src/bin/datadog-static-analyzer.rs:208:48:
cannot map severity: "unknown severity"
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Option with correct severities

```
$ datadog-static-analyzer --directory /Users/julien.delange/git/rosie-tests --output /tmp/plop.plop --fail-on-any-violation=none,warning,notice,error        
...
Found 8 violation(s) in 8 file(s) using 126 rule(s) within 0 sec(s)
$ echo $? 
1
```

Removing the option

```
datadog-static-analyzer --directory /Users/julien.delange/git/rosie-tests --output /tmp/plop.plop  
Configuration
...
Found 8 violation(s) in 8 file(s) using 126 rule(s) within 1 sec(s)
$ echo $?
0
```

### DIFF
--- a/crates/bins/src/bin/datadog-static-analyzer.rs
+++ b/crates/bins/src/bin/datadog-static-analyzer.rs
@@ -6,13 +6,15 @@ use cli::file_utils::{
     are_subdirectories_safe, filter_files_by_diff_aware_info, filter_files_by_size,
     filter_files_for_language, get_files, read_files_from_gitignore,
 };
-use cli::rule_utils::{get_languages_for_rules, get_rulesets_from_file};
+use cli::rule_utils::{
+    count_violations_by_severities, get_languages_for_rules, get_rulesets_from_file,
+};
 use itertools::Itertools;
 use kernel::analysis::analyze::analyze;
 use kernel::constants::{CARGO_VERSION, VERSION};
 use kernel::model::analysis::{AnalysisOptions, ERROR_RULE_TIMEOUT};
 use kernel::model::common::OutputFormat;
-use kernel::model::rule::{Rule, RuleInternal, RuleResult};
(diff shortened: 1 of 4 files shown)


crates/bins/src/bin/datadog-static-analyzer.rs
```diff
@@ -128,6 +131,26 @@ fn main() -> Result<()> {
         "print-violations",
         "print a list with all the violations that were found",
     );
+    opts.optflag(
+        "",
+        "fail-on-any-violation-error",
+        "exit a non-zero return code if there is one violation with the error severity",
+    );
+    opts.optflag(
```
jacobotb, line 139: Normally, this kind of thing is an enum flag, like

`--fail-on-violation=error`

I think it would be easier and more obvious for the user if there were only one flag that works like that.

juli1: yes, I was thinking about this and I am changing this. Will follow up with another PR

jacobotb (approved):

PR #257, opened 2024-03-08
Validator for the static analyzer configuration parser

jacobotb:
We now have a JSON Schema that defines the valid syntax, but the static analyzer configuration parser is more permissive than that. This could lead to situations where the user has a configuration that is accepted by the static analyzer but not by the IDE.

## What is your solution?
Build a tool to validate configurations and scripts to execute it against the examples that come with the JSON Schema, then adjust the parser so it will accept the valid examples and reject the invalid examples.

Add the script to the on-push tests so that changes that change the parser's behavior are rejected unless the examples and schema are modified too.

Some existing configuration files may turn out to be invalid. In fact, some unit and integration tests had to be fixed.

This is limited to configuration files that were modified to use `ignore`/`only`, and that use the "map" syntax, so not many of them.


juli1:
What is the difference between this binary and using [pajv](https://www.npmjs.com/package/pajv)?

jacobotb:
This binary checks that our parser is conformant with the schema.

Pajv checks one configuration; this binary checks our parser.

juli1:
> This binary checks that our parser is conformant with the schema.
> 
> Pajv checks one configuration; this binary checks our parser.

Thank you for clarifying, it makes sense since we want to check the parser validity. Should we have a binary or rather just have this as unit tests that will still be executed at each commit? I am trying to understand the need for a new binary: the goal is for testing, we could implement this with unit tests?

jacobotb:
I originally thought of implementing with unit tests, but I felt icky at the idea of having tests that reach up into the source tree (so they read files in `../../schema/examples`.)

juli1:
> I originally thought of implementing with unit tests, but I felt icky at the idea of having tests that reach up into the source tree (so they read files in `../../schema/examples`.)

We already have unit tests that do this (see [file_utils.rs](https://github.com/DataDog/datadog-static-analyzer/blob/main/crates/cli/src/file_utils.rs#L357) for example). It is "common" to have file to test, especially for configuration file, parsing `.gitignore`, etc. So we can totally bring this as unit tests.

I gave more thought and probably both could be useful (sorry for the long justification below).

 - **Using unit tests first**: unit test runs at each push/commit. They allow us to have quick iteration (really good when we will change the file format). Those tests could be unit tests.
 - **Integration tests**: we could also add integration tests but in this case, we should probably add the integration tests on our binary (e.g. run the static analyzer on an invalid configuration). An integration tests ensures that the *actual code* run by the customer (e.g. on `datadog-static-analyzer`). As is, with a new binary, there is no guarantee the binary we uses the same code as the cli (it is in this commit but there is no guarantee). By running the integration tests on the cli, we guarantee we test what is shipped to the customer.

Also, we could avoid having another binary to compile (save compilation time at each CI), especially if we do not ship it to the customer.

Sorry for the long explanation/details, hope it clarifies that we can do unit test with files and/or how we could have the integration tests for this.

jacobotb:
Ah, but for those tests that read files, the files are in a subdirectory of the directory where the test is, not in `../../` :-)

I'll change that to a unit test, anyway.

juli1 (approved):
As a follow-up, maybe we can also add some integration test with the binary and when an invalid configuration file is passed, we checked the binary returns a given error code (e.g. `42` or something like this). Not a blocker for this PR but something else we can add.

PR #261, opened 2024-03-11
Add secrets-core engine

jasonforal:
Adds modules:

### rule_evaluator
Given an input source, executes a sequence of `Matcher` rules with arbitrary, caller-specified boolean logic.
**Notes**
* The complexity in the `Scanner` iterator's `ScannerState` cache is slightly unavoidable: a vectorized `Matcher` like Hyperscan can produce matches for multiple rules in a single scan, but we want to expose a more intuitive single-rule API. Thus, we need to handle the caching behind-the-scenes so that multiple calls to query the same data for different rules won't trigger a rescans.
* The caches here will be hit in a hot loop, so they are designed for high performance.

### worker
A lightweight driver of a `RuleEvaluator`. Converts all identified secret candidates to owned data and attaches their line/col location.

### engine
The entrypoint for consumers of the library. Creates workers. Exposes a `scan_file` function, which reads a single file and runs all loaded rules against it. The library itself does not force multi-threading on the consumer, but a `rayon` threadpool can easily call this function in a parallel iterator.

---

#### Reviewer notes
* Validation will come in a follow-up PR

### DIFF
--- a/crates/secrets-core/src/capture.rs
+++ b/crates/secrets-core/src/capture.rs
@@ -181,6 +181,7 @@ impl<'b> CaptureSlots<'b> {
     }
 
     /// Returns the number of capture slots.
+    #[allow(clippy::len_without_is_empty)]
     pub fn len(&self) -> usize {
         self.0.len()
     }
--- a/crates/secrets-core/src/common.rs
+++ b/crates/secrets-core/src/common.rs
@@ -38,6 +38,7 @@ impl ByteSpan {
     }
 
     /// Returns the length represented by the ByteSpan
+    #[allow(clippy::len_without_is_empty)]
     #[inline]
     pub fn len(&self) -> usize {
         debug_assert!(self.end_index >= self.start_index);


crates/secrets-core/src/engine.rs
```diff
@@ -0,0 +1,133 @@
...
+    Worker(#[from] WorkerError),
+}
+
+pub struct Engine<T: ThreadPool> {
+    thread_pool: T,
+    rules: Vec<Rule>,
+    matchers: Vec<Matcher>,
+    validators: Vec<Arc<Mutex<Box<dyn Validator + Send + Sync>>>>,
+    run_validation: bool,
+}
```
juli1, line 34: Why use a custom `ThreadPool` and start to put our finger into concurrency issues when it's a solved problem with rayon and other libraries? Ideally, this should run in the same environment than the static analyzer and this should just be something we can parallelize with rayon. There is a good change that starting our own (untested) concurrency implementation will lead to many bugs down the line.

juli1, line 34: Why there is the concept of an engine? We should only have a series of rules and the rules executes sequentially or in parallel. The matcher should be in the rule, not in the engine.

juli1 (commented):
Commented on the first few aspects. I am not sure we should have our own concurrency. The rest of the analyzer is concurrency agnostic and we should probably implement the secrets the same way.

Why not starting without having all concurrent aspects for now and leaving this to rayon?

juli1 (approved):

PR #276, opened 2024-03-18
[STAL-1322] Support Vectorscan build for Windows

jasonforal:
* `vectorscan-sys` currently does not build on Windows

## What is your solution?
* Updates to `build.rs`
  * Vectorscan will officially support Windows in their next release, but in the mean time, on Windows, build [Hyperscan 5.4.2](https://github.com/intel/hyperscan/releases/tag/v5.4.2). This works because Vectorscan has not changed its ABI, so the two are interchangeable.
  * Disable compilation of Hyperscan's "tools" by default (we don't use these, and they require pulling in `pcre`).
* Updates to GitHub actions
  * OS-specific setup is implemented as a separate, "composite action"
  * `vectorscan-sys` can take 15+ minutes to build, so a build cache was implemented. A separate, tailored step via `actions/cache` is required because `Swatinem/rust-cache` does not cache workspace crates, only third party dependencies.
  * Integration tests were DRYed up via a matrix (motivated by not wanting to duplicate the setup 5x times)
  
  ## After
  You can view https://github.com/DataDog/datadog-static-analyzer/commit/2fc675f3121cac8f9ab3ec8104d1b3c97e7dec6e, which contains the secrets scanning integration that builds on top of this branch's build refactor
  https://github.com/DataDog/datadog-static-analyzer/actions/runs/8332242705
Screenshot 2024-03-18 at 12 43 03

## Alternatives considered
  * A single `setup.sh` file that handles the `os.runner` branching, and only a single step that calls the `setup.sh`. See below for rationale to skip this for now.

* Simplifying the Windows setup to a single step script like Linux/macOS was taking me too much time:
  * I kept running into PATH/environment variable issues when it came to boost, vcpkg, and cmake.
  * Building of Boost alone took ~8 minutes -- the third-party actions use a cached build.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -36,8 +36,8 @@ jobs:
             os: windows-latest
     runs-on: ${{ matrix.os }}
     steps:
-      - uses: ./.github/workflows/setup-build-environment.yaml
       - uses: actions/checkout@v3
+      - uses: ./.github/actions/setup-build-env
       - name: Install Stable + Rustfmt + Clippy
         uses: actions-rs/toolchain@v1
         with:


crates/vectorscan-sys/src/hyperscan_bindings_hs.rs
```diff
@@ -0,0 +1,739 @@
+/* automatically generated by rust-bindgen 0.69.4 */
```
juli1, line 1: How do we generate these files? Can we put instructions somewhere?

juli1 (approved):

PR #289, opened 2024-03-27
feat(server): add IDE module

robertohuertasm:

The IDEs need to handle static configuration file changes for several reasons:
- onboarding
- ignoring rules globally
- future requirements

## What is your solution?

The solution is to add 3 endpoints to the server as per this [RFC](https://docs.google.com/document/d/116yeGE9zC144aPD37F65ep15qMb70Yr0tnJ1g8ejGN0/edit#heading=h.99wkb9gyb761).

As agreed on that same [RFC](https://docs.google.com/document/d/116yeGE9zC144aPD37F65ep15qMb70Yr0tnJ1g8ejGN0/edit#heading=h.99wkb9gyb761), this PR introduces a new `ide` module which provides the before mentioned 3 endpoints:
- POST rulesets: will add some rulesets to the configuration file and return the content as a string.
- GET rulesets: will return the list of rulesets contained by the configuration file.
- POST ignore-rule: will ignore a rule in the configuration file and return the content as a string.

The `ide` folder has a defined ownership via `CODEOWNERS` file.

We have also introduced in there a `requests.http` file to help while debugging/testing.

Other things to consider are:

We needed some dependencies that were not included in the `bins` Cargo.toml:
- `thiserror`: for custom error support.
- `serde` & `serde_yaml`: for serialization purposes and error handling.
- `indexmap`: for the onboarding process, to insert a new collection (we need ordering).

These dependencies have been added to the `bins` Cargo.toml, and also some of them have been moved to the workspace Cargo.toml file, as they are now shared by several crates (indexmap & serde_yaml).

The `CustomHeader` fairing has been modified to avoid calling the `get_version` and `get_revision` endpoints and just extracted that logic to the `utils` module. This avoids calling these endpoints for every requests. This can be useful for accurately tracing the number of calls to these endpoints while debugging.

Finally, we added a `Makefile.toml` file to the root of the project so we can use [cargo make](https://github.com/sagiegurari/cargo-make) which makes it easier to automate several tasks:
- start the server in watch mode
- test the ide related tests in watch mode
- run format or clippy with different configuration (now we only have defaults)
- checking unused dependencies

[cargo make](https://github.com/sagiegurari/cargo-make) is super-useful and works for all the operating systems. You can execute even rust scripts or cross platform shell scripts.

_VS Code related change_: we changed the port from 9090 to 49159 in the `launch.json` file to make it easier for us to debug the binary from the extension.


robertohuertasm:
@DataDog/static-analysis-core I think we should explicitly add the @DataDog/ide-integration to this repo for the it to be notified as a codeowner.

juli1:
> @DataDog/static-analysis-core I think we should explicitly add the @DataDog/ide-integration to this repo for the it to be notified as a codeowner.

Just added it :)

Makefile.toml
```diff
@@ -0,0 +1,45 @@
+[tasks.server-dev]
```
juli1, line 1: What is the value/use of the `Makefile.toml`?

juli1:
Added @jasonforal as a reviewer

Makefile.toml
robertohuertasm: It's explained in the description of the PR. But it's basically a convenient tool to use to automate some repetitive tasks. In this case, this task is to start the server in watch mode in a particular port. So devs don't have to remember or copy paste a long command. Just do `cargo make server-dev`. Or if you want to run the tests in watch mode (in this case for the IDE side of the code), we would just do `cargo make ide-test` or `makers ide-test` (both ways are ok).

One of the nice things about this tool compare to Make files is that it's cross-platform and it's rust based.

juli1: Can we have this outside of the repo? Seems something nice to have as a developer to test but probably not something we should commit in the repo (and added to `.gitignore`). @jasonforal any thoughts on this?

robertohuertasm: We could have it ignored, that's always an option. But one of the benefits is to provide shared scripts. Another option is to host that file in the IDE folder. That would make things a little less convenient as we would have to use it like this: `makers --makefile crates/bins/src/bin/datadog_static_analyzer_server/ide/Makefile.toml server-dev` instead of just `makers server-dev` but it would work.

robertohuertasm: Let me know what's your preference and I'll move the file if needed.

crates/bins/src/bin/datadog_static_analyzer_server/ide/configuration_file/endpoints.rs
```diff
@@ -0,0 +1,63 @@
...
+        rulesets=?&rulesets,
+        content=&configuration_base64
+    );
+    let result = StaticAnalysisConfigFile::with_added_rulesets(&rulesets, configuration_base64);
+    to_json_result(result, encoded)
+}
+
+#[instrument()]
+#[rocket::get("/v1/config/rulesets/<content>", format = "application/json")]
+pub fn get_rulesets(content: &str) -> Json<Vec<String>> {
```
albertvaka, line 50: If the file is incorrectly formatted, this endpoint should be able to return an error (with 500 status code), so we can tell the user that SA is disabled because the config file is invalid. Otherwise, they don't have any way to know.

An alternative would be to add a separate `/config/validate` endpoint. No strong preference.

crates/bins/src/bin/datadog_static_analyzer_server/ide/configuration_file/endpoints.rs
```diff
@@ -0,0 +1,63 @@
...
+        ..
+    } = request.into_inner();
+    tracing::debug!(rule, content = &configuration_base64);
+    let result = StaticAnalysisConfigFile::with_ignored_rule(rule.into(), configuration_base64);
+    to_json_result(result, encoded)
+}
+
+#[instrument()]
+#[rocket::post("/v1/config/rulesets", format = "application/json", data = "<request>")]
+pub fn post_rulesets(
```
albertvaka, line 31: There are cases where we shouldn't offer to onboard the user to a new languge, eg: if the config file uses a version greater than the current known version, if there are [exclusions applied](https://docs.google.com/document/d/1jjloCvTOAvK99WDbfdcnttjYtjq2MJfPE8YKvDv432s/edit#heading=h.ndmldug56n2w) or if the file is invalid. 

We might need a new `can_onboard` endpoint for that.

crates/bins/src/bin/datadog_static_analyzer_server/ide/configuration_file/endpoints.rs
```diff
@@ -0,0 +1,63 @@
...
+}
+
+#[instrument()]
+#[rocket::get("/v1/config/rulesets/<content>", format = "application/json")]
+pub fn get_rulesets(content: &str) -> Json<Vec<String>> {
+    tracing::debug!(%content);
+    Json(StaticAnalysisConfigFile::to_rulesets(content.to_string()))
+}
+
+fn to_json_result(
```
albertvaka, line 55: Not for this PR, but in a future, we should improve this so it doesn't re-serialize the whole file. Since this is a file that the users might modify manually, by re-writing it we are going to be overwriting comments, specific formatting, the order of the entries, etc. We should be able to preserve that.

albertvaka (approved):
LGTM. I've left some comments on things we might want to improve to provide a better user experience, specially for errors handling (since the IDEs are the only UI of the static analyser the users can see, we shouldn't silently ignore errors), but they shouldn't be blocking and we can work on them on a separate PR.

crates/bins/src/bin/datadog_static_analyzer_server/ide/configuration_file/endpoints.rs
robertohuertasm: Yes, we could extend it like that. I didn't go as far as that for this PR because I thought that if the endpoint returned an empty collection that would implicitly mean that the parsing went wrong. The IDE knows that the file is there, hence the only option is that either the file is empty or is bad formatted. In both cases, it could be seen as an invalid config file.

crates/bins/src/bin/datadog_static_analyzer_server/ide/configuration_file/endpoints.rs
robertohuertasm: We might, I agree. Although we could also consider other options like expanding the `get_rulesets` returning collection with that information.

robertohuertasm:
@albertvaka not sure about the comment of ignoring errors, given that out of 3 endpoints 2 expose errors. The reasons for the third one to not expose it is that it could be inferred just by the return of an empty collection.

Makefile.toml
robertohuertasm: Fair enough. I'll move it to the ide folder as we can then use it there, either by moving the cwd there or by using an alias to not have to type the --makefile path. I thought that it could be convenient for your team to maybe use it for some of the repetitive tasks you have. But I may be biased as I'm very fond of this tool, and use it in almost every project. Not only for complex tasks but even for trivial ones. For instance, in the `udeps` tasks it will automatically change the toolchain to nightly, install the udeps crate for you and run it. I moved it already 😀!

crates/bins/src/bin/datadog_static_analyzer_server/ide/configuration_file/endpoints.rs
robertohuertasm: Note: After discussing about this in one of our RFCs, @albertvaka pointed out that it would be better to avoid conflating rulesets and the fact of being able to onboard, because this can change in the future.

robertohuertasm:
@jasonforal @albertvaka added a new endpoint to cover a new case for the onboarding where we will ignore onboarding if any path has been defined at root level of the config file.

crates/bins/src/bin/datadog_static_analyzer_server/ide/configuration_file/endpoints.rs
albertvaka: I still find it a bit cleaner to return an error status code, although your solution works. It however limits us in a future if we want to eg make this function only return the rulesets that apply to a file, after filtering the ignores/only. But no strong opinion, we could also bump the api version later if we implement that behavior, so up to you.

robertohuertasm: I think I'll leave it as it is right now and evolve it as needed. The case you're picturing, I guess we would have to change the signature anyways because we would need to pass the file path or something. So that would make us bump either the version, or think of a different endpoint. Is that right? If not, it's not it's not too much work to change it.

PR #296, opened 2024-03-29
[STAL-1493] Add ability to use a fake clock for tests

jasonforal:
To have proper coverage of the HTTP validator, we need to be able to test the retry, request timeout, and rate-limiting behavior.

## What is your solution?
* Introduce a `MockClock` for the test environment, which can be manually advanced to simulate time passing.
  * The `retry` crate hardcodes `thread::sleep`, so it would require a complicated shim to keep using this crate and use mocked time. Because I was already fighting against this crate's API design for unrelated reasons, it made more sense to remove the dependency and implement the core retry logic ourselves.
* The `governor` rate limiter has its own abstraction over time, so this PR also implements governor's "Clock" trait for our `MockClock`.
* Unrelated chore: Fix clippy violations from Rust 1.77

## Alternatives considered
* Using `thread::sleep` with smaller durations (e.g. 500ms) instead of mocking time.

* This PR has been sliced off from a larger, more fully-featured branch. So this code has been thoroughly unit tested, but the unit tests were omitted in this PR because they rely on extra code that will land in separate PRs.

### DIFF
--- a/crates/secrets-core/src/validator/http.rs
+++ b/crates/secrets-core/src/validator/http.rs
@@ -126,10 +126,12 @@ struct RequestGenerator {
 
 /// A wrapper around [`thread::sleep`](std::thread::sleep) that advances a [`MockClock`](crate::common::time::MockClock)
 /// when running tests.
+#[cfg(test)]
 fn thread_sleep(duration: Duration) {
-    #[cfg(test)]
     time::MockClock::advance(duration);
-    #[cfg(not(test))]
+}
+#[cfg(not(test))]
+fn thread_sleep(duration: Duration) {
     std::thread::sleep(duration);
 }
 


crates/secrets-core/src/validator/http.rs
```diff
@@ -53,27 +55,41 @@ pub(crate) enum RetryPolicy {
     },
 }
 
-type DynFnResponseParser =
-    dyn Fn(Result<ureq::Response, ureq::Error>) -> Result<NextAction, HttpValidatorError>;
+type DynFnResponseParser = dyn Fn(&Result<ureq::Response, ureq::Error>) -> NextAction;
 
-pub struct HttpValidator {
+type RateLimiter<T> = governor::RateLimiter<
```
juli1, line 60: Can you put a comment that explains what the rate limiter does and how it is used across the codebase?

crates/secrets-core/src/validator/http.rs
```diff
@@ -53,27 +55,41 @@ pub(crate) enum RetryPolicy {
...
+    governor::state::NotKeyed,
+    governor::state::InMemoryState,
+    T,
+    NoOpMiddleware<<T as Clock>::Instant>,
+>;
+
+pub struct HttpValidator<T: Clock = MonotonicClock> {
```
juli1, line 67: Similar: have documentation about this struct, what it does and how it works. Important for `pub` data structure.

juli1 (approved):

PR #297, opened 2024-03-29
[STAL-1493] Add builders for HTTP validator

jasonforal:
Validators are designed to be constructed at runtime. `secrets-core` currently lacks builders that library consumers can use to ergonomically build validators.

The `secrets` crate will use these builders to construct validators from YAML definitions.

### Builders for runtime HTTP validator creation
Library users can build a validator at runtime that generates a unique-per-secret HTTP request:
```rust
let url_generator = Box::new(|candidate: &Candidate| format!("https://example.com?token={}", candidate.text()));
let header_generator = Box::new(|candidate: &Candidate| format!("Bearer {}", candidate.text()));
let request_generator = RequestGeneratorBuilder::http_get(Agent::new(), url_generator)
    .header("Accept", "text/plain")
    .dynamic_header("Authorization", header_generator)
    .build();
```

Likewise, they can use a builder to create a response handler:
```rust
let response_handler = ResponseParserBuilder::new()
    .on_status_code(200, NextAction::ReturnResult(SecretCategory::Valid(Severity::Error)))
    .on_status_code(401, NextAction::ReturnResult(SecretCategory::Invalid))
    .on_status_code(404, NextAction::Abort)
    .build();
```

And put this all together to build the entire HTTP validator:
```rust
let validator = HttpValidatorBuilder::new("http".into(), "my-rule".into(), request_generator, response_handler)
    .retry_config(/* ... */)
    .max_attempt_duration(/* ... */)
    .rate_limit(/* ... */)
    .request_timeout(/* ... */)
    // ...
    .build();
```

### Unit tests
This PR additionally adds the missing HTTP request unit tests.

## Alternatives considered
* Small durations of "real" time (e.g. 500ms) in unit tests.

* All requests are made with `Datadog/StaticAnalyzer` as a user agent.
* HTTP timeout logic is hard to unit test without using "real" time. It will be tested via a future integration test.
* xorshift jitter was implemented manually to avoid pulling in a third-party dependency for something so small.

### DIFF
--- a/crates/secrets-core/src/validator/http.rs
+++ b/crates/secrets-core/src/validator/http.rs
@@ -262,6 +262,8 @@ pub struct HttpValidator<T: Clock = MonotonicClock> {
     response_parser: Box<DynFnResponseParser>,
     /// A function that generates an iterator of [`Duration`] that implements this validator's retry policy.
     backoff_generator: Box<DynFnBackoffGenerator>,
+    /// The duration the validator will wait for a server response before considering a request failed.
+    request_timeout: Duration,
 }
 
 /// The next action to take after an HTTP request has received a response.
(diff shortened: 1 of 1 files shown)


crates/secrets-core/src/validator/http.rs
```diff
@@ -267,6 +479,564 @@ impl Display for HttpMethod {
     }
 }
 
+const DEFAULT_REQ_PER_SECOND: u32 = 50;
```
juli1, line 482: This is the default requests per seconds max?

juli1 (approved):

PR #298, opened 2024-03-29
[STAL-1493] Cache results of HTTP requests

jasonforal:
The HTTP validator makes duplicate requests, even if we already know if a given text is a valid or invalid secret.

## What is your solution?
* Cache the results of HTTP validation when possible. This cache is shared amongst all threads.
* The essential properties of each request are hashed and used as a 128 bit key. By using a hash as a key instead of plaintext, we avoid storing potential secrets in memory longer than necessary.
* Multiple threads could be racing on the same request, so at every return point, the cached value is used (or freshly stored). When a thread times out due to its retry limit, no result is cached.

* `xxhash-rust` was chosen for low lines of code, no dependencies, and the ability to generate a 128-bit hash. A hash larger than 64 bits is desirable because we never store the actual secret in memory (and thus can't compare equality in case of a hash collision)

### DIFF
--- a/LICENSE-3rdparty.csv
+++ b/LICENSE-3rdparty.csv
@@ -41,3 +41,4 @@ uuid,https://crates.io/crates/uuid,MIT,Copyright 2018 The Uuid Developers
 valico,https://github.com/s-panferov/valico,MIT,Copyright (c) 2014 Stanislav Panferov
 vectorscan,https://github.com/VectorCamp/vectorscan,BSD-3-Clause,Copyright (c) 2020 VectorCamp PC
 walkdir,https://github.com/BurntSushi/walkdir,MIT,Copyright (c) 2015 Andrew Gallant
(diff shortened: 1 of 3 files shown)


juli1:
Why are we caching HTTP requests? To me this is early optimization and be the source of a lot of bugs. Any strong reason to do so?

juli1:
> Yeah, my thinking is just that if a user is running the scanner from their IDE and they save a file, or they are doing something that triggers a re-scan, we can avoid unnecessary HTTP requests, which felt to me more important for a dev laptop than, say, a CI instance. (FWIW in my initial prototyping, duplicate requests happened pretty often for "simple" rules like a 20 byte hex string)
> 
> However, I am happy to revert this if it's overly complex.

We are really not at this stage yet, let's focus on having the whole system working and we can optimize later.

PR #304, opened 2024-04-03
fix(ide): support for new properties when ignoring rules

robertohuertasm:

When globally ignoring a rule, the rule was completely overriden. As we have to support new properties we must keep them even if the rule is ignored.

## What is your solution?

Properly handle this case and keep the existing properties. So basically, mutate the `paths.ignore` property if the rule was already present.

re #301

IDE-2627


m-paternostro (approved):
Looks good!

PR #310, opened 2024-04-05
fix(serialization): correctly serialize arguments

robertohuertasm:

Serialization of the `ArgumentValues` struct is not properly done. This causes issues for the IDEs when using serialization for ignoring rules.

So, this:

```yaml
- typescript-code-style:
  rules:
    max-params:
      arguments:
        max-params: 3
```

becomed:

```yaml
- typescript-code-style:
  rules:
    max-params:
      ignore:
      - '**'
      arguments:
        max-params:
          by_subtree:
            '': '3'
```

## What is your solution?

Added a custom serialization method and some tests covering it.

Feel free to modify or expand this implementation. I was wondering why we need the `ArgumentValues` at all given that we chose to ignore the `by_subtree` property. If there's no point, maybe we could use a HashMap or a a tuple struct. Although, I guess we would still want to have a custom serializer/deserializer to avoid having this case:
```yaml
arguments:
  max-params:
    '': '3'
```

So... I guess it's ok.

IDE-2650

### DIFF
--- a/crates/static-analysis-kernel/src/config_file.rs
+++ b/crates/static-analysis-kernel/src/config_file.rs
@@ -1176,4 +1176,117 @@ rulesets:
 
         assert_eq!(serialized, expected);
     }
+
+    #[test]
+    fn test_serialize_arguments() {
+        let mut rulesets = IndexMap::new();
+
+        let mut rules: IndexMap<String, RuleConfig> = IndexMap::new();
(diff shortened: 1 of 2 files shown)


jacobotb (approved):

robertohuertasm:
Thanks @jacobotb for that last push! Really appreciate the `IndexMap` inclusion ❤️!

PR #320, opened 2024-04-12
[STAL-1818] Integrate secrets scanning into the main binary

jasonforal:
Introduce secrets scanning to the static analyzer

## What is your solution?
* Replace hardcoded integration test with production implementation
* Adds support for a multi-rule file, as per the spec.
  * Currently expects this rule file to be present on the filesystem. The rationale is that it allows the user to control the exact ruleset being run instead of trusting an external HTTP request. The Datadog GitHub action will include a step to `curl` the ruleset from the Git repo that hosts the default ruleset.
* The files scanned are those initially discovered by the `files_to_analyze` logic.

Introduces three new CLI flags:
```
./datadog-static-analyzer --secrets-scan --secrets-validate --secrets-rules path/to/rules.yml
```

With a rules file like:
```yml
rule-one:
   schema-version: v1
   id: rule-one
   matcher:
      # ...
   validator:
     # ...
rule-two:
   schema-version: v1
   id: rule-two
   matcher:
      # ...
   validator:
     # ...
# ...
```

When run without validation, the output is like:
Screenshot 2024-04-12 at 12 14 25

And with validation:
Screenshot 2024-04-12 at 12 13 55

* The validation attempts use a thread pool to prevent resource exhaustion. Rayon is used for a quick dev implementation without needing to pull in extra 3rd party crates.
* `// dd-no-sa` is not supported, so currently there is no way to skip a certain line.
* Flexible rule config integration has not been done

### DIFF
--- a/crates/bins/Cargo.toml
+++ b/crates/bins/Cargo.toml
@@ -27,7 +27,7 @@ features = [
 
 [features]
 default = []
-secrets = ["dep:secrets", "threadpool", "cli/secrets"]
+secrets = ["dep:secrets", "cli/secrets"]
 
 [dependencies]
 # local
@@ -53,7 +53,6 @@ tracing-subscriber = { version = "0.3.18", features = ["fmt", "env-filter"] }
 thiserror = "1"
 # Secret Scanning
 secrets = { path = "../secrets", optional = true, default-features = false }
-threadpool = { version = "1.8.1", optional = true }
 
 
 # For linux and macos, we need the vendored ssl
--- a/crates/bins/src/bin/datadog-static-analyzer.rs
+++ b/crates/bins/src/bin/datadog-static-analyzer.rs
@@ -579,10 +579,9 @@ fn main() -> Result<()> {
         use cli::secrets::{as_position, ValidationStatus};
         use secrets::core::validator::Candidate;
         use secrets::ScannerBuilder;
-        use std::sync::mpsc;
+        use std::sync::atomic::{AtomicBool, Ordering};
         use std::sync::Arc;
         use std::time::Duration;
-        use threadpool::ThreadPool;
 
         let rule_file = secrets_rule_file.expect("should have been checked");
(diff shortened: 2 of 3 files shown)


crates/bins/Cargo.toml
```diff
@@ -51,8 +51,9 @@ rayon = "1.7.0"
 rocket = { version = "=0.5.0", features = ["json"] }
 tracing-subscriber = { version = "0.3.18", features = ["fmt", "env-filter"] }
 thiserror = "1"
-# Temporary integration test
+# Secret Scanning
 secrets = { path = "../secrets", optional = true, default-features = false }
+threadpool = { version = "1.8.1", optional = true }
```
juli1, line 56: Why using this? We already have everything working with `rayon`?

crates/bins/src/bin/datadog-static-analyzer.rs
```diff
@@ -549,141 +571,127 @@ fn main() -> Result<()> {
...
-        use secrets::temp_integration_test::{build_secrets_engine, shannon_entropy};
+    if scan_for_secrets && secrets_rule_file.is_some() {
+        use cli::secrets::{as_position, ValidationStatus};
+        use secrets::core::validator::Candidate;
+        use secrets::ScannerBuilder;
         use std::sync::mpsc;
-        use std::sync::mpsc::RecvTimeoutError;
         use std::sync::Arc;
         use std::time::Duration;
+        use threadpool::ThreadPool;
```
juli1, line 585: Why using this? I believe we said everything happened sequentially and we will rely on having `rayon` doing all parallelism?

crates/bins/Cargo.toml
juli1: Yes but it's adding more dependencies when we do not really have proof is slows things down *significantly*. Can you do without?

juli1 (approved):

PR #326, opened 2024-04-19
Bump `deno_core` to 0.275.0

jasonforal:
# [DO NOT MERGE]

This is being tested on staging.

We previously pinned to `0.196.0` because it was the (second to) last official GitHub release (https://github.com/denoland/deno_core/releases)

However, the reason for this is that the deno team changed their release process such that the `deno_core` that ships with an official deno release (https://github.com/denoland/deno/releases) is considered production ready.

As of this PR, the latest deno is v1.42.4, which ships with `deno_core` 0.275.0.

## What is your solution?
Bump `deno_core` to `0.275.0`. Use its re-exported `serde_v8` instead of pulling in our own.

* Motivations for this are the various improvements to the library (e.g. some segfault fixes), but it also improves some aspects of the JavaScript <> Rust interop.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -85,7 +85,7 @@ test-and-build-amd64:
 trigger_internal_image:
   stage: release
   only:
-    - main
+    - jf/STAL-1210
   trigger:
     project: DataDog/images
     branch: ${DD_IMAGES_BRANCH}


.gitlab-ci.yml
```diff
@@ -85,7 +85,7 @@ test-and-build-amd64:
 trigger_internal_image:
   stage: release
   only:
-    - main
+    - jf/STAL-1210
```
juli1, line 88: maybe change this?

PR #327, opened 2024-04-19
[STAL-1886] JavaScript -> Rust Proof of Concept

jasonforal:
## NOTE:  Despite CI failures: this is a fully working proof of concept, but kept as a draft

To run:
1. Create an empty directory to test this repo.
2. In this repo, create a TypeScript file with a variable assignment to the string "hello". For example: `echo 'let foo = "hello"\nlet bar = "world"' > index.ts`
3. Run the analyzer, pointing it to this directory: `cargo run --bin datadog-static-analyzer -- --format sarif --output result.sarif --directory ~/sandbox`
4. You'll see the result of the Rust function call printed in stdout:
Screenshot 2024-04-19 at 17 34 23

Demonstrate that we can call into Rust from JavaScript to perform expensive tree-based operations.

## What is your solution?
Quick and messy prototype that shows ability to pass `tree_sitter::Tree` into a Rust function that can be called from from the JavaScript runtime.

This function counts the number of named nodes in the file's tree.

* Just a sketch, no consideration given to code organization.
* CI only fails because I circumvented standard operation to make the POC run without config

### DIFF
--- a/crates/static-analysis-kernel/src/analysis/javascript.rs
+++ b/crates/static-analysis-kernel/src/analysis/javascript.rs
@@ -57,9 +57,9 @@ fn op_tree_node_count(
 }
 
 deno_core::extension!(
-    poc_math,
+    ddsa_stdlib,
     ops = [op_tree_node_count],
-    esm_entry_point = "ext:poc_math/bootstrap.js",
+    esm_entry_point = "ext:ddsa_stdlib/bootstrap.js",
     esm = [dir "src/analysis/js", "bootstrap.js"],
     options = {
         res_tree: ResourceTree,
@@ -76,7 +76,7 @@ lazy_static! {
     static ref STARTUP_DATA: Vec<u8> = {
         let code: FastString = FastString::from_static(include_str!("./js/stella.js"));
         let mut rt = JsRuntimeForSnapshot::new(RuntimeOptions {
-            extensions: vec![poc_math::ext()],
+            extensions: vec![ddsa_stdlib::ext()],
             ..Default::default()
         });
         rt.execute_script("common_js", code).unwrap();
@@ -145,7 +145,7 @@ pub fn execute_rule(
         scope.spawn(move || {
             let mut runtime = JsRuntime::new(RuntimeOptions {
                 startup_snapshot: Some(&STARTUP_DATA),
-                extensions: vec![poc_math::init_ops(tree)],
+                extensions: vec![ddsa_stdlib::init_ops(tree)],
                 ..Default::default()
             });
 


crates/static-analysis-kernel/src/analysis/javascript.rs
```diff
@@ -15,10 +15,65 @@ use serde::{Deserialize, Serialize};
...
+        }
+    }
+
+    Ok(node_count)
+}
+
+deno_core::extension!(
+    poc_math,
+    ops = [op_tree_node_count],
+    esm_entry_point = "ext:poc_math/bootstrap.js",
```
juli1, line 57: Where this `ext:poc_math` is defined?

crates/static-analysis-kernel/src/analysis/javascript.rs
```diff
@@ -31,6 +86,25 @@ struct StellaExecution {
...
+        Self(tree)
+    }
+
+    pub fn tree(&self) -> &tree_sitter::Tree {
+        use std::borrow::Borrow;
+        self.0.borrow()
+    }
+}
+
+impl deno_core::Resource for ResourceTree {}
```
juli1, line 106: We should isolate this somewhere else that clearly defines where the runtime is. We may have more types like this, what about a place like `src/analysis/javascript/runtime`? The current file could then go to `src/analysis/javascript/main.rs` or something similar

crates/static-analysis-kernel/src/analysis/javascript.rs
```diff
@@ -61,10 +136,11 @@ pub fn execute_rule(
...
     let started = lock.lock().expect("should lock mutex");
 
-    thread::scope(|scope| {
-        scope.spawn(|| {
+    thread::scope(move |scope| {
+        scope.spawn(move || {
             let mut runtime = JsRuntime::new(RuntimeOptions {
-                startup_snapshot: Some(Snapshot::Static(&STARTUP_DATA)),
+                startup_snapshot: Some(&STARTUP_DATA),
+                extensions: vec![poc_math::init_ops(tree)],
```
juli1, line 143: Do we still need this even it this is contained in the `STARTUP_DATA`?

crates/static-analysis-kernel/src/analysis/javascript.rs
```diff
@@ -15,10 +15,65 @@ use serde::{Deserialize, Serialize};
...
+            }
+            if cursor.goto_next_sibling() {
+                // Travel down the sibling
+                break 'inner;
+            }
+        }
+    }
+
+    Ok(node_count)
+}
```
juli1, line 52: This should rather belong to a runtime element in `src/analysis/javascript.rs`

crates/static-analysis-kernel/src/analysis/javascript.rs
```diff
@@ -39,6 +113,7 @@ pub fn execute_rule(
     filename: String,
     analysis_options: AnalysisOptions,
     file_context: &FileContext,
+    tree: ResourceTree,
```
juli1, line 116: Instead of passing the tree, we should have a structure like `JavascriptRuntimeContext` or `JavascriptExecutionContext` that will contain all the data/values we need. This way, if we need more data/values in the JavaScript context, we just need to add more attribute in this type/structure.

juli1 (commented):
This is very encouraging and exciting. Still a few comments:

1. We should be able to pass values/arguments from the JS world to Rust. Can we include this? We could just have a function in Rust that takes a string as input and returns the same string for this POC (and we could call this function "echo" or something like this)
2. We should separate the "execution" part of `javascript.rs` from the "runtime" aspect. Right now, everything is in one file and there is no clear separation of concerns in this file. It may be the right time to do this now.
3. We should document how to write a new function for the JavaScript runtime.

PR #331, opened 2024-04-22
[STAL-1172] Optimize threading and instantiation of v8 runtimes

jasonforal:
PR 2/3

We're currently instantiating  a `deno` `JsRuntime` for every analysis. Under the hood, this is calling [deno_core/core/runtime/setup.rs at a0f1b573cd7712ba72c6a4e64346f347e8547637 · denoland/deno_core · GitHub](https://github.com/denoland/deno_core/blob/a0f1b573cd7712ba72c6a4e64346f347e8547637/core/runtime/setup.rs#L14), which (expensively) initializes v8. It's also relatively expensive to de-initialize this struct.

## What is your solution?
Reduce the time complexity of expensive work from `O(N x M)`, where N = file count, M = rule count to `O(1)`.

Currently, for each analysis, we create a new thread, initialize a `JsRuntime` on it, and use the rayon thread as the watchdog to enforce a timeout.

This PR inverts that logic, (re)using a `JsRuntime` on the rayon thread and spawning a watchdog thread for each analysis to enforce a timeout. In practice, this means that each rayon thread instantiates a single `JsRuntime` and re-uses it. Spawning watchdog threads is cheap, as they only allocate an `Arc<AtomicBool>` and mostly just remain parked.

### Before (101s)
Note how over 16% of the time is spent creating and destroying `JsRuntime`.
before

### After (81s: 1.25x speedup)
after

* We no longer need to use a static snapshot, as it only adds extra complexity with no noticeable gain.
* Despite re-using the runtime, the code isolation behavior is the exact same as before. There are no side-effects between executions or ability for one execution to modify the environment for another execution.
  * In this sense, there are no security issues.
* The `ParserBase<v8::internal::Parser>::ParseObjectPropertyDefinition` hotspot will be addressed in the next PR.

### DIFF
--- a/crates/static-analysis-kernel/src/analysis/javascript.rs
+++ b/crates/static-analysis-kernel/src/analysis/javascript.rs
@@ -184,6 +184,7 @@ return {{
                 iso_handle_clone.terminate_execution();
                 break true;
             } else if done_flag_clone.load(Ordering::Relaxed) {
+                // The main thread that was executing the JavaScript has toggled this atomic flag, indicating that it's done.
                 break false;
             }
             // This was a spurious wakeup. Adjust the timeout for the next call to `park_timeout`.


crates/static-analysis-kernel/src/analysis/javascript.rs
```diff
@@ -218,6 +134,12 @@ fn execute_rule_internal(
     // node context with the file-context we calculated for each file.
     let js_code = format!(
         r#"
+(() => {{
+// Clear arrays that might've been populated by a prior node's execution.
+// NOTE: These variables exist in the scope immediately outside of this IIFE.
+stellaAllErrors.length = 0;
+console.lines.length = 0;
```
juli1, line 141: 1. Why not reassign an empty array?
2. If we want to have a function that cleans the runtime, we should put this in `stella.js` and call it here.

More globally, I wonder why we need to do this. I believe that it's not necessary since we run an[ isolate](https://deno.com/blog/anatomy-isolate-cloud) for each execution. So the environment and variables should not be shared?

crates/static-analysis-kernel/src/analysis/javascript.rs
```diff
@@ -248,9 +170,44 @@ res
...
+    let timed_out = std::thread::spawn(move || {
+        let start = Instant::now();
+        let timeout = JAVASCRIPT_EXECUTION_TIMEOUT;
+        let mut timeout_remaining = timeout;
+        loop {
+            std::thread::park_timeout(timeout_remaining);
+            let elapsed = start.elapsed();
+            if elapsed > timeout {
+                iso_handle_clone.terminate_execution();
+                break true;
```
juli1, line 190: why not `return`?

crates/static-analysis-kernel/src/analysis/javascript.rs
```diff
@@ -248,9 +170,44 @@ res
...
+            } else if done_flag_clone.load(Ordering::Relaxed) {
```
juli1, line 191: Can we put a comment about this condition? For developers not familiar with concurrency, they need to understand it comes from the other thread

juli1 (commented):
The main question/concern is about isolate and the need to clean the VM.

In the previous execution mode, we had a new V8 engine/VM. It ensures that each execution was "cleaned".

With this mode, we have a runtime per thread. I guess that because we are creating a new isolate for each execution, we should not reset the runtime. If we need to reset the runtime, we should do this in the JS code and not the rust code.

crates/static-analysis-kernel/src/analysis/javascript.rs
juli1: No it's fine, just checking if I missed anything.

crates/static-analysis-kernel/src/analysis/javascript.rs
juli1: Agreed. Let's have a function in `stella.js` that clean the state instead of putting it here.

juli1 (approved):

PR #332, opened 2024-04-22
[STAL-1172] Serialize Rust structs directly to v8 objects

jasonforal:
PR 3/3

When we pass data between Rust and v8, we currently serialize it into JSON, pass it in as a string, and let the v8 runtime parse the objects and re-construct them into v8 objects.

This leads to large spikes in memory usage, as well as a lot of CPU cycles serializing and deserializing.

## What is your solution?
Serialize Rust structs directly into v8 objects, and expose those as globals in the JavaScript context. This leads to significant speedups in real-world workloads.

It additionally enables features that were previously infeasible, as we now have the ability to quickly and cheaply drop into JavaScript.

### Before (81s)
Note how over 37% of the time is spent serializing and executing the rule (of which 20% is parsing the JSON string we are sending in)
before

### After (58s: 1.4x speedup)
Now, only 17% of the time is spent serializing data and executing the rule.
after

* A small optimization was made to the JavaScript: we check if the `fileContext` is non-empty before spreading the object. While it's possible v8 might be smart enough to optimize this, if it doesn't, we'll needlessly reallocate a potentially large amount of data.
* The script execution behavior has no regressions (if anything, it improves it by reporting more errors because of the new use of Try/Catch)
* The "v8 string max length" check is no longer relevant anymore. Before, it was possible to go over this limit when serializing a large number of nodes with a large number of children into a string. Now, the rule's JavaScript text would need to be over this limit (~1GB), which is practically impossible.

### DIFF
--- a/crates/static-analysis-kernel/src/analysis/javascript.rs
+++ b/crates/static-analysis-kernel/src/analysis/javascript.rs
@@ -1,8 +1,9 @@
 use crate::model::analysis::{
-    AnalysisOptions, MatchNode, ERROR_RULE_CODE_TOO_BIG, ERROR_RULE_EXECUTION, ERROR_RULE_TIMEOUT,
+    AnalysisOptions, MatchNode, ERROR_RULE_EXECUTION, ERROR_RULE_TIMEOUT,
 };
 use crate::model::rule::{RuleInternal, RuleResult};
 use crate::model::violation::Violation;
+use deno_core::v8::NewStringType::Internalized;
 use deno_core::{v8, JsRuntime};
 use std::cell::RefCell;
 use std::sync::atomic::{AtomicBool, Ordering};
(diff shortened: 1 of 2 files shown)


crates/static-analysis-kernel/src/analysis/javascript.rs
```diff
@@ -140,41 +137,71 @@ fn execute_rule_internal(
...
+// The rule's JavaScript code
+//////////////////////////////
 {}
+//////////////////////////////
 
-{}.forEach(n => {{
-    n.context = {{...n.context, ...file_context}};
-    visit(n, filename, n.context.code);
-}});
+for (const n of GLOBAL_nodes) {{
```
juli1, line 145: Put a comment where `GLOBAL_nodes` is populated.

juli1:
The changes look good. However, there is a bug somewhere.

I ran this version of the analyzer along with the version 0.3.1 and the results are different:

```
with these changes
===============
dd-source: Found 11330 violation(s) in 39690 file(s) using 237 rule(s) within 158 sec(s)
dogweb: Found 19016 violation(s) in 15534 file(s) using 94 rule(s) within 89 sec(s)


v0.3.1
======
dd-source: Found 11329 violation(s) in 39690 file(s) using 237 rule(s) within 212 sec(s)
dogweb: Found 19018 violation(s) in 15534 file(s) using 94 rule(s) within 332 sec(s)
```

It shows what we can expect in terms of performance gains. But the problem is that we do not have the same number of violations.

What I recommend is that we build a utility we put in the `misc` utility that compares the SARIF output. This way, we will see where the change from from (e.g. what rule and file).

juli1:
Checked with @jasonforal, we will add a tool to diff SARIF results and this will not be shipped but used for debugging.

juli1 (approved):

PR #339, opened 2024-04-30
[STAL-1686] Add is_testing to rules

modernplumbing:
We would like to create a system in which we can test rules before adding them to the pool of production-ready rules.

## What is your solution?
As part of an effort to build a system in which rules can be tested but not be disruptive, we're adding an `is_testing` boolean flag on rules. This tag can be used for filtering, tagging, etc.

This PR adds handling to the analyzer for the `is_testing` boolean flag on rules retrieved from the Rules API. 

# Testing
Create a ruleset that has both a testing rule and a non-testing rule. Populate with rules that are known to have violations in the datadog-static-analyzer-test-repo.

Make the config use the test ruleset:
```
 ~/g/s/g/D/datadog-static-analyzer: cat ../datadog-static-analyzer-test-repo/static-analysis.datadog.yml          
rulesets:
  - celia-testing
```

Run a scan without including the test rule
```
cargo run --bin datadog-static-analyzer -- --directory ../datadog-static-analyzer-test-repo --format sarif --output test.sarif --staging
```

See lack of a `DATADOG_TESTING:true` tag in the SARIF output:
```
 "properties": {
            "tags": [
              "DATADOG_CATEGORY:BEST_PRACTICES"
            ]
          },
```

Run a scan including the test rule
```
cargo run --bin datadog-static-analyzer -- --directory ../datadog-static-analyzer-test-repo --format sarif --output test.sarif --staging --include-testing-rules
```

See at least one more rule in the rule count, and see in the output sarif something like this (focusing on the tags): 
```
{
          "fixes": [],
          "level": "error",
          "locations": [
            {
              "physicalLocation": {
                "artifactLocation": {
                  "uri": "java/path-traversal-file-read.java"
                },
                "region": {
                  "endColumn": 56,
                  "endLine": 7,
                  "startColumn": 51,
                  "startLine": 7
                }
              }
            }
          ],
          "message": {
            "text": "Path traversal using an un-sanitized value"
          },
          "partialFingerprints": {
            "DATADOG_FINGERPRINT": "97b8fe5189cabfa9223611d586882bb7bfe422493b1b2bc65c81f59732cb4fa4"
          },
          "properties": {
            "tags": [
              "DATADOG_CATEGORY:BEST_PRACTICES",
              "DATADOG_TESTING:true"
            ]
          },
          "ruleId": "celia-testing/celia-testing",
          "ruleIndex": 0
        }
```

### DIFF
--- a/crates/cli/src/model/cli_configuration.rs
+++ b/crates/cli/src/model/cli_configuration.rs
@@ -150,7 +150,6 @@ mod tests {
                 tree_sitter_query_base64: None,
                 arguments: vec![],
                 tests: vec![],
-                is_testing: false,
             }],
             path_restrictions: PathRestrictions::default(),
             argument_provider: ArgumentProvider::new(),
--- a/crates/cli/src/model/cli_configuration.rs
+++ b/crates/cli/src/model/cli_configuration.rs
@@ -150,7 +150,7 @@ mod tests {
                 tree_sitter_query_base64: None,
                 arguments: vec![],
                 tests: vec![],
-                is_testing: false,
+                is_testing: Some(false),
             }],
             path_restrictions: PathRestrictions::default(),
             argument_provider: ArgumentProvider::new(),
--- a/crates/cli/src/model/cli_configuration.rs
+++ b/crates/cli/src/model/cli_configuration.rs
@@ -150,7 +150,7 @@ mod tests {
                 tree_sitter_query_base64: None,
                 arguments: vec![],
                 tests: vec![],
-                is_testing: Some(false),
+                is_testing: false,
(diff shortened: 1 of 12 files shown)


crates/cli/src/model/cli_configuration.rs
```diff
@@ -150,6 +150,7 @@ mod tests {
                 tree_sitter_query_base64: None,
                 arguments: vec![],
                 tests: vec![],
+                is_testing: Some(false),
```
modernplumbing, line 153: From the API, it will always be true or false cuz we're not using omitempty or a bool pointer.

PR #341, opened 2024-05-01
Use higher-level constructs for configuration serializer/deserializer.

jacobotb:
We'll need to add more configuration options soon, and the current code is excessively complex. This simplifies it and suggests easy ways to add new stuff later.

For `NamedRulesetConfig` I used the `StringOrStruct` thing instead of the enum like in `ArgumentValues` so that, if one of the fields in the ruleset config fails to parse, we'll get a specific error for that field. This happens because it sees a string or a map and commits to that while deserializing. Otherwise it will try both options, see that both fail, and we just get a generic "expected a string or ruleset configuration" error.


juli1:
For clarity, can we put what problem this PR is solving and why we did this? Not sure why we are doing this change.

jacobotb:
There's a request to add per-directory severity settings, this will make it easier to do that.

PR #342, opened 2024-05-02
[STAL-2002] Add documentation

juli1:

Users do not have clear instructions to test production rules.

## What is your solution?

Add the documentation to run the script locally

### DIFF
--- a/misc/test-production-rules.py
+++ b/misc/test-production-rules.py
@@ -1,5 +1,11 @@
 # This script is used only for CI/CD purposes
 # We assume that everything is correctly built and available in the `target` directory.
+
+# If you want to try this locally
+# 1. Build the binary locally (e.g. cargo run -r)
+# 2. Bootstrap a Python environment with requests (e.g python -mvenv venv && source venv/bin/activate && pip install requests)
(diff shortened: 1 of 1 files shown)


misc/test-production-rules.py
```diff
@@ -1,5 +1,11 @@
 # This script is used only for CI/CD purposes
 # We assume that everything is correctly built and available in the `target` directory.
+
+# If you want to try this locally
+# 1. Build the binary locally (e.g. cargo run -r)
+# 2. Bootstrap a Python environment with requests (e.g python -mvenv venv && source venv/bin/activate && pip install requests)
+# 3. Execute the script: misc/test-production-rules.py -r <RULSET-NAME> -c target/release/datadog-static-analyzer -s target/release/datadog-static-analyzer-server
```
modernplumbing, line 7: you also need to add execute permissions on the file, and remember not to commit the new permission changes. unless we are ok with making this file executable permanently

juli1: Just execute Python directly, specified this.

modernplumbing (approved):

PR #373, opened 2024-05-23
Add the ability to call other endpoints with JWT tokens.

jacobotb:
We will need to have the static analyzer call our internal API endpoints with JWT tokens for auth. This lets the program that runs the static analyzer provide the hostname and JWT token.

## What is your solution?
A `DD_HOSTNAME` envvar and a `DD_JWT_TOKEN` envvar. Those are read when requests are built and used if present.

`DD_HOSTNAME` has priority over `use_staging` and `DD_SITE`; `DD_JWT_TOKEN` is used if `DD_API_KEY` and `DD_APP_KEY` are not available.

### DIFF
--- a/crates/cli/src/datadog_utils.rs
+++ b/crates/cli/src/datadog_utils.rs
@@ -1,12 +1,13 @@
 use std::env;
 
 use crate::constants::{
-    DATADOG_HEADER_API_KEY, DATADOG_HEADER_APP_KEY, HEADER_CONTENT_TYPE,
+    DATADOG_HEADER_API_KEY, DATADOG_HEADER_APP_KEY, DATADOG_HEADER_JWT_TOKEN, HEADER_CONTENT_TYPE,
     HEADER_CONTENT_TYPE_APPLICATION_JSON,
 };
 use anyhow::{anyhow, Result};
 use kernel::model::rule::Rule;
 use kernel::model::ruleset::RuleSet;
+use reqwest::blocking::RequestBuilder;
 use uuid::Uuid;
 
 use crate::model::datadog_api::{
@@ -15,8 +16,8 @@ use crate::model::datadog_api::{
     DiffAwareResponse,
 };
 
-const STAGING_DATADOG_SITE: &str = "datad0g.com";
-const DEFAULT_DATADOG_SITE: &str = "datadoghq.com";
+const STAGING_DATADOG_HOSTNAME: &str = "api.datad0g.com";
+const DEFAULT_DATADOG_HOSTNAME: &str = "api.datadoghq.com";
 
 const DEFAULT_RULESETS_LANGUAGES: &[&str] = &[
     "CSHARP",
@@ -42,7 +43,7 @@ pub fn get_rules_from_rulesets(rulesets_name: &[String], use_staging: bool) -> R
 // Get environment variables for Datadog. First try to get the variables
 // prefixed with DD_ and then, try DATADOG_.
 // If nothing works, just returns an error.
-pub fn get_datadog_variable_value(variable: &str) -> anyhow::Result<String> {
+pub fn get_datadog_variable_value(variable: &str) -> Result<String> {
     let prefixes = vec!["DD", "DATADOG"];
     for prefix in prefixes {
         let name = format!("{}_{}", prefix, variable);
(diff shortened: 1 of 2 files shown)


crates/cli/src/datadog_utils.rs
```diff
@@ -15,8 +16,8 @@ use crate::model::datadog_api::{
     DiffAwareResponse,
 };
 
-const STAGING_DATADOG_SITE: &str = "datad0g.com";
-const DEFAULT_DATADOG_SITE: &str = "datadoghq.com";
+const STAGING_DATADOG_HOSTNAME: &str = "api.datad0g.com";
+const DEFAULT_DATADOG_HOSTNAME: &str = "api.datadoghq.com";
```
juli1, line 20: This does not compliant with the [official values](https://docs.datadoghq.com/getting_started/site/#access-the-datadog-site). If the values must be overriden, we can have the `DD_SITE` variable defined.

crates/cli/src/datadog_utils.rs
```diff
@@ -57,42 +58,68 @@ pub fn get_datadog_variable_value(variable: &str) -> anyhow::Result<String> {
...
+// If a DD_HOSTNAME envvar has been specified, use it; otherwise, use the staging hostname
+// if use_staging is true; otherwise, use the DD_SITE envvar with 'api.' prepended;
+// otherwise, use the default hostname.
+fn get_datadog_hostname(use_staging: bool) -> String {
+    if let Ok(hostname) = get_datadog_variable_value("HOSTNAME") {
+        hostname
+    } else if use_staging {
+        STAGING_DATADOG_HOSTNAME.to_string()
+    } else if let Ok(site) = get_datadog_variable_value("SITE") {
+        format!("api.{}", site).to_string()
```
juli1, line 70: I do not understand this change. Why not just passing `DD_SITE` when invoking the binary? It seems to be doing the same thing? Adding `api` to the site is a breaking change, it would not comply with the docs and conventions we have.

crates/cli/src/datadog_utils.rs
```diff
@@ -57,42 +58,68 @@ pub fn get_datadog_variable_value(variable: &str) -> anyhow::Result<String> {
...
-    );
+#[derive(PartialEq)]
+enum UseKeys {
+    // Do not add APP/API keys or a JWT token to the request.
+    DoNotUse,
+    // Add APP/API keys or JWT token to the request, if available.
+    NotRequired,
+    // Add APP/API keys or JWT token to the request, and fail if not available.
+    Required,
+}
```
juli1, line 84: Why this? We should always use the keys when possible as it prevents rate-limits. Anonymous request are all accounted for the same rate-limit (and can lead to rate-limiting issues). When we use an api/app key, the rate limit is calculated for the account. For this reason, requests should be made authenticated if a token is available.

juli1 (changes requested):

crates/cli/src/datadog_utils.rs
jacobotb: Prepending `api.` is the current behavior. `DD_SITE` only provides the part that comes after `api.`. We need `DD_HOSTNAME` because the internal endpoints don't start with `api.`.

crates/cli/src/datadog_utils.rs
jacobotb: I do this to preserve the current behavior of the functions I'm replacing. `get_ruleset` uses the keys if they are available but works without them; `get_default_rulesets_name_for_language` does not use the keys at all; `get_diff_aware_information` requires the keys.

I can change `get_default_rulesets_name_for_language` to pass the keys if available, but that's not its current behavior.

juli1: The part that I do not understand is why we need to have `DoNotUse`? We should use whatever keys we have when we have it.

jacobotb: My PR keeps the current behavior, in which one of the functions does not include keys. If the current behavior is wrong, I can fix it in the same PR and remove `DoNotUse`.

crates/cli/src/datadog_utils.rs
juli1: my comment was rather that there are two concerns at play here: detect the site and detect the hostname. And we put everything in a single function. I would have kept a function to get the site and add another function to get the hostname (that calls the function to get the site). This separates concerns.

Might be overdoing it and not a blocker.

juli1 (approved):

PR #382, opened 2024-05-28
[STAL-1960] Introduce patterns for refactored JavaScript modules

jasonforal:
To implement tight interop between Rust and JavaScript, both languages need to have the same model of the data. This PR addresses 3 problems that this introduces:
1. Ensuring that even across future refactors, the Rust implementation and JavaScript implementation are always in sync (i.e. that they serialize/deserialize to the same v8 object shape).
2. Ensuring high performance serialization/deserialization to/from v8.
3. Doing all of this without requiring contributors to be experts in v8's lower level abstractions.

## What is your solution?
**1. Unit test canary:**
Ensuring parity without codegen is tricky, especially because JavaScript isn't strictly typed. For example, imagine as part of a large refactor, we are adding numerous properties to `Fix`, one of which is a "category" string field. If we implemented this in Rust, but forgot to add "category" on the JavaScript side, v8 is happy to send a `v8::Value` (`undefined`) in response to a lookup of this property `category`. Depending on how the developer writes their unit tests (and the nature of the business logic), there might be some subtle, uncaught edge-cases.
* For each JavaScript class, [all its properties are explicitly specified](https://github.com/DataDog/datadog-static-analyzer/commit/52a34651314c9273268361cde667ad07b79ab7ae#diff-d0ef49f2887fd5052a7abbb3e0973d1f8a28bf25e095bfce3f187f62c292337cR111-R125) in a unit test. If the JavaScript changes in any way, this test will fail.
    * Note: this isn't 100% robust (e.g. a refactor in Rust could introduce a logic "bug" independent of the JavaScript object's shape), but it serves as a canary to make a developer explicitly acknowledge the JavaScript drift.
    * To reduce test boilerplate, I wrote a lot of [test helper functions](https://github.com/DataDog/datadog-static-analyzer/blob/224f9bf515a898c877e19d38bc0db08074964ec4/crates/static-analysis-kernel/src/analysis/ddsa_lib/test_utils.rs). These additionally allow us to keep the [`ddsa_lib` Deno extension](https://github.com/DataDog/datadog-static-analyzer/blob/224f9bf515a898c877e19d38bc0db08074964ec4/crates/static-analysis-kernel/src/analysis/ddsa_lib/extension.rs#L7-L22) as a single-source of truth.

**2. Manually-specified serialization/deserialization:**
`serde_v8` is slow and does not provide APIs to amortize allocations (which we can/should do because we know the shapes of our objects ahead of time). Furthermore, the deno team plans to eventually deprecate this crate.
* [We build on the recently-introduced `V8Converter` trait](https://github.com/DataDog/datadog-static-analyzer/commit/52a34651314c9273268361cde667ad07b79ab7ae#diff-84b7a5afe75de01aadcd871617b2d1072605a1e3c0d9e76fdec2a809d6d3c65dR77-R83) to simplify this.
* This solution requires a struct be serialized/deserialized by specifying individual fields. I added [some ergonomic functions](https://github.com/DataDog/datadog-static-analyzer/commit/52a34651314c9273268361cde667ad07b79ab7ae#diff-b6424b794288792913f6e5108c4a1bddbbb336aec1aed5befc5a4327332c5c70R108) for this, but the "real" solution is a procedural macro to make this declarative (out of this PR's scope).

**3. File organization**
This is still not "solved" in this PR, but the way I am trying to address this is to create consistent, logical structure to the file organization. Conceptually, what this PR proposes is:
* `edit.js` - The JavaScript class
* `edit.rs` - **Only** the Rust code necessary to serialize/deserialize. These files and structs are _not_ used for any business logic (e.g. [we convert from `EditInstance`](https://github.com/DataDog/datadog-static-analyzer/blob/52a34651314c9273268361cde667ad07b79ab7ae/crates/static-analysis-kernel/src/analysis/ddsa_lib/js/edit.rs#L41) to the struct internally used for business logic).
```
├── js
│   ├── edit.js
│   ├── edit.rs
│   ├── fix.js
│   ├── fix.rs
│   ├── violation.js
│   └── violation.rs
├── js.rs
```
This should make the API boundaries more clear (and git history more readable).

## Alternatives considered
* Using a [generalized deserializer](https://github.com/DataDog/datadog-static-analyzer/commit/224f9bf515a898c877e19d38bc0db08074964ec4#diff-853135d9aef45be68bb678db252253d24b6db0ff970f0bdca53866f6375d8d06L241) like we currently use via `serde_v8` (reason: we don't need to design for handling data of an unknown type).
* Having Rust be the single-source-of-truth. Instead of defining classes (e.g. `class Fix { ... }`) in a JavaScript file, we would [define all v8 objects directly via the v8 API](https://v8.github.io/api/head/classv8_1_1ObjectTemplate.html). We would then need codegen to produce JavaScript type definitions. (reason: enormous complexity).
* Writing a procedural macro to generate high performance serialization/deserialization in a declarative way (reason: out of scope).

This is one of a chain of PRs fully implementing the `ddsa_lib`.
* `Edit`, `Fix`, and `Violation` didn't "need" to be refactored, as their core JavaScript representation is basically the same. However, to keep the library unified, they were re-implemented as JavaScript modules instead of reusing the `StellaEdit`, `StellaFix`, and `StellaError` functions.
* These structs were chosen to keep this PR small, because they are implemented basically identical to their `stella.js` counterparts. This is in contrast to other structs like `TreeSitterNode` and `Capture`, which need to be high performance, and so require a lot of under-the-hood refactoring on the Rust side of things.

### DIFF
--- a/crates/static-analysis-kernel/src/analysis/ddsa_lib/js/edit.rs
+++ b/crates/static-analysis-kernel/src/analysis/ddsa_lib/js/edit.rs
@@ -145,8 +145,32 @@ impl V8Converter for EditConverter {
 
 #[cfg(test)]
 mod tests {
-    use crate::analysis::ddsa_lib::js::EditInstance;
-    use crate::analysis::ddsa_lib::test_utils::{js_class_eq, js_instance_eq};
+    use crate::analysis::ddsa_lib::common::DDSAJsRuntimeError;
+    use crate::analysis::ddsa_lib::js::{EditConverter, EditInstance};
+    use crate::analysis::ddsa_lib::test_utils::{
+        cfg_test_runtime, js_class_eq, js_instance_eq, try_execute,
+    };
+    use crate::analysis::ddsa_lib::v8_ds::V8Converter;
+    use deno_core::v8::HandleScope;
+
+    #[rustfmt::skip]
+    fn assert_de_ok(input: &str, s: &mut HandleScope) {
+        let c = EditConverter::new();
+        assert!(try_execute(s, input).is_ok_and(|v| c.try_convert_from(s, v).is_ok()));
+    }
+    #[rustfmt::skip]
+    fn assert_de_fail(input: &str, scope: &mut HandleScope, expected: &str) {
+        let c = EditConverter::new();
+        assert!(try_execute(scope, input).is_ok_and(|v| {
+            let err = c.try_convert_from(scope, v).unwrap_err();
+            match expected {
+                "NotFound" => { matches!(err, DDSAJsRuntimeError::VariableNotFound { .. })}
+                "WrongType" => { matches!(err, DDSAJsRuntimeError::WrongType { .. })}
+                "InvalidValue" => { matches!(err, DDSAJsRuntimeError::InvalidValue { .. })}
+                _ => unreachable!()
+            }
+        }));
+    }
 
     #[test]
     fn js_properties_canary() {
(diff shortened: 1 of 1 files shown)


crates/static-analysis-kernel/src/analysis/ddsa_lib/common.rs
```diff
@@ -8,12 +8,48 @@ use std::fmt::Debug;
...
+    #[error("type should be \"{expected}\", not \"{got}\"")]
+    WrongType {
+        expected: &'static str,
+        got: &'static str,
+    },
+    #[error("invalid value for `{identifier}`: expected {expected}")]
+    InvalidValue {
+        identifier: &'static str,
+        expected: &'static str,
+    },
```
juli1, line 22: We should have tests to check that these errors are correctly caught/raised?

crates/static-analysis-kernel/src/analysis/ddsa_lib/common.rs
```diff
@@ -8,12 +8,48 @@ use std::fmt::Debug;
...
+    })?;
+    let func = v8::Local::<v8::Function>::try_from(id_value).map_err(|_| {
+        DDSAJsRuntimeError::WrongType {
+            expected: "function",
+            got: id_value.type_repr(),
+        }
+    })?;
+    Ok(v8::Global::new(scope, func))
+}
+
```
juli1, line 52: Seems to be only used in tests. Do you plan to use it to load the new runtime or it's purely done for testing? If the latter, maybe move it into tests?

crates/static-analysis-kernel/src/analysis/ddsa_lib/common.rs
```diff
@@ -38,3 +74,91 @@ pub fn v8_string<'s>(scope: &mut HandleScope<'s>, str: &str) -> v8::Local<'s, v8
 pub fn v8_uint<'s>(scope: &mut HandleScope<'s>, number: u32) -> v8::Local<'s, v8::Integer> {
     v8::Integer::new_from_unsigned(scope, number)
 }
+
+/// A shorthand for requiring that a v8 value be defined, otherwise an error is returned.
+pub fn expecting_var<T>(value: Option<T>, name: &str) -> Result<T, DDSAJsRuntimeError> {
+    value.ok_or_else(|| DDSAJsRuntimeError::VariableNotFound {
```
juli1, line 80: Here, this is a good way to test the `VariableNotFound` error: we can just try a rule with an `Edit` that does not have a `start_line` or `end_line`.

crates/static-analysis-kernel/src/analysis/javascript.rs
```diff
@@ -237,7 +246,19 @@ return {{
         ExecutionError::Execution { reason }
     })?;
 
-    let StellaExecution { mut violations } = serde_v8::from_v8(tc_scope, execution_result)
+    let v8_array: v8::Local<v8::Array> =
+        execution_result.try_into().map_err(|err: v8::DataError| {
+            let reason = err.to_string();
+            ExecutionError::UnexpectedReturnValue { reason }
```
juli1, line 252: Can we have a test for this to make sure we test this codepath?

juli1 (approved):
Very solid. Good architecture and well documented.

Left comments to add tests. Feel free to ship without it if needed.

PR #385, opened 2024-06-04
[STAL-1960] Add ddsa FileContextGo

jasonforal:
We want to be able to expose per-filetype metadata to a JavaScript rule in order to enable a both more powerful and less verbose API in JavaScript.

## What is your solution?
This implements the exact same functionality we have in our current [`analysis/file_context/go.rs`](https://github.com/DataDog/datadog-static-analyzer/blob/476a3038dbab6c67f027a1b589a6c614dc42f196/crates/static-analysis-kernel/src/analysis/file_context/go.rs) implementation (basically copy/pasting the code [to here](https://github.com/DataDog/datadog-static-analyzer/compare/jf/STAL-1960?expand=1#diff-95bd376f25623426e1a734b4a6638233c6de9441624392774c909c8634a21124R47-R80))

Additionally, two optimizations were made:
  * The tree-sitter query is built once per runtime instead of once per query.
  * The tree-sitter query can now be executed lazily (if no rules request package aliases, the query is never performed).
  
This PR also demonstrates for future contributors:
1. How to add a file context (this was squashed into a single commit to show all touchpoints that need to be updated).
2. How to perform a tree-sitter query from JavaScript (via an op)

* This code is inert for one main reason: this PR uses a different data model (e.g. it treats "packages" as a derived value, instead of pre-computing into its own array [like we currently do](https://github.com/DataDog/datadog-static-analyzer/blob/476a3038dbab6c67f027a1b589a6c614dc42f196/crates/static-analysis-kernel/src/analysis/file_context/go.rs#L9)). Were to to actually swap out our current implementation, the refactor would be involved, and it is something that will just end up being thrown away when we fully transition to the ddsa lib. I thus skipped it.
* In the future, when we swap over to ddsa, we will delete our current [`analysis/file_context/go.rs`](https://github.com/DataDog/datadog-static-analyzer/blob/476a3038dbab6c67f027a1b589a6c614dc42f196/crates/static-analysis-kernel/src/analysis/file_context/go.rs) implementation and replace it with the one in this PR.

### DIFF
--- a/crates/static-analysis-kernel/src/analysis/ddsa_lib/context/file.rs
+++ b/crates/static-analysis-kernel/src/analysis/ddsa_lib/context/file.rs
@@ -2,11 +2,27 @@
 // This product includes software developed at Datadog (https://www.datadoghq.com/).
 // Copyright 2024 Datadog, Inc.
 
+use crate::analysis::ddsa_lib::context::file_go::FileContextGo;
+
 #[derive(Debug, Default)]
 pub struct FileContext {
     // Supported file contexts:
+    go: Option<FileContextGo>,
 }
 
 impl FileContext {
-    //
+    // Returns a mutable reference to the [`FileContextGo`] owned by this `FileContext`, if it exists.
+    pub fn go_mut(&mut self) -> Option<&mut FileContextGo> {
+        self.go.as_mut()
+    }
+
+    // Returns a reference to the [`FileContextGo`] owned by this `FileContext`, if it exists.
+    pub fn go(&self) -> Option<&FileContextGo> {
+        self.go.as_ref()
+    }
+
+    // Assigns the [`FileContextGo`] to this `FileContext`, returning the old value, if it exists.
+    pub fn set_go(&mut self, file_ctx_go: FileContextGo) -> Option<FileContextGo> {
+        Option::replace(&mut self.go, file_ctx_go)
+    }
 }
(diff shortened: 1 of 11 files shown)


crates/static-analysis-kernel/src/analysis/ddsa_lib/context/file.rs
```diff
@@ -2,11 +2,27 @@
 // This product includes software developed at Datadog (https://www.datadoghq.com/).
 // Copyright 2024 Datadog, Inc.
 
+use crate::analysis::ddsa_lib::context::file_go::FileContextGo;
+
 #[derive(Debug, Default)]
 pub struct FileContext {
     // Supported file contexts:
+    go: Option<FileContextGo>,
 }
```
juli1, line 11: We had previous conversation about making it an `enum` which we did? Why did it came back as a `struct`?

crates/static-analysis-kernel/src/analysis/ddsa_lib/js/context_file_go.js
```diff
@@ -0,0 +1,96 @@
...
+     * Note: The Rust static-analysis-kernel is responsible for setting this flag to `false` when the file
+     * changes within the execution context.
+     * @type {boolean}
+     */
+    ensureCached() {
+        if (!this.__js_isCached) {
+            op_context_file_go_sync_module_aliases();
+            this.__js_isCached = true;
+        }
+    }
```
juli1, line 95: Why do we even cache? Seems premature optimization at this stage.

juli1 (commented):
Since it uses cache, we should have a test to guarantee the cache restarts at each invocation. We should have a test (either integration or not) that runs two Go rules make use of the Go file context and check they produce the right output.

juli1 (approved):

PR #387, opened 2024-06-04
[STAL-1960] Add bridge for (TsSymbol <> Name) mappings

jasonforal:
One of the fundamental design constraints for ddsa was to represent everything that needed to be passed between Rust and JavaScript with a [v8 "smi"](https://github.com/v8/v8/blob/main/src/objects/smi.h) (a `uint31` on 64 bit platforms). At the same time, we want ergonomic APIs that aren't constrained by performance optimizations like this.

Each tree-sitter CST node (that we care about) has a string name (that it internally calls a `TSSymbol`) to represent its type:
```js
const myNumber = 123;
```
The node type names for the nodes in this expression are:
```
const myNumber = 123;     "lexical_declaration"
myNumber = 123            "variable_declarator"
myNumber                  "identifier"
123                       "number"
```

We want users to be able to author JavaScript rules that can filter nodes by a string name, like `"identifier"`. We also want to be able to let JavaScript rules request nodes of a given type as well.

In pseudo-code:
```js
const nodes = ddsa.getChildren(targetNode, "variable_declarator");
```

## What is your solution?
In the above example (`const nodes = ddsa.getChildren(node, "variable_declarator");`), this is a function that will make a call to Rust to request additional tree-sitter nodes. Under the hood, we want to be able to translate this to something like:
```js
const nodes = op_get_children(32 /* nodeId (smi) */, 149 /* TSSymbolId (smi) */);
```

To achieve this, we create and pre-calculate a v8 map with data for each tree-sitter Language. For example, in JavaScript pseudo-code, this would be like:
```js
const symbolMap = new Map([
    [1, "identifier"],
    ["identifier, 1],
    // ...
    [146, "expression_statement"],
    ["expression_statement", 146],
    // ...
    [225, "decorator"],
    ["decorator", 225],
    // ...
]);
```
With this map, we can always internally represent the type as a numeric id, but use string names in the user-facing JavaScript API.

This map will be exposed on the JavaScript `globalThis` on a per-language basis so that `ddsa` will have access to this map.

* This is just the bridge implementation -- this code is not live. It will be activated when we switch over to ddsa.

### DIFF
--- a/crates/static-analysis-kernel/src/analysis/ddsa_lib/bridge/ts_symbol_map.rs
+++ b/crates/static-analysis-kernel/src/analysis/ddsa_lib/bridge/ts_symbol_map.rs
@@ -1,0 +1,181 @@
+// Unless explicitly stated otherwise all files in this repository are licensed under the Apache License, Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2024 Datadog, Inc.
+
+use crate::analysis::ddsa_lib::common::{v8_interned, v8_uint};
+use crate::analysis::ddsa_lib::v8_ds::MirroredIndexMap;
+use deno_core::v8;
+use deno_core::v8::HandleScope;
+use std::cell::RefCell;
+use std::collections::HashMap;
+
+/// A stateful bridge holding a collection of [`MirroredTsSymbolMap`].
+#[derive(Default)]
+pub struct TsSymbolMapBridge(RefCell<HashMap<tree_sitter::Language, MirroredTsSymbolMap>>);
+
+impl TsSymbolMapBridge {
+    /// Creates a new, empty `TsSymbolMapBridge`.
+    pub fn new() -> Self {
+        Self::with_capacity(0)
+    }
+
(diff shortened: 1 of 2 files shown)


crates/static-analysis-kernel/src/analysis/ddsa_lib/bridge/ts_symbol_map.rs
```diff
@@ -0,0 +1,181 @@
...
+}
+
+/// A mapping and reverse mapping between a tree-sitter node's string name and its [`TSSymbol`](tree_sitter::ffi::TSSymbol).
+///
+/// Only "visible" and "named" nodes are included in this map.
+#[derive(Debug)]
+struct MirroredTsSymbolMap(MirroredIndexMap<NameOrSymbol, NameOrSymbol>);
+
+impl MirroredTsSymbolMap {
+    pub fn new(scope: &mut HandleScope, ts_language: &tree_sitter::Language) -> Self {
```
juli1, line 58: Where this is being called? Ideally, we should call this when we initialize the runtime and do this for all tree-sitter languages loaded in the runtime.

juli1 (approved):

PR #390, opened 2024-06-06
[STAL-1960] Implement ddsa TreeSitterNode + bridge

jasonforal:
The JavaScript environment necessarily needs to create and use many `TreeSitterNode` objects. This is a current performance bottleneck via both CPU cycles and GC pressure.

## What is your solution?
With ddsa, we refactor the JavaScript object to be represented entirely by [v8 smi](https://github.com/v8/v8/blob/main/src/objects/smi.h)s.

### TreeSitterNode

**Currently**
```typescript
// Current implementation
interface Position {
    line: number,
    col: number,
}
interface TreeSitterNode {
    astType: string,
    start: Position,
    end: Position,
    fieldName: string | undefined,
    children: TreeSitterNode[],
}
```
**New for ddsa**
```typescript
// New implementation
interface TreeSitterNode {
    id: number, /* newly introduced by this PR and explained below */
    _startLine: number,
    _startCol: number,
    _endLine: number,
    _endCol: number,
    _nodeTypeId: number
}
```
A significant difference is that we no longer store all of a node's children on the node itself. Instead, these will be retrieved on-demand (this functionality is not implemented by this PR, but will be in a future PR)

But even beyond that, whereas the prior implemented allocated (minimum) 4 additional heap objects per node, ddsa requires 0 additional allocations, as each smi is stored directly within the object's property storage.

### TreeSitterNode id
```typescript
interface TreeSitterNode {
    id: number
    // ... 
}
```
The `id` is a newly introduced field that allows us to refer to each node by a smi. This integer is assigned by [a `MirroredIndexMap` data structure](https://github.com/DataDog/datadog-static-analyzer/pull/390/files#diff-08abf96dbacec753e49da6c519caccd610a00179d43c2fb6af10205b91cc0ad7R23), and thus the insertion order of a node determines its id (making it stable across executions).

This design also ensures we only serialize a node once (whereas in the current library, we serialize a node and all of its children potentially redundantly for every rule execution).

### Use of unsafe
[We use unsafe here for ops](https://github.com/DataDog/datadog-static-analyzer/pull/390/files#diff-26f4d9f443f20bcbb4a010c53170b9bfe10ef4fd8d4bca2c233ca50729621cd6R85).

A `tree_sitter::Node` has a lifetime `<'tree>`, which references the lifetime of the `tree_sitter::Tree` it comes from. This introduces a problem for us, because we are implementing contexts within the `JsRuntime` with owned values.

The way we get around this is to store a node via its `tree_sitter::ffi::TSNode` form, which has no Rust lifetime:
```rust
// mod ffi
pub struct TSNode {
    pub context: [u32; 4usize],
    pub id: *const ::std::os::raw::c_void,
    pub tree: *const TSTree,
}
```
This contains a pointer to the tree (`pub tree`) and the node itself (`pub id`).

And so we store this raw TSNode, and during an op, when we need to perform an operation on it (e.g. get its position, fetch its children, etc) we convert it to a `Node<'tree>`. This is inherently "unsafe" by Rust's semantics because the compiler can't know that the tree still exists. However, we can, as [we are storing an owned `Tree`](https://github.com/DataDog/datadog-static-analyzer/blob/30af728c8930362ed914376f614c248e4d252c23/crates/static-analysis-kernel/src/analysis/ddsa_lib/context/root.rs#L13) during the execution (which tree-sitter [implements as a reference-counted smart pointer](https://github.com/tree-sitter/tree-sitter/blob/b40f342067a89cd6331bf4c27407588320f3c263/lib/src/subtree.c#L584-L589)), so we know tree-sitter hasn't de-allocated the tree from underneath us. And, as an additional requirement for safety, we never mutate the tree/prune any of its nodes, so we know the node must exist as well.

### JavaScript API
As detailed above, for performance, we deconstruct each Position (line/col) into individual fields.
```ts
{
    // ...
    _startLine: number,
    _startCol: number,
    _endLine: number,
    _endCol: number,
    // ...
}
```

However, this is just an internal detail -- this PR preserves the existing `Position`-based access API:
```js
// Before
const startLine = node.start.line;
const startCol = node.start.col;
// After  (no change)
const startLine = node.start.line;
const startCol = node.start.col;
// ...which is effectively sugar for:
// const startLine = node._startLine;
```

### Bridge design
As mentioned above, we insert these (raw) tree-sitter nodes into a `MirroredIndexMap`, which exposes a JavaScript `Map` within the rule execution context, which maps `id` to a [JavaScript `TreeSitterNode`](https://github.com/DataDog/datadog-static-analyzer/pull/390/files#diff-cca60bdc8f55e9d3bdaeb8a9f5432f2e07f0060c23821f94724a3bdd1276240cR30).

Thus, tree-sitter nodes are passed from Rust to JavaScript by:
1. Serializing the Rust struct directly to a `v8::Object` via the v8 API.
2. Inserting it into the `MirroredIndexMap`, taking note of its assigned id.
3. Passing that id to JavaScript.
4. JavaScript then looks up the id via the exposed global `v8::Map`.

## Alternatives considered
* A TreeSitterNode's id is assigned via its sequential insertion order into an IndexMap. It is likely that a `v8::Array` would provide faster lookup performance than a `v8::Map` here, but I did not do that because I wanted to keep the `MirroredIndexMap` solution simple. I'd be open to refactoring this later, though.

* In https://github.com/DataDog/datadog-static-analyzer/pull/385, we decided that the added complexity of caching go module aliases wasn't worth it. And so, you will notice we implement a [cache for the text of each TreeSitterNode here](https://github.com/DataDog/datadog-static-analyzer/commit/157fef8949974e6132c8b17d51cc37475f10b6d2#diff-cca60bdc8f55e9d3bdaeb8a9f5432f2e07f0060c23821f94724a3bdd1276240cR66-R77) and might think its identical to the cache decision we made earlier. However, it's really not--the main difference is each TreeSitterNode only exists _within_ a single execution, whereas the go module alias cache existed/was mutated across every execution. So the complexity here is much smaller. Additionally, the importance of this cache is much higher, as inspecting node text is very common.

### DIFF
--- a/crates/static-analysis-kernel/src/analysis/ddsa_lib/bridge/ts_node.rs
+++ b/crates/static-analysis-kernel/src/analysis/ddsa_lib/bridge/ts_node.rs
@@ -94,10 +94,11 @@ impl TsNodeBridge {
         } = node.end_position();
         let ts_node = TreeSitterNode {
             id,
-            start_line: start_line as u32,
-            start_col: start_col as u32,
-            end_line: end_line as u32,
-            end_col: end_col as u32,
+            // NOTE: We normalize the 0-based `tree_sitter::Point` to be 1-based.
+            start_line: start_line as u32 + 1,
+            start_col: start_col as u32 + 1,
+            end_line: end_line as u32 + 1,
+            end_col: end_col as u32 + 1,
             node_type_id: node.grammar_id(),
             _pd: PhantomData,
         };
@@ -169,7 +170,7 @@ impl TsNodeBridge {
 mod tests {
     use crate::analysis::ddsa_lib::bridge::ts_node::TsNodeBridge;
     use crate::analysis::ddsa_lib::bridge::ContextBridge;
-    use crate::analysis::ddsa_lib::common::{attach_as_global, get_field};
+    use crate::analysis::ddsa_lib::common::{attach_as_global, get_field, v8_interned};
     use crate::analysis::ddsa_lib::test_utils::{cfg_test_runtime, try_execute};
     use crate::analysis::ddsa_lib::RawTSNode;
     use crate::analysis::tree_sitter::get_tree_sitter_language;
@@ -235,10 +236,11 @@ mod tests {
             get_field::<v8::Integer>(obj, name, scope, "integer").unwrap().value() as usize == other
         };
         // We intentionally do not check `id` here because that is our abstraction, not tree-sitter's.
-        equals("_startLine", node.start_position().row)
-            && equals("_startCol", node.start_position().column)
-            && equals("_endLine", node.end_position().row)
-            && equals("_endCol", node.end_position().column)
+        // Because we also normalize the `tree_sitter::Point` to be 1-based, we need to do that here.
+        equals("_startLine", node.start_position().row + 1)
+            && equals("_startCol", node.start_position().column + 1)
+            && equals("_endLine", node.end_position().row + 1)
+            && equals("_endCol", node.end_position().column + 1)
             && equals("_typeId", node.grammar_id() as usize)
     }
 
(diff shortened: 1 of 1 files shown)


crates/static-analysis-kernel/src/analysis/ddsa_lib.rs
```diff
@@ -15,3 +15,46 @@ pub(crate) use runtime::JsRuntime;
...
+
+    /// Constructs a [`tree_sitter::Node`] from the raw tree-sitter node.
+    ///
+    /// # Safety
+    /// The caller must ensure the that the [`ffi::Tree`](tree_sitter::ffi::Tree) this node comes from
+    /// has not been dropped.
+    unsafe fn to_node(&self) -> tree_sitter::Node {
+        tree_sitter::Node::from_raw(self.0)
+    }
+}
```
juli1, line 37: What is the value of this? Can't we just use `TSNode`?

crates/static-analysis-kernel/src/analysis/ddsa_lib/bridge/ts_node.rs
```diff
@@ -0,0 +1,359 @@
...
+use crate::analysis::tree_sitter::{TSCaptureContent, TSQueryCapture};
+use deno_core::v8;
+use deno_core::v8::HandleScope;
+use std::marker::PhantomData;
+
+/// A stateful bridge holding a collection of [`RawTsNode`].
+#[derive(Debug)]
+pub struct TsNodeBridge {
+    mirrored_im: MirroredIndexMap<RawTSNode, NodeId>,
+    js_class: js::TreeSitterNodeFn<Class>,
```
juli1, line 18: What is the goal of the `js_class`?

crates/static-analysis-kernel/src/analysis/ddsa_lib/js/ts_node.rs
```diff
@@ -0,0 +1,80 @@
...
+
+/// A deserialized JavaScript object representation of a [`tree_sitter::Node`].
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Hash)]
+pub struct TreeSitterNode<T> {
+    pub id: u32,
+    pub start_line: u32,
+    pub start_col: u32,
+    pub end_line: u32,
+    pub end_col: u32,
+    pub node_type_id: u16,
```
juli1, line 20: Why not reusing the position object we have?

juli1 (approved):
1. Let's make it platform-independent
2. Keep the `start` and `end` as `Position` object

PR #393, opened 2024-06-07
[STAL-1960] Implement ddsa QueryMatch + bridge

jasonforal:
To execute a rule, the ddsa runtime needs to pass the [`tree_sitter::QueryMatch`](https://github.com/tree-sitter/tree-sitter/blob/604d38e6b327ed33877e1285680b505b9484a71c/lib/binding_rust/lib.rs#L192) to JavaScript.

### QueryMatch
The [`QueryMatch` JavaScript class](https://github.com/DataDog/datadog-static-analyzer/blob/123f840463f6d1b9cd81c586c59ef5e7823ce87c/crates/static-analysis-kernel/src/analysis/ddsa_lib/js/query_match.js#L8) is a wrapper around an array of `NamedCapture` (which were introduced in https://github.com/DataDog/datadog-static-analyzer/pull/391).
It has two getters:

---
* `get`: this replaces our current pattern of property access on `captures`.

**Old**
```js
// For a hypothetical capture named "func_param"
const { func_param } = query.captures;
const func_param_node = query.captures["func_param"];
const func_param_node = query.captures.func_param;
```
**New**
```js
const func_param = captures.get("func_param");
```

---

* `getMany`: this replaces our current pattern of property access on `capturesList`.

**Old**
```js
// For a hypothetical capture named "id" (that contains multiple nodes)
const { id } = query.capturesList;
const id_nodes = query.capturesList["id"];
const id_nodes = query.capturesList.id;
```
**New**
```js
const id_nodes = captures.getMany("id");
```

---

The set of named captures is represented as an array, _not_ a hashmap mainly for consistency -- we will be storing these on the Rust side in a `Vec`. From a performance standpoint, in theory, creating a small array with exact capacity (as opposed to a map with unknown capacity), and iterating a small array (as opposed to a hash + map lookup) should be faster than a JavaScript `Map`. _(Big O complexity analysis isn't relevant here, as N will be tiny, e.g. < 10)_.

### Compatibility
A compatibility layer has been implemented for JavaScript syntax. Existing rule code will function exactly as it currently does (albeit with a slight performance penalty).
```
const { func_param } = query.captures; // Valid
const func_param_node = query.captures["func_param"]; // Valid
const func_param_node = query.captures.func_param; // Valid
//
const func_param_node = query.captures.get("func_param"); // Valid
```

This is achieved in two steps:
1. A [JavaScript proxy](https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Global_Objects/Proxy) was added that wraps a `QueryMatch` class instance. Whenever a rule writes `captures["..."]`, the proxy intercepts this and [uses our logic](https://github.com/DataDog/datadog-static-analyzer/pull/393/commits/ce9b0aacfd5089c7a19fe90f159e34ca4f27dff6#diff-3297c914c4c3afd6e426e312d87a0fca4d37ff27c8abe1991fff92fecea587faR97-R119) to determine how to access the value. This has a slight performance penalty.
2. On the Rust side, we make this more ergonomic with a [newtype wrapper: `StellaCompat`](https://github.com/DataDog/datadog-static-analyzer/pull/393/commits/ce9b0aacfd5089c7a19fe90f159e34ca4f27dff6#diff-b6424b794288792913f6e5108c4a1bddbbb336aec1aed5befc5a4327332c5c70R44).

### Bridge
The bridge is a mostly straightforward `MirroredVec` -- the only thing to note is that in order to set the data, we need a [mutable reference to the `TsNodeBridge`](https://github.com/DataDog/datadog-static-analyzer/blob/123f840463f6d1b9cd81c586c59ef5e7823ce87c/crates/static-analysis-kernel/src/analysis/ddsa_lib/bridge/query_match.rs#L34). This is because [each named capture contains a NodeId](https://github.com/DataDog/datadog-static-analyzer/blob/123f840463f6d1b9cd81c586c59ef5e7823ce87c/crates/static-analysis-kernel/src/analysis/ddsa_lib/js/capture.rs#L45), and we need the TsNodeBridge to assign a `NodeId`.

### DIFF
--- a/crates/static-analysis-kernel/src/analysis/ddsa_lib/bridge/query_match.rs
+++ b/crates/static-analysis-kernel/src/analysis/ddsa_lib/bridge/query_match.rs
@@ -3,7 +3,7 @@
 // Copyright 2024 Datadog, Inc.
 
 use crate::analysis::ddsa_lib::bridge::ts_node::TsNodeBridge;
-use crate::analysis::ddsa_lib::common::{Class, DDSAJsRuntimeError, NodeId};
+use crate::analysis::ddsa_lib::common::{Class, DDSAJsRuntimeError, NodeId, StellaCompat};
 use crate::analysis::ddsa_lib::js;
 use crate::analysis::ddsa_lib::v8_ds::MirroredVec;
 use crate::analysis::tree_sitter::QueryMatch;
@@ -11,7 +11,9 @@ use deno_core::v8;
 use deno_core::v8::HandleScope;
 
 /// A stateful bridge holding a collection of [`QueryMatch<NodeId>`].
-pub struct QueryMatchBridge(MirroredVec<QueryMatch<NodeId>, js::QueryMatch<Class>>);
+pub struct QueryMatchBridge(
+    MirroredVec<StellaCompat<QueryMatch<NodeId>>, js::QueryMatchCompat<Class>>,
+);
 
 impl QueryMatchBridge {
     /// Constructs a new `QueryMatchBridge` for the given `scope`. The scope's [`v8::Context::global`] must
@@ -24,7 +26,7 @@ impl QueryMatchBridge {
         /// Even so, we allocate an (arbitrary non-zero) initial capacity, as we know all executions will contain query matches.
         const CAPACITY: u32 = 16;
 
-        let js_class = js::QueryMatch::try_new(scope)?;
+        let js_class = js::QueryMatchCompat::try_new(scope)?;
         Ok(Self(MirroredVec::with_capacity(js_class, scope, CAPACITY)))
     }
 
@@ -49,6 +51,7 @@ impl QueryMatchBridge {
                     .into_iter()
                     .map(|capture| node_bridge.insert_capture(scope, capture))
                     .collect::<Vec<_>>()
+                    .into()
             })
             .collect::<Vec<_>>();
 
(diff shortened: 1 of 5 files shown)


crates/static-analysis-kernel/src/analysis/ddsa_lib/bridge/query_match.rs
```diff
@@ -0,0 +1,145 @@
...
+/// A stateful bridge holding a collection of [`QueryMatch<NodeId>`].
+pub struct QueryMatchBridge(MirroredVec<QueryMatch<NodeId>, js::QueryMatch<Class>>);
+
+impl QueryMatchBridge {
+    /// Constructs a new `QueryMatchBridge` for the given `scope`. The scope's [`v8::Context::global`] must
+    /// have class functions with the following identifiers:
+    /// * [`js::QueryMatch::CLASS_NAME`]
+    pub fn try_new(scope: &mut HandleScope) -> Result<Self, DDSAJsRuntimeError> {
+        let js_class = js::QueryMatch::try_new(scope)?;
+        Ok(Self(MirroredVec::with_capacity(js_class, scope, 16)))
```
juli1, line 22: Let's put `16` somewhere as a constant and explains why we choose this number

crates/static-analysis-kernel/src/analysis/ddsa_lib/js/query_match.rs
```diff
@@ -0,0 +1,227 @@
...
+/// A function representing the ES6 class `QueryMatch`.
+#[derive(Debug)]
+pub struct QueryMatch<T> {
+    class: v8::Global<v8::Function>,
+    single_capture: SingleCaptureTemplate,
+    multi_capture: MultiCaptureTemplate,
+    _pd: PhantomData<T>,
+}
+
+rust_converter!(
```
juli1, line 26: 🔥

juli1 (commented):
Why changing `const func_param_node = query.captures["func_param"];` and `const func_param_node = query.captures.func_param;` with `const func_param = captures.get("func_param");`? Can't we just avoid backward-incompatible behavior?

juli1 (approved):

PR #398, opened 2024-06-17
[STAL-1960] Implement ddsa JsRuntime scoped execution

jasonforal:
We need a high performance way to execute the JavaScript portion of a static analysis rule, so we are re-using the same v8 isolate across executions.

The ideal model for us is to create a new [`Context`](https://v8docs.nodesource.com/node-20.3/df/d69/classv8_1_1_context.html) and [`Context::Scope`](https://v8docs.nodesource.com/node-20.3/de/d41/classv8_1_1_context_1_1_scope.html) for every rule execution. v8 Contexts ([background info here](https://v8.dev/docs/embed#contexts)) are relatively cheap to create, and as soon as they drop out of scope, all local handles are able to be garbage collected.

Each context has effectively its own state, and each context its own global environment. This is problematic for us, because when we bootstrap the `deno_core` runtime, we populate the v8 isolate's default context's global environment with DDSA-specific variables (see [`__bootstrap.js`](https://github.com/DataDog/datadog-static-analyzer/blob/a2aea86950b006ecce5dc67b0e14c0d35a38c4b7/crates/static-analysis-kernel/src/analysis/ddsa_lib/js/__bootstrap.js#L15-L21)).

Because of this, we either need to share a global environment across contexts, or allocate and re-assign the global object with all of our needed values every time we create a new context. I previously didn't know v8 well enough to know how to achieve the former (hence why I [re-used a v8 Context](https://github.com/DataDog/datadog-static-analyzer/blob/224f9bf515a898c877e19d38bc0db08074964ec4/crates/static-analysis-kernel/src/analysis/javascript.rs#L160) across executions and [used a hacky closure-based solution](https://github.com/DataDog/datadog-static-analyzer/blob/a2aea86950b006ecce5dc67b0e14c0d35a38c4b7/crates/static-analysis-kernel/src/analysis/ddsa_lib/js/stella_compat.js#L83-L88) to simulate different v8 Contexts).

This PR addresses this problem.

### Ideal solution
The ideal solution would be to pass in a handle to a pre-allocated global object ("global_object") when the v8 context is created. This is possible in the C++ API. We'd do something like the following:
```cpp
auto ddsa_global = prev_ctx->Global();
prev_ctx->Exit();
prev_ctx->DetachGlobal();
auto new_ctx = v8::Context::New(/* ...isolate, ...template */, ddsa_global)
```
([See documentation for `static Local< Context > New`](https://v8docs.nodesource.com/node-20.3/df/d69/classv8_1_1_context.html#af43e9532271b5c7afb4e5b816e1b8b3c))

However, this is not possible with `rusty_v8`. As of writing, `v8::Context::new` will [always pass in null](https://github.com/denoland/rusty_v8/blob/3ec0ad6ea1228a98107a68113602a27133495432/src/context.rs#L78) for the "global_object" parameter.

`rusty_v8` does have `v8::Context::new_from_template`, which would let us [pass in an `ObjectTemplate`](https://github.com/denoland/rusty_v8/blob/3ec0ad6ea1228a98107a68113602a27133495432/src/context.rs#L92), however that isn't ideal for two reasons: 1) it would allocate an entirely new object and 2) ObjectTemplates can only store primitives or other ObjectTemplates/FunctionTemplates. 2 would be a significant thorn because the `ddsa` lib is designed around stateful "bridge" objects being available to all JavaScript executing, and hydrating their state from scratch would be complicated.

Thus, to get around `rusty_v8`'s inability to pass in a pre-allocated object to be used as the global object, we simulate this behavior by manually setting the [prototype](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Objects/Object_prototypes) of a context's global proxy object to our pre-allocated object. In order to re-use this object across contexts, we have to make it a `v8::Global`.

Wait, global this, global proxy that...uhh...what? 

### "Global"
Yes, this terminology is quite confusing. Quick summary:
* `v8::Global` - The name `rusty_v8` has given to a v8 [`Persistent` handle](https://v8docs.nodesource.com/node-20.3/d4/dca/classv8_1_1_persistent_base.html#ad845ec8872174be0a9ca9a3dd1898d30). TL;DR a value that has been forgotten by the garbage collector and so we can re-use it between v8 Contexts.
* Global Object - The "true" global object, which holds all v8 global values (functions and variables) for a specific v8 Context. See v8's [`bootstrapper.cc`](https://github.com/v8/v8/blob/55ef0f9991e3ebda4b3213caf67b7f687593e6bf/src/init/bootstrapper.cc#L230) for some additional commentary.
* Global Proxy - An object that intercepts property access to the "true" global object. v8 uses this for encapsulation and security. In JavaScript, this is exposed as the variable `globalThis`. `rusty_v8` somewhat imprecisely calls this "global" (accessed with `Context::global`).

So in sum, we are creating a "persistent" v8 object, and we configure each v8 Context to use this v8 object. Within the execution context, mutations to this v8 object can happen, but they never persist beyond the context. Executions are thus isolated from each other.

### Benefits
By doing this, we no longer have to [use a closure to encapsulate](https://github.com/DataDog/datadog-static-analyzer/blob/a2aea86950b006ecce5dc67b0e14c0d35a38c4b7/crates/static-analysis-kernel/src/analysis/javascript.rs#L136-L153) the rule's code. I suspect this will be a lot easier for v8's JIT to optimize.

* A future PR (edit: https://github.com/DataDog/datadog-static-analyzer/pull/402) will introduce the `execute_rule` function on `JsRuntime`, which will be a thin wrapper around the `scoped_execute` function introduced in this PR.
* A helper function for creating a `v8::UnboundScript` was added. UnboundScripts will be used so we only have to compile a v8 script once per rule, not once per execution as we currently do.

### DIFF
--- a/crates/static-analysis-kernel/src/analysis/ddsa_lib.rs
+++ b/crates/static-analysis-kernel/src/analysis/ddsa_lib.rs
@@ -10,6 +10,7 @@ pub use context::*;
 pub mod extension;
 pub(crate) mod js;
 pub(crate) mod ops;
+#[allow(dead_code)]
 pub(crate) mod runtime;
 pub(crate) use runtime::JsRuntime;
 #[allow(dead_code)]


crates/static-analysis-kernel/src/analysis/ddsa_lib/runtime.rs
```diff
@@ -67,3 +185,156 @@ impl JsConsole {
...
+
+        let prop_diff = default_ctx_global_proxy_props
+            .symmetric_difference(&scoped_global_proxy_props)
+            .collect::<HashSet<&String>>();
+        assert!(prop_diff.is_empty());
+        assert_ne!(default_ctx_id_hash, scoped_exe_id_hash);
+    }
+
+    /// Ensures that properties edited on `globalThis` do not persist between executions.
+    #[rustfmt::skip]
```
juli1, line 258: why skipping this?

crates/static-analysis-kernel/src/analysis/ddsa_lib/common.rs
```diff
@@ -12,10 +12,13 @@ pub type NodeId = u32;
 
 #[derive(Debug, thiserror::Error)]
 pub enum DDSAJsRuntimeError {
+    #[error("execution error: {reason}")]
+    Execution { reason: String },
     #[error("expected `{name}` to exist within the v8 context")]
     VariableNotFound { name: String },
     #[error("type should be \"{expected}\", not \"{got}\"")]
     WrongType {
+        identifier: Option<String>,
```
juli1, line 21: The `identifier` should not be used above. Should we make use of it?

crates/static-analysis-kernel/src/analysis/ddsa_lib/common.rs
```diff
@@ -232,17 +239,52 @@ pub fn set_key_value<G>(
...
+            .exception()
+            .expect("return value should only be `None` if an error was caught");
+        let reason = exception.to_rust_string_lossy(tc_scope);
+        tc_scope.reset();
+        DDSAJsRuntimeError::Interpreter { reason }
+    })?;
+
+    let unbound_script = script.get_unbound_script(tc_scope);
+    Ok(v8::Global::new(tc_scope, unbound_script))
+}
```
juli1, line 262: It seems this function is only used for tests purposes. Should we either?

1. Only enable it for tests?
2. Put a comment to explain this?

crates/static-analysis-kernel/src/analysis/ddsa_lib/runtime.rs
```diff
@@ -4,22 +4,134 @@
...
     console: Rc<RefCell<JsConsole>>,
+    // v8-specific
+    /// A JavaScript "global" object (i.e. `globalThis`) augmented with ddsa variables. This is _not_
+    /// a global proxy object, but rather, the global object itself.
+    ///
+    /// This is cached so that whenever we create a new v8 context, we can explicitly set this as the
+    /// prototype of the context's global proxy object, exposing all ddsa globals to the v8 context.
+    /// Without caching, we'd have to create this object every time.
+    ///
+    /// NOTE: This is a hack to work around the fact `rusty_v8` does not currently have bindings for 100% of the C++ API.
```
juli1, line 23: Should we open an issue on https://github.com/denoland/rusty_v8/issues ?

juli1 (approved):
See comments and please address them if necessary.

PR #399, opened 2024-06-18
[STAL-1960] Return TreeSitterNode instances from QueryMatch `get` and `getMany`

jasonforal:
Currently, the JavaScript runtime returns a [`NodeId`](https://github.com/DataDog/datadog-static-analyzer/blob/ea3f0a01c83bd55ba6507ab55925688b4d3c0c12/crates/static-analysis-kernel/src/analysis/ddsa_lib/js/ts_node.js#L7-L12) or array of `NodeId` when calling QueryMatch.get or QueryMatch.getMany.

```js
const value = captures.get("cap_name");
value === 2; // true
```

We will be exposing `QueryMatch` directly to the `visit` function in the JavaScript API, so we don't want to require users to write boilerplate to look up the NodeId.

## What is your solution?
The getter now looks up the `TreeSitterNode` instance stored on the TsNodeBridge (`__RUST_BRIDGE__ts_node`). The TsNodeBridge guarantees that this node will exist.

```js
const value = captures.get("cap_name");
value instanceof TreeSitterNode; // true
```

* **Don't pay attention to the superficial `QueryMatchCompat` changes**. They are replaced with https://github.com/DataDog/datadog-static-analyzer/pull/400. ("Why not combine them?" I want to keep this PR separate in order to make the code history more clear to facilitate eventually removing `QueryMatchCompat`).
* A future PR will properly "mount" the bridge at `__RUST_BRIDGE__ts_node`.

### DIFF
--- a/crates/static-analysis-kernel/src/analysis/ddsa_lib/js/query_match.js
+++ b/crates/static-analysis-kernel/src/analysis/ddsa_lib/js/query_match.js
@@ -19,6 +19,20 @@ export class QueryMatch {
     }
 
     /**
+     * Returns the node with the given capture name, following semantics from {@link QueryMatch._getId}
+     * @param {string} name
+     * @returns {TreeSitterNode | undefined}
+     */
+    get(name) {
+        const nodeId = this._getId(name);
+        if (nodeId === undefined) {
+            return undefined;
+        }
+        // Note: the Rust bridge guarantees that this map element will be present (assuming `_captures` was not mutated).
+        return globalThis.__RUST_BRIDGE__ts_node.get(nodeId);
+    }
+
+    /**
      * Returns the id of the node with the given capture name. If there are multiple matching captures,
      * only the last will be returned, and the rest will be silently ignored.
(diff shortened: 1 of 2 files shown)


crates/static-analysis-kernel/src/analysis/ddsa_lib/common.rs
```diff
@@ -232,17 +239,52 @@ pub fn set_key_value<G>(
...
+            .exception()
+            .expect("return value should only be `None` if an error was caught");
+        let reason = exception.to_rust_string_lossy(tc_scope);
+        tc_scope.reset();
+        DDSAJsRuntimeError::Interpreter { reason }
+    })?;
+
+    let unbound_script = script.get_unbound_script(tc_scope);
+    Ok(v8::Global::new(tc_scope, unbound_script))
+}
```
juli1, line 262: Same comment than in other PR (maybe put this in tests).

juli1 (approved):

PR #403, opened 2024-06-20
[STAL-1960] Implement compatibility layer for Stella visit function

jasonforal:
The stella runtime uses a visit function with three arguments:
```js
function visit(queryMatch, filename, code) { /* ... */ }
```

ddsa uses a visit function with one argument:
```js
function visit(queryMatch) { /* ... */ }
```

This is because ddsa exposes the `filename` and `code` as part of the RootContext, not as parameters to the visit function.

We need stella-style visit functions to function as normal (as well as ddsa-style visit functions).

## What is your solution?
The way we've set up the ddsa RootContext, we utilize a deno `op` to lazily fetch the filename or the file contents because it allows us to avoid an allocation for rules that never end up accessing the filename or the file contents.

With the stella function signature, though, if we were to pass in strings for `filename` and `code` into the visit function, we'd have to trigger the string allocation even if the rule never uses the string.

The solution implemented in this PR is to use JavaScript [Proxy](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy) objects for the `filename` and `code`. We allocate a single global object for the filename and a single global object for the file contents. These objects do nothing until accessed.

So a rule like this will not trigger a string allocation:
```js
function visit(query, filename, code) { 
    const node = query.captures["cap_name"];
    const errorMessage = "error message";
    const error = buildError(node.start.line, node.start.col, node.end.line, node.end.col, errorMessage);
    addError(error);
}
```

But a rule like this would:
```js
function visit(query, filename, code) { 
    const node = query.captures["cap_name"];
    // Note the access of the "code" variable.
    const errorMessage = `node text: ${getCodeForNode(node, code)}`; 
    const error = buildError(node.start.line, node.start.col, node.end.line, node.end.col, errorMessage);
    addError(error);
}
```

Note: what we pass in as `filename` and `code` are technically objects (`typeof filename === "object" // true`), but any user code that uses them can interact with them like string primitives. So any rules that do things like `filename.contains("abc")` or `filename.length` or `filename.charAt(0)` "just work" (this is possible because [we bind any function calls to an actual string primitive](https://github.com/DataDog/datadog-static-analyzer/commit/cb26a7c3b1790af29a5f8caf08ea096ad30f5cbc#diff-96846e03f040296c97f5e8eb6e120ac71ebbfc500a147f36fa9b41bd8de25351R131), so to the caller, they might as well be interacting with a string).

We thus preserve the performance of ddsa while still supporting the old style of rule visit function. We will eventually transition all rules to the new ddsa visit function, at which point we can easily remove this compatibility layer.
* You can confirm behavior parity: A test asserts a [ddsa-style rule execution](https://github.com/DataDog/datadog-static-analyzer/commit/cb26a7c3b1790af29a5f8caf08ea096ad30f5cbc#diff-994c58d79e1d82b4513246e0ed000503007a0ba771b3f9d7569fb79c7ed94efaR611-R621) returns a specific violation. Another test asserts a [stella-style rule execution](https://github.com/DataDog/datadog-static-analyzer/pull/403/files#diff-994c58d79e1d82b4513246e0ed000503007a0ba771b3f9d7569fb79c7ed94efaR651-R662) returns the exact same violation. The rule logic is identical except for how the variables are accessed.
* We allocate these Proxy objects _once_ per runtime. This is possible because the proxy just delegates to the `RootContext`'s getter. So the Proxy object can be stateless, and we don't need to re-create or manage its value between different executions.

### DIFF
--- a/crates/static-analysis-kernel/src/analysis/ddsa_lib/runtime.rs
+++ b/crates/static-analysis-kernel/src/analysis/ddsa_lib/runtime.rs
@@ -26,8 +26,8 @@ const BRIDGE_QUERY_MATCH: &str = "__RUST_BRIDGE__query_match";
 const BRIDGE_TS_NODE: &str = "__RUST_BRIDGE__ts_node";
 const BRIDGE_TS_SYMBOL: &str = "__RUST_BRIDGE__ts_symbol_lookup";
 const BRIDGE_VIOLATION: &str = "__RUST_BRIDGE__violation";
-const STELLA_COMPAT_FILENAME: &str = "__STELLA_COMPAT__filename";
-const STELLA_COMPAT_FILE_CONTENTS: &str = "__STELLA_COMPAT__file_contents";
+const STELLA_COMPAT_FILENAME: &str = "STELLA_COMPAT_FILENAME";
+const STELLA_COMPAT_FILE_CONTENTS: &str = "STELLA_COMPAT_FILE_CONTENTS";
 
 /// The Datadog Static Analyzer JavaScript runtime
 pub struct JsRuntime {
@@ -325,7 +325,7 @@ impl JsRuntime {
         format!(
             r#"
 for (const queryMatch of globalThis.__RUST_BRIDGE__query_match) {{
-    visit(queryMatch, globalThis.__STELLA_COMPAT__filename, globalThis.__STELLA_COMPAT__file_contents);
+    visit(queryMatch, globalThis.STELLA_COMPAT_FILENAME, globalThis.STELLA_COMPAT_FILE_CONTENTS);
 }}
 
 // The rule's JavaScript code


crates/static-analysis-kernel/src/analysis/ddsa_lib/runtime.rs
```diff
@@ -310,7 +325,7 @@ impl JsRuntime {
         format!(
             r#"
 for (const queryMatch of globalThis.__RUST_BRIDGE__query_match) {{
-    visit(queryMatch);
+    visit(queryMatch, globalThis.__STELLA_COMPAT__filename, globalThis.__STELLA_COMPAT__file_contents);
```
juli1, line 328: Can't we use `STELLA_COMPAT_FILENAME` and `STELLA_COMPAT_FILE_CONTENTS`?

juli1 (approved):
one minor comment

PR #406, opened 2024-06-24
[STAL-2335] ci: fix error when multiple repos have potential regressions detected

amaanq:

Our CI for checking regressions currently tries to upload the files that have potential regressions under the names `failures-before` and `failures-after`. This is problematic because if two repos fail, both steps will try to upload an artifact with the same name, which is not allowed as GH artifacts must have unique names on a per-run basis.

## What is your solution?

The solution is to simply tack on the repo name that is failing to the uploaded artifact name. For example, if `golang/go` fails with potential regressions detected, the artifact names for uploaded files here would be `failures-before-golang-go` and `failures-after-golang-go`

## Alternatives considered

N/A

The changes are pretty minimal, I simply broke up the repo matrix into two parts since GH does not have a split function for us to extract each component and use a `-` instead of a `/` for the artifact name.


amaanq:
rebasing with no changes for cleaner history, then merging

PR #412, opened 2024-06-25
[STAL-1960] Optimize memory usage of JS timeout functionality

jasonforal:
A "watchdog" thread is currently spawned for every JavaScript execution in order to terminate v8 execution if a script runs for too long.

From a CPU standpoint, while theoretically inefficient, this is mostly a non-issue. V8 is single-threaded, and so we only ever spawn watchdog threads sequentially, meaning no contention spawning these threads and minimal context-switching overhad.

However, from a memory standpoint, this is highly inefficient, as we spawn `O(N x M)` threads, where `N` is the number of files and `M` the number of rules. The OS needs to allocate stack space for each thread, so this adds up very quickly (even if we tune the stack space allocated to the OS minimum).

## What is your solution?
Spawn a single watchdog thread per `JsRuntime`, reducing memory allocations from `O(N x M)` to `O(1)`. A watchdog thread is "bound" to the runtime, and only monitors executions for that specific runtime.

### Before
Here's an allocation profile spawning a watchdog thread-per-execution:
multiple-watchdog

**All Anonymous VM**
| # Persistent | # Transient | Total Bytes |
|-|-|-|
|16|382,810|346.26 GiB|

Note the O(N x M) behavior in the "All Anonymous VM" allocations - these are from spawning watchdog threads
spawn-watchdog-thread

### After
And an allocation profile with spawning a single watchdog thread:
single-watchdog

**All Anonymous VM**
| # Persistent | # Transient | Total Bytes |
|-|-|-|
|15|89|96.73 MiB|

### Impact: 238s -> 199s (~15% wall clock speedup), 99%+ allocation reduction
I was not expecting the CPU speedup to be so significant -- it's a pleasant surprise.

The test case was as follows:
```
Analyzing 2075 JavaScript files using 7 rules
Analyzing 66178 TypeScript files using 15 rules
```
Meaning we did 14525 + 992670 = 1,007,195 executions.

I suspect what is happening here is that v8 execution is so fast (either from genuine ddsa runtime speed improvements, or just from small files and simple rules) that the overhead of spawning a thread and synchronizing with it slows down the execution. The absolute time delta is 39s. With 1 million executions, that suggests the overhead of spawning a thread per-execution is **39 microseconds**, which seems entirely plausible to me.

### Technical Notes

The implementation of this solution is quite simple because we don't have to worry about concurrency: a `JsRuntime` exists on a single thread and v8 execution is single-threaded. Because the watchdog only needs to track sequential executions, we can get away with only a condvar and corresponding mutex to handle the bookkeeping for tracking timeouts.

* The graphs above are an apples-to-apples comparison -- both use the ddsa runtime, except one spawns thread-per-execution instead of per-runtime.

### DIFF
--- a/crates/static-analysis-kernel/src/analysis/ddsa_lib/runtime.rs
+++ b/crates/static-analysis-kernel/src/analysis/ddsa_lib/runtime.rs
@@ -19,7 +19,8 @@ use deno_core::v8;
 use std::cell::{RefCell, RefMut};
 use std::collections::HashMap;
 use std::rc::Rc;
-use std::sync::Arc;
+use std::sync::{Arc, Condvar, Mutex};
+use std::time::{Duration, Instant};
 
 const BRIDGE_CONTEXT: &str = "__RUST_BRIDGE__context";
 const BRIDGE_QUERY_MATCH: &str = "__RUST_BRIDGE__query_match";
@@ -38,6 +39,10 @@ pub(crate) const DEFAULT_REMOVED_GLOBAL_PROPS: &[&str] = &[
 /// The Datadog Static Analyzer JavaScript runtime
 pub struct JsRuntime {
     runtime: deno_core::JsRuntime,
+    /// Each `JsRuntime` spawns a thread that lives for as long as the `JsRuntime`, and it is used
+    /// to manually terminate JavaScript executions that go on for too long. Synchronization between
+    /// the `JsRuntime` and this thread is performed through this `watchdog_pair`.
+    watchdog_pair: Arc<(Mutex<JsExecutionState>, Condvar)>,
     console: Rc<RefCell<JsConsole>>,
     bridge_context: Rc<RefCell<ContextBridge>>,
     bridge_query_match: QueryMatchBridge,
@@ -150,8 +155,12 @@ impl JsRuntime {
         let console = Rc::new(RefCell::new(JsConsole::new()));
         op_state.put(Rc::clone(&console));
 
+        let v8_isolate_handle = runtime.v8_isolate().thread_safe_handle();
+        let watchdog_pair = spawn_watchdog_thread(v8_isolate_handle);
+
         Ok(Self {
             runtime,
+            watchdog_pair,
             console,
             bridge_context: context,
             bridge_query_match: query_match,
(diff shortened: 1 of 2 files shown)


crates/static-analysis-kernel/src/analysis/ddsa_lib/runtime.rs
```diff
@@ -387,6 +433,101 @@ for (const queryMatch of globalThis.__RUST_BRIDGE__query_match) {{
...
+    should_shut_down: bool,
+}
+
+impl JsExecutionState {
+    pub fn new() -> Self {
+        Self {
+            timeout_duration: Duration::default(),
+            start_time: Instant::now(),
+            execution_complete: true,
+            should_shut_down: false,
```
juli1, line 526: What about `watchdog_should_shut_down`?

crates/static-analysis-kernel/src/analysis/ddsa_lib/runtime.rs
```diff
@@ -387,6 +433,101 @@ for (const queryMatch of globalThis.__RUST_BRIDGE__query_match) {{
...
+    execution_complete: bool,
+    should_shut_down: bool,
+}
+
+impl JsExecutionState {
+    pub fn new() -> Self {
+        Self {
+            timeout_duration: Duration::default(),
+            start_time: Instant::now(),
+            execution_complete: true,
```
juli1, line 525: When we create the state, the execution is not complete. Why setting it to `true` then?

crates/static-analysis-kernel/src/analysis/ddsa_lib/runtime.rs
```diff
@@ -387,6 +433,101 @@ for (const queryMatch of globalThis.__RUST_BRIDGE__query_match) {{
...
+    timeout_duration: Duration,
+    start_time: Instant,
+    execution_complete: bool,
+    should_shut_down: bool,
+}
+
+impl JsExecutionState {
+    pub fn new() -> Self {
+        Self {
+            timeout_duration: Duration::default(),
```
juli1, line 523: Shouldn't we put the real value of the timeout here?

juli1 (commented):
I do not understand, there is a execute_rule_internal in [javascript.rs](http://javascript.rs/) and in [runtime.rs](http://runtime.rs/). It seems confusing. What is the plan to reconcile all of this?

juli1 (approved):

PR #417, opened 2024-06-28
Update op to only return named children

jasonforal:
Currently, when a rule requests the children of a tree-sitter node, they get both [anonymous and named nodes](https://tree-sitter.github.io/tree-sitter/using-parsers#named-vs-anonymous-nodes), which is unintuitive and does not follow the semantics of [what we currently do with the stella library](https://github.com/DataDog/datadog-static-analyzer/blob/e0f4f7685a69ac825a80f44055ff505dcf9c931f/crates/static-analysis-kernel/src/analysis/tree_sitter.rs#L317-L318).

```js
function echo(a, b, c) {}
```

For the query:
```
((function_declaration
    (formal_parameters) @paramList))
```

The (anonymous + named) children of `@paramList` are:
`(`, `a`, `,`, `b`, `,`, `c`, `)`

The named children are:
`a`, `b`, `c`

## What is your solution?
Use tree-sitter's [`named_children`](https://github.com/tree-sitter/tree-sitter/blob/9d1ac2139221b2e19389ca620767c537607a2f63/lib/binding_rust/lib.rs#L1268) API instead of `children`.

### DIFF
--- a/crates/static-analysis-kernel/src/analysis/ddsa_lib/extension.rs
+++ b/crates/static-analysis-kernel/src/analysis/ddsa_lib/extension.rs
@@ -10,7 +10,7 @@ deno_core::extension!(
         ops::op_current_filename,
         ops::op_console_push,
         ops::op_current_ts_tree_text,
-        ops::op_ts_node_children,
+        ops::op_ts_node_named_children,
         ops::op_ts_node_text,
     ],
     esm_entry_point = "ext:ddsa_lib/__bootstrap.js",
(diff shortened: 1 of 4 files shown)


juli1:
Not opposed to the change but how will we get the unnamed nodes? It's generally not useful for the analysis but it's often required for calculating the fixes (like the place of a comma, etc). We need a way to get the unnamed nodes. What about we name a `node.getChildren()` and `node.getNamedChildren()` ?

juli1 (approved):

PR #421, opened 2024-07-01
[STAL-2368] refactor: add a retry mechanism anywhere requests are made

amaanq:

This PR will help avoid transient/flaky failures in our CI, as seen [here](https://github.com/DataDog/datadog-static-analyzer/actions/runs/9746106627/job/26895629440#step:7:269). The problem is we don't have a retry mechanism in case a request fails due to transient errors/flakiness that is not our fault, which is not ideal because having a retry or two could fix this in some cases.

## What is your solution?

I've added a retry mechanism in three spots that has an exponential backoff, so each successive failure will double the delay in between the next retry, up to five time. Initially the delay is 1 second, and at most it will be 16 seconds (on the 5th attempt). The spots covered are when we fetch the repository from GitHub (an observed failure here is the reason for this ticket), and when we query our API for rules or diff aware.

I've refactored the request logic + retry mechanism into the `perform_request` function in `datadog_utils.rs` since we have two different spots where we make a request.

### DIFF
--- a/crates/cli/src/datadog_utils.rs
+++ b/crates/cli/src/datadog_utils.rs
@@ -113,14 +113,33 @@ fn make_request(
     }
 }
 
+fn perform_request(request_builder: RequestBuilder) -> Result<reqwest::blocking::Response> {
+    let mut server_response = None;
+    let mut retry_time = 1;
+    for _ in 0..5 {
+        match request_builder.try_clone().unwrap().send() {
+            Ok(r) => server_response = Some(r),
+            Err(_) => {
+                eprintln!("Error when querying the datadog server");
+                std::thread::sleep(std::time::Duration::from_secs(retry_time));
+                retry_time *= 2; // Exponential backoff
+            }
+        }
+    }
+    let Some(server_response) = server_response else {
+        return Err(anyhow!("Error when querying the datadog server"));
+    };
+
+    Ok(server_response)
+}
+
 // get rules from one ruleset at datadog
 // it connects to the API using the DD_SITE, DD_APP_KEY and DD_API_KEY and retrieve
 // the rulesets. We then extract all the rulesets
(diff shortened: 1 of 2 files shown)


crates/cli/src/datadog_utils.rs
```diff
@@ -113,14 +113,33 @@ fn make_request(
...
 }
 
+fn perform_request(request_builder: RequestBuilder) -> Result<reqwest::blocking::Response> {
+    let mut server_response = None;
+    let mut retry_time = 1;
+    for _ in 0..5 {
+        match request_builder.try_clone().unwrap().send() {
+            Ok(r) => server_response = Some(r),
+            Err(_) => {
+                eprintln!("Error when querying the datadog server");
```
juli1, line 123: 1. This should only be shown if we run in debug mode.
2. We should show more information: what request has been made? what url?

amaanq: I was debating adding additional information, e.g. the error message, should we here? FYI this is if the request actually fails (internet issue or whatever), not if it's okay but we get a 404/500 etc (those errors are handled after this call)

juli1: 1. Not descriptive enough. Our errors should describe the problem and help the user (good developer tools help developers, think about the rust compiler vs. any C compiler)
2. The real error is reported in the main function if it fails and if we add more debugging, we should only add it if `configuration.use_debug` is `true`

amaanq: sounds good, and I agree. only one annoying point is that we pass `debug` down many function calls. It might be better to transform the request methods to be struct methods of our own `Request` type that wraps reqwest, and we can easily store `debug` here once instead of passing it down many function calls, but this isn't needed though

amaanq:
squashing & merging

PR #434, opened 2024-07-03
[STAL-2379] docs: document how to use the Docker container

amaanq:

We do not have any documentation on how to use our Docker container

## What is your solution?

Add docs

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,14 @@ datadog-static-analyzer is the static analyzer engine for Datadog [static analys
 datadog-static-analyzer --directory /path/to/directory --output report.csv --format csv
 ```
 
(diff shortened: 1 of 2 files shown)


juli1:
We should update the [Quickstart](https://github.com/DataDog/datadog-static-analyzer?tab=readme-ov-file#quick-start) section of the `README.md` and show the command to use the container directly from there.

This way, people do not have to download anything. Docker is common among everybody and users can try the analyzer in one command-line.

juli1:
We should also write about using a pinned version. A lot of user want to use a pinned version to prevent regressions. We should put something about it here.

juli1 (approved):

PR #436, opened 2024-07-05
StaticAnalysisConfigFile: do not error if the configuration is empty

albertvaka:
Do not error if the user deletes everything in the config file, then tries to onboard from scratch again. The previous code would fail because an empty string is technically invalid YAML.

## What is your solution?
When we see an empty string as the config, assume there's no config.

This code is used from the IDE-plugin-facing API to modify the config file.

### DIFF
--- a/crates/bins/src/bin/datadog_static_analyzer_server/ide/configuration_file/static_analysis_config_file.rs
+++ b/crates/bins/src/bin/datadog_static_analyzer_server/ide/configuration_file/static_analysis_config_file.rs
@@ -22,6 +22,9 @@ impl TryFrom<String> for StaticAnalysisConfigFile {
 
     fn try_from(base64_str: String) -> Result<Self, Self::Error> {
         let content = decode_base64_string(base64_str)?;
+        if content.trim().is_empty() {
+            return Ok(Self(ConfigFile::default()));
+        }
         let config = parse_config_file(&content)?;
         Ok(config.into())
     }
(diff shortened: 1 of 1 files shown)


robertohuertasm (approved):
LGTM 🚀

Some suggestions for your consideration:

I would add a new test to cover the new functionality of the `TryFrom` impl:
```rust
 #[test]
    fn try_from_returns_default_config_file_if_empty_string() {
        let expected = StaticAnalysisConfigFile::default();
        let config = StaticAnalysisConfigFile::try_from(String::new()).unwrap();
        assert_eq!(config, expected);
    }
```

No need to create a new module for that, we can include it in the test `mod`.

Note that this tests assumes several things: we **must** have implemented several traits like `Default` and `PartialEq`.

We can leverage the fact that `ConfigFile` already implements those and implement all the ones that the inner type implements by adding this macro to the `StaticAnalysisConfigFile` type:

```rust
#[derive(Debug, Default, Clone, PartialEq)]
```

If we do that, we can simplify line 26 like this:
```rust
return Ok(Self::default());
```
So, aside from the extra test, the changes would be good for the `StaticAnalysisConfigFile` [newtype](https://doc.rust-lang.org/book/ch19-03-advanced-traits.html#using-the-newtype-pattern-to-implement-external-traits-on-external-types) to implement the same traits as the type it wraps.

albertvaka:
@robertohuertasm thanks for the tips. I've pushed some changes. One question, though: did I really need all of `Debug`, `Default`, `Clone` and `PartialEq`, or only `Default` was needed and the rest you added because it's good to always have them?

robertohuertasm:
Regarding your question, the only "needed" one was `PartialEq` as it was helpful for the test assertion. The other ones were more a nice-to-have. `Default` was convenient to write less. `Debug` and `Clone` are not used, but it's nice to have them implemented.

It is a good practice to implement a few traits when you're dealing with public APIs to make your structs much more versatile and easier to work with:

- Debug: in order to display them nicely when you use format `{:?}`
- Clone: duplication
- PartialEq and Eq: comparison
- PartialOrd and Ord: ordering
- Default: provides a default value for your struct
- Serialize/Deserialize
- Hash: to use it as key in a hashmap

In our case, this struct is not strictly part of a public facing library so we can/could omit some implementations. The suggestions were indeed just a way to start a discussion about this, not really critical changes.

robertohuertasm (approved):
LGTM 🚀

PR #441, opened 2024-07-11
feat(docker): add github-action.sh CI script

amaanq:

We need a robust way to run our github action with a script that sets up the necessary environment variables and cli arguments without being too hacky with multi-step composite workflows.

## What is your solution?

Add a `github-action.sh` script (stripped-down version of this [entrypoint.sh](https://github.com/DataDog/datadog-static-analyzer-github-action/blob/main/entrypoint.sh) script) that will set up the necessary environment variables and arguments given the input.

## Alternatives considered

Use the composite workflow [here](https://github.com/DataDog/datadog-static-analyzer-github-action/pull/40)

I talked with Julien about this and this is the solution we came up with, the added benefits are that whenever we add new features to our analyzer, we can update this script here and the container will be updated in one go instead of having to update the github-action repo as well.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,7 +22,8 @@ RUN cargo build --release --bin datadog-static-analyzer
 
 FROM base
 
-COPY --from=build /app/target/release/datadog-static-analyzer /usr/local/bin/datadog-static-analyzer
+COPY --from=build /app/target/release/datadog-static-analyzer /usr/bin/datadog-static-analyzer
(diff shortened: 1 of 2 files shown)


juli1:
@amaanq out of curiosity, did you try to build the image locally? Can you confirm the script is correctly built/copied? I do not see the testing steps here.

amaanq:
Yes, I've built and ran the image locally and can confirm it works, and the `github-action.sh` script is available and eXecutable (x flag set on file).

image

juli1 (approved):

PR #443, opened 2024-07-12
Fix line/col search in string

juli1:

When resolving the line/col based on an index, our previous code did not take into account is there was any emoji/grapheme.

## What is your solution?

Use `bstr` and use the grapheme iterator to find the right line/col.

## Testing

Tested against the code we [previously had](https://github.com/DataDog/datadog-static-analyzer/blob/644540d87a6aa39958b7dfbb5d0eaae8e788f919/crates/secrets-core/src/location.rs#L408). Test for the secrets scanning still works.

### DIFF
--- a/crates/common/src/utils/position_utils.rs
+++ b/crates/common/src/utils/position_utils.rs
@@ -4,6 +4,10 @@ use bstr::ByteSlice;
 
 /// Get position of an offset in a code and return a [Position].
 pub fn get_position_in_string(content: &str, offset: usize) -> anyhow::Result<Position> {
+    if offset > content.len() {
+        anyhow::bail!("offset is larger than content length");
+    }
+
     let bstr = BStr::new(&content);
 
     let mut line_number: u32 = 1;
--- a/crates/common/src/utils/position_utils.rs
+++ b/crates/common/src/utils/position_utils.rs
@@ -4,7 +4,7 @@ use bstr::ByteSlice;
 
 /// Get position of an offset in a code and return a [Position].
 pub fn get_position_in_string(content: &str, offset: usize) -> anyhow::Result<Position> {
-    if offset > content.len() {
+    if offset >= content.len() {
         anyhow::bail!("offset is larger than content length");
     }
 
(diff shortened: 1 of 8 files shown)


crates/common/src/utils/position_utils.rs
```diff
@@ -0,0 +1,75 @@
...
+                        line: line_number,
+                        col: col_number,
+                    });
+                }
+
+                // The offset is within the grapheme we are looking for, it's the next col.
+                if offset >= start_index + grapheme.0 && offset < start_index + grapheme.1 {
+                    return Ok(Position {
+                        line: line_number,
+                        col: col_number + 1,
```
amaanq, line 30: It might be nicer to destructure the iterator value, since using `.0` and `.1` isn't clear on what it actually is:

```suggestion
            for (grapheme_start, grapheme_end, _) in line.grapheme_indices() {
                // It's exactly the index we are looking for.
                if offset == start_index + grapheme_start {
                    return Ok(Position {
                        line: line_number,
                        col: col_number,
                    });
                }

                // The offset is within the grapheme we are looking for, it's the next col.
                if offset > start_index + grapheme_start && offset < start_index + grapheme_end {
                    return Ok(Position {
                        line: line_number,
                        col: col_number + 1,
```

Also minor nit, we don't need `>=` and can use `>` to check if the offset is greater, we already checked if it's equal above.

juli1: Thank you for pointing this out, great idea to destructure!
I addressed both feedback!

crates/common/src/utils/position_utils.rs
```diff
@@ -0,0 +1,150 @@
+use crate::model::position::Position;
+use bstr::BStr;
+use bstr::ByteSlice;
+
+/// Get position of an offset in a code and return a [Position].
+pub fn get_position_in_string(content: &str, offset: usize) -> anyhow::Result<Position> {
+    if offset > content.len() {
```
amaanq, line 7: still missing `>=`

```suggestion
    if offset >= content.len() {
```

juli1: No, if we do this, you do not catch the case where your position on the last character and will fail this [test case](https://github.com/DataDog/datadog-static-analyzer/pull/443/files#diff-5f47238d06c82545f9ce76f3c3a41dc0254731fdb43dfe210ae71156f82fe558R145)

juli1: If you want to do this, the location will be at the end of the third line. But I think this level of detail do not really matter that much tbh

PR #451, opened 2024-07-16
[STAL-2508] add php to default rulesets language

amaanq:

PHP rulesets are not being fetched by default

## What is your solution?

Fetch them by default by adding PHP to the `DEFAULT_RULESETS_LANGUAGES` array

Testing:

With no configuration file, we can see the default rulesets of `php-security` and `php-best-practices` running.
image

We can also see the API returning the correct default rulesets here.
image


juli1:
Please add a test section in your PR that shows the full execution of the current analyzer without a file for a PHP codebase. It should shows violations that number of rules loaded.

amaanq:
done

juli1 (approved):

PR #471, opened 2024-07-25
[STAL-2591] Add ability to look up a node's parent from JavaScript

jasonforal:
The JavaScript runtime needs to be able to look up the parent of a CST node.

### Per-tree (parent <> child) relationship caching
Tree-sitter nodes [do not contain a link to their parent](https://github.com/tree-sitter/tree-sitter/blob/800f2c41d0e35e4383172d7a67a16f3933b86039/lib/src/node.c#L23-L27), but they do contain a link to their children ([via the subtree pointer](https://github.com/tree-sitter/tree-sitter/blob/800f2c41d0e35e4383172d7a67a16f3933b86039/lib/src/subtree.h#L237-L238)). Thus, to get the parent of an arbitrary `Node A`, we must traverse down the tree from the root until we reach a node that has `Node A` as a child.

An expected access pattern is for the JavaScript runtime to request a traversal _up_ the tree (i.e. request a node's parent, and then its parent's parent, and so on). For example:
```js
// Find the function containing this `varDeclNode`
let node = varDeclNode;
while (node.type !== "function_declaration") {
    node = ddsa.getParent(node);
}
```

Needing to handle this access pattern performantly necessitates a cache. If we naively traverse from the root down to the parent each time, we end up with `O(D^2)` time complexity (where `D` is the depth of the original node within the tree). Trees can get quite deep, and so this can have a noticeable impact on performance. By trading off space (40 bytes per child -> parent relationship), we bring that time complexity down to `O(D)`.

The cache logic is [extensively tested](https://github.com/DataDog/datadog-static-analyzer/blob/ee1e10e1c7b31834f907fc70352074f07ad6d0d6/crates/static-analysis-kernel/src/analysis/ddsa_lib/context/root.rs#L237-L384).

### `OpSafeRawTSNode`: Safety
We need another way of creating a [RawTSNode](https://github.com/DataDog/datadog-static-analyzer/blob/6d8a8d05e7f72d93ea8496fc7e1ae9c70a6f8530/crates/static-analysis-kernel/src/analysis/ddsa_lib.rs#L21) so that it can be accessed in `tree_sitter::Node` form via a deno op. We [previously](https://github.com/DataDog/datadog-static-analyzer/blob/ee1e10e1c7b31834f907fc70352074f07ad6d0d6/crates/static-analysis-kernel/src/analysis/ddsa_lib/ops.rs#L156-L159) could only do this by reading a `RawTSNode` from the `TsNodeBridge` (which is guaranteed safe). However, that isn't applicable here because we _don't_ (and don't want to) preemptively serialize the nodes to v8 (i.e. they are not already contained in the bridge). We address this by implementing the [cache on ddsa_lib::RootContext](https://github.com/DataDog/datadog-static-analyzer/blob/ee1e10e1c7b31834f907fc70352074f07ad6d0d6/crates/static-analysis-kernel/src/analysis/ddsa_lib/context/root.rs#L23-L34), and then making a [small addition](https://github.com/DataDog/datadog-static-analyzer/blob/6d8a8d05e7f72d93ea8496fc7e1ae9c70a6f8530/crates/static-analysis-kernel/src/analysis/ddsa_lib/ops.rs#L208-L214) to `OpSafeRawTSNode` to uphold the safety guarantee. Added documentation on safety is [here](https://github.com/DataDog/datadog-static-analyzer/blob/6d8a8d05e7f72d93ea8496fc7e1ae9c70a6f8530/crates/static-analysis-kernel/src/analysis/ddsa_lib/ops.rs#L201-L203) and [here](https://github.com/DataDog/datadog-static-analyzer/blob/6d8a8d05e7f72d93ea8496fc7e1ae9c70a6f8530/crates/static-analysis-kernel/src/analysis/ddsa_lib/context/root.rs#L103-L107).

### ddsa JavaScript API
The last thing this PR implements is the actual deno op and JS bindings. A rule can now call:
```js
const parent = ddsa.getParent(node);
```

This functionality will be used in future additions to the JavaScript runtime.

### DIFF
--- a/crates/static-analysis-kernel/src/analysis/ddsa_lib/extension.rs
+++ b/crates/static-analysis-kernel/src/analysis/ddsa_lib/extension.rs
@@ -11,6 +11,7 @@ deno_core::extension!(
         ops::op_console_push,
         ops::op_current_ts_tree_text,
         ops::op_ts_node_named_children,
+        ops::op_ts_node_parent,
         ops::op_ts_node_text,
         #[cfg(test)]
         ops::cfg_test_op_rust_option,
--- a/crates/static-analysis-kernel/src/analysis/ddsa_lib/js/ddsa.js
+++ b/crates/static-analysis-kernel/src/analysis/ddsa_lib/js/ddsa.js
@@ -5,7 +5,7 @@
 import { SEALED_EMPTY_ARRAY } from "ext:ddsa_lib/utility";
 import { TreeSitterFieldChildNode } from "ext:ddsa_lib/ts_node";
 
-const { op_ts_node_named_children } = Deno.core.ops;
+const { op_ts_node_named_children, op_ts_node_parent } = Deno.core.ops;
 
 /**
  * The main entrypoint to the ddsa JavaScript runtime's API.
@@ -38,5 +38,19 @@ export class DDSA {
             }
         }
         return children;
+    }
+
+    /**
+     * Fetches and returns the provided node's parent in the tree-sitter tree.
+     * If the node is the root node of the tree, `undefined` will be returned.
+     * @param {TreeSitterNode | TreeSitterFieldChildNode} node
+     * @returns {TreeSitterNode | undefined}
+     */
+    getParent(node) {
+        const parentId = op_ts_node_parent(node.id);
+        if (parentId === null) {
+            return undefined;
+        }
+        return globalThis.__RUST_BRIDGE__ts_node.get(parentId);
     }
 }
--- a/crates/static-analysis-kernel/src/analysis/ddsa_lib/js/ddsa.rs
+++ b/crates/static-analysis-kernel/src/analysis/ddsa_lib/js/ddsa.rs
@@ -20,6 +20,7 @@ mod tests {
         let expected = &[
             // Methods
             "getChildren",
+            "getParent",
         ];
         assert!(js_instance_eq(CLASS_NAME, expected));
         let expected = &[];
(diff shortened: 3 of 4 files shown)


crates/static-analysis-kernel/src/analysis/ddsa_lib/context/root.rs
```diff
@@ -16,6 +19,22 @@ pub struct RootContext {
...
+    /// traverse down the tree from the root until we reach a node that has `Node A` as a child.
+    ///
+    /// An expected access pattern is for the JavaScript runtime to request a traversal _up_ the tree
+    /// (i.e. request a node's parent, and then its parent's parent, and so on). Without a cache,
+    /// this has `O(D^2)` time complexity (`D` being the depth of the original node within the tree).
+    /// This cache reduces that time complexity to `O(D)` (costing `O(D)` space complexity: 8 + 32 == 40 bytes per relationship).
+    parent_map: RefCell<HashMap<usize, RawTSNode>>,
+    /// A counter for how many times a tree traversal was triggered in `get_ts_node_parent`.
+    #[cfg(test)]
+    _cfg_test_tree_traversal_count: std::cell::Cell<usize>,
```
juli1, line 37: Why do we even need this?

crates/static-analysis-kernel/src/analysis/ddsa_lib/context/root.rs
```diff
@@ -51,4 +71,315 @@ impl RootContext {
...
+        }
+
+        ///// Test hook
+        #[cfg(test)]
+        {
+            // Increment the counter, as we're going to traverse the tree.
+            let current = &self._cfg_test_tree_traversal_count.get();
+            self._cfg_test_tree_traversal_count.set(current + 1);
+        }
+        /////
```
juli1, line 118: Do we really need this?

juli1 (approved):
The code is fine as is. My worry is that we start to take a *very* complicated architecture with caches all over the place, which makes the code more complex to understand and maintain, for an optimization cost that is not certain. And now, there is only 1.5 engineers who understand how it works.

We have a cache for the parent/child with this PR. With Datafow, we are starting to also take the direction to have a cache with the data being calculated on demand/lazily. There is also the content cache

If we continue that route, we will end up with dozens of cache calculated lazily, which will make the maintenance of the analyzer very complex and not maintainable. Instead, we should simplify and build this data at analysis time. It will have a small performance hit but it will simplify the maintenance.

For example, browsing the tree before analyzing a file has a small cost, it could fill a global cache for both data-flow, the parent/child and any other data we may need. We then cache for data for each TsNode (similar to this one). Once cache for everything and any new operation is just about parsing and adding more to this cache.

PR #478, opened 2024-08-02
Add results when running a Git Hook

juli1:

For statistics purposes, we want to export the results of the Git hook.

## What is your solution?

Export all the results in a file with the `--output` parameter.

### DIFF
--- a/crates/bins/src/bin/datadog-static-analyzer-git-hook.rs
+++ b/crates/bins/src/bin/datadog-static-analyzer-git-hook.rs
@@ -560,8 +560,8 @@ fn main() -> Result<()> {
     if let Some(output_file) = output_opt {
         let sarif_content = generate_sarif_file(
             &configuration,
-            &all_rule_results,
-            &secrets_results,
+            all_rule_results,
+            secrets_results,
             SarifReportMetadata {
                 add_git_info: false,
                 debug: configuration.use_debug,
--- a/crates/bins/src/bin/datadog-static-analyzer-git-hook.rs
+++ b/crates/bins/src/bin/datadog-static-analyzer-git-hook.rs
@@ -12,8 +12,10 @@ use cli::model::cli_configuration::CliConfiguration;
 use cli::rule_utils::{
     check_rules_checksum, convert_rules_to_rules_internal, get_languages_for_rules,
 };
+use cli::sarif::sarif_utils::{generate_sarif_file, SarifReportMetadata};
 use cli::utils::{choose_cpu_count, get_num_threads_to_use, print_configuration};
 use common::analysis_options::AnalysisOptions;
+use common::model::diff_aware::DiffAware;
 use getopts::Options;
 use git2::Repository;
 use itertools::Itertools;
@@ -29,9 +31,11 @@ use secrets::model::secret_result::SecretResult;
 use secrets::scanner::{build_sds_scanner, find_secrets};
 use secrets::secret_files::should_ignore_file_for_secret;
 use std::collections::HashMap;
+use std::io::Write;
 use std::path::PathBuf;
 use std::process::exit;
 use std::sync::Arc;
+use std::time::Instant;
 use std::{env, fs, io};
 use terminal_emoji::Emoji;
 
@@ -124,6 +128,13 @@ fn main() -> Result<()> {
         "484df034272a3f63c1796ef44f83a790d7729590",
     );
 
+    opts.optopt(
+        "",
+        "output",
+        "file that contains all findings from the Git hooks (SARIF output)",
+        "/tmp/file-output.sarif",
+    );
+
     opts.optopt("d", "debug", "use debug mode", "yes/no");
 
     opts.optopt(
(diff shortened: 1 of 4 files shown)


crates/bins/src/bin/datadog-static-analyzer-git-hook.rs
```diff
@@ -124,6 +128,13 @@ fn main() -> Result<()> {
         "484df034272a3f63c1796ef44f83a790d7729590",
     );
 
+    opts.optopt(
+        "",
```
amaanq, line 132: why not allow -o for short?

```suggestion
        "o",
```

crates/bins/src/bin/datadog-static-analyzer-git-hook.rs
```diff
@@ -124,6 +128,13 @@ fn main() -> Result<()> {
...
+        "output",
+        "file that contains all findings from the Git hooks (SARIF output)",
```
amaanq, line 134: it's nice that we're describing the file, but it doesn't explain what the relationship between the file and the cli arg is (e.g. in the main cli we say "output file name", which is clearer)

```suggestion
        "output file name to write all findings from the Git hooks to",
```

crates/bins/src/bin/datadog-static-analyzer-git-hook.rs
```diff
@@ -448,31 +462,47 @@ fn main() -> Result<()> {
...
+                        lines.contains(&v.start.line)
+                    } else {
+                        false
+                    }
+                })
+                .collect();
+            new_result.violations = filtered_violations;
+            new_result
+        })
+        .collect();
```
amaanq, line 485: I don't think it makes sense to consume an iterator and then clone each element inside `map`, we should mutate the vec in place instead. You can do this with `iter_mut` + `for_each` or just do `for rr in &mut all_rule_results`, and then in the body you modify the `violations` field with `retain` to retain only the elements that satisfy the clojure. Or you could also perform the filtering logic *before* you extend `all_rule_results`. My suggestion uses `iter_mut` + `for_each` with `retain`, but I think it'd be best to do this before you extend `all_rule_results` above.

```suggestion
    all_rule_results.iter_mut().for_each(|rr| {
        let path = PathBuf::from(&rr.filename);
        rr.violations.retain(|v| {
            if let Some(lines) = modifications.get(&path) {
                lines.contains(&v.start.line)
            } else {
                false
            }
        });
    });
```

crates/bins/src/bin/datadog-static-analyzer-git-hook.rs
```diff
@@ -508,30 +538,69 @@ fn main() -> Result<()> {
...
+                            lines.contains(&v.start.line)
+                        } else {
+                            false
+                        }
+                    })
+                    .collect();
+                new_result.matches = filtered_violations;
+                new_result
+            })
+            .collect();
```
amaanq, line 561: Same story as above, though in this case you'll need to make `secrets_results` `mut`

```rs
        secrets_results.iter_mut().for_each(|rr| {
            rr.matches.retain(|v| {
                let path = PathBuf::from(&rr.filename);
                if let Some(lines) = modifications.get(&path) {
                    lines.contains(&v.start.line)
                } else {
                    false
                }
            });
        });
```

crates/cli/src/sarif/sarif_utils.rs
```diff
@@ -694,6 +694,53 @@ pub fn generate_sarif_report(
...
+        .map(SarifRuleResult::try_from)
+        .collect::<Result<Vec<_>, _>>()
+        .map_err(anyhow::Error::msg)?;
+
+    let secret_results = secrets_rule_results
+        .iter()
+        .cloned()
+        .map(SarifRuleResult::try_from)
+        .collect::<Result<Vec<_>, _>>()
+        .map_err(anyhow::Error::msg)?;
```
amaanq, line 729: I'm not really fond of all these `cloned` invocations, especially since this is at the end of both callsites typically, so it doesn't make much sense to not consume it here imo. You will have to change the call sites to consume these vectors instead. 

```suggestion
pub fn generate_sarif_file(
    configuration: &CliConfiguration,
    static_analysis_rule_results: Vec<RuleResult>,
    secrets_rule_results: Vec<SecretResult>,
    sarif_report_metadata: SarifReportMetadata,
) -> Result<String> {
    let static_rules_sarif: Vec<SarifRule> = configuration
        .rules
        .iter()
        .cloned()
        .map(|r| r.into())
        .collect();

    let secrets_rules_sarif: Vec<SarifRule> = configuration
        .secrets_rules
        .clone()
        .into_iter()
        .map(|r| r.into())
        .collect();

    let static_analysis_results = static_analysis_rule_results
        .into_iter()
        .map(SarifRuleResult::try_from)
        .collect::<Result<Vec<_>, _>>()
        .map_err(anyhow::Error::msg)?;

    let secret_results = secrets_rule_results
        .into_iter()
        .map(SarifRuleResult::try_from)
        .collect::<Result<Vec<_>, _>>()
        .map_err(anyhow::Error::msg)?;
```

amaanq (approved):

PR #485, opened 2024-08-09
Various QOL improvements for unit testing

jasonforal:
We've had a number of unit test pain points (writing and reading) accumulate over time. This PR implements a grab bag of small fixes.

## What is your solution?
Highlights:
* Implement a test-only deno extension, which will allow us to build a reusable library of JavaScript helper functions that can be used across tests.
* Add support for a `console.log`-based debugging workflow.
* Add a `TsTree` helper struct to make locating specific tree-sitter nodes easier in unit tests.

Additionally, this PR exposes JavaScript error stack traces, which makes it easier to debug issues. Previously, when we had an error, all we would get is:
```
TypeError: Cannot read properties of undefined (reading 'someProp')
```

Now, we can get exact module, function, line, and column:
```
TypeError: Cannot read properties of undefined (reading 'someProp')
  at SomeClass.someOtherFunction (ext:ddsa_lib/someModule:572:29)
  at SomeClass.someFunction (ext:ddsa_lib/someModule:157:22)
  at <anonymous>:1:13
```

### DIFF
--- a/crates/static-analysis-kernel/src/analysis/ddsa_lib/common.rs
+++ b/crates/static-analysis-kernel/src/analysis/ddsa_lib/common.rs
@@ -4,7 +4,7 @@
 
 use deno_core::v8;
 use deno_core::v8::HandleScope;
-use std::fmt::Debug;
+use std::fmt::{Debug, Display, Formatter};
 use std::ops::Deref;
 use std::time::Duration;
 
@@ -13,8 +13,8 @@ pub type NodeId = u32;
 
 #[derive(Debug, thiserror::Error)]
 pub enum DDSAJsRuntimeError {
-    #[error("execution error: {reason}")]
-    Execution { reason: String },
+    #[error("{error}")]
+    Execution { error: JsError },
     #[error("JavaScript execution timeout")]
     JavaScriptTimeout { timeout: Duration },
     #[error("expected `{name}` to exist within the v8 context")]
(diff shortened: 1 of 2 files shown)


crates/static-analysis-kernel/src/analysis/ddsa_lib/common.rs
```diff
@@ -13,8 +13,10 @@ pub type NodeId = u32;
 
 #[derive(Debug, thiserror::Error)]
 pub enum DDSAJsRuntimeError {
-    #[error("execution error: {reason}")]
-    Execution { reason: String },
+    #[error("{error}")]
+    Execution {
+        error: Box<deno_core::error::JsError>,
```
juli1, line 18: What is the motivation for doing this? Should we couple the `ddsa` with `deno`? (or rather, should we expose deno-specific stuff in the ddsa layer?

crates/static-analysis-kernel/src/analysis/ddsa_lib/bridge/ts_node.rs
```diff
@@ -153,57 +153,17 @@ mod tests {
...
-            let lang = get_tree_sitter_language(&language);
-            let mut parser = tree_sitter::Parser::new();
-            parser.set_language(&lang).unwrap();
-            Self(parser)
-        }
-        fn parse(&mut self, text: &'static str) -> Tree {
-            Tree(Arc::new(self.0.parse(text, None).unwrap()), text)
-        }
-    }
-
```
juli1, line 206: What is the reason for removing this from the core and putting only in the tests?

crates/static-analysis-kernel/src/analysis/ddsa_lib/test_utils.rs
```diff
@@ -171,26 +171,54 @@ fn cfg_test_deno_ext() -> deno_core::Extension {
...
-        // Create a unique (arbitrary) variable name for the import.
-        let var_name = "a".repeat(idx + 1);
-        entrypoint_code += &format!(
-            r#"
+        if efs.specifier == "ext:ddsa_lib_cfg_test/helpers" {
+            // Only import under the "TEST" namespace.
+            entrypoint_code += &format!(
+                r#"
+import * as TEST from "{}";
+globalThis.TEST = TEST;
```
juli1, line 200: Why putting this under `TEST`? Why not adding this either with

 - the global scope (`assert(something)`)?
 - the `ddsa` scope `ddsa.assert()`)?

juli1 (commented):
A few questions to understand the approach. Given that `assert()` is so common, it would make sense to have it define either in the global namespace or the `ddsa` namespace.

juli1 (approved):

PR #486, opened 2024-08-09
[STAL-2635] Add JavaScript `Digraph`

jasonforal:
To support more sophisticated analysis techniques, we want to be able to construct arbitrary graphs from the source code we're analyzing.

We're initially iterating on the construction and traversal logic in JavaScript for iteration speed. These can be ported up to Rust if/when it makes sense.

This PR has two main goals:
1. Implement the graph data structure.
2. Establish a pattern to write unit tests against an expected graph structure.

### JavaScript-based Digraph
We'll initially be implementing taint analysis, and so our initial graph will be stored as an adjacency list containing directed edges that indicate the flow of tainted values from sources to targets. We need to be able to type the edges to distinguish between various semantics (for now: "assignment" and "dependence")

Thus, in pseudo-code, an edge looks like

```rs
type VertexId = u16;
#[repr(u8)]
enum EdgeKind {
    Assignment,
    Dependence,
}
#[derive(Copy)]
struct Edge {
    target: VertexId,
    kind: EdgeKind,
}
type Adjacencies = Vec<EdgeKind>;
```

Defining and parsing edges as structs makes sense in Rust, as LLVM can optimize these into registers.

The straightforward JavaScript equivalent of this would be an object:
```js
const edge1 = {
    target: 123,
    kind: EdgeKind_DEPENDENCE,
};
const adjacencies = [edge1, /* edge2, edge3, ... */];
```

However, this isn't ideal because:
* Equality checking (e.g. to traverse and/or detect cycles) requires additional complexity (`number` equality is simply by-value, whereas [`object` equality is by-reference](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Equality#comparison_of_objects))
* Objects always perform a heap allocation for each edge, as v8 cannot optimize in a similar manner, and we expect to create a lot of edges.

Instead, we use [bit packing](https://github.com/DataDog/datadog-static-analyzer/blob/1e9bcb9997f25012c9dedede3c52b666667c9f79/crates/static-analysis-kernel/src/analysis/ddsa_lib/js/flow/graph.rs#L112-L120) to represent the edge as a number (specifically: integer).

I believe this ends up being a more simple and more maintainable solution in the long-run, particularly because we will eventually be introducing vertex types, so were we to use an object here, we'd have to write something like:
```js
const edge1 = {
    target: {
        id: 123,
        kind: VertexKind_CST,
    },
    kind: EdgeKind_DEPENDENCE,
};
```
We'd then need a way to represent this as an adjacency list. We can't easily do `Map<Vertex, Edge[]>` because objects are by-reference, not by-value, so we would need to come up with some one-off serialization of `Vertex` into a number/string map key. At that point, it's more simple just to have everything use the same serialization to number.

### Unit Tests
We'll be building more complex graphs that are specific to the semantics of each programming language. We need to be able to easily read (and write) unit tests that make assertions about the structure of a graph generated from source code. To that end, this PR introduces the following pattern:

* Expected graphs are specified using [DOT Language](https://graphviz.org/doc/info/lang.html).
* Unit tests specify a transformer function to map a manually-written graph to an algorithmically generated one.

To illustrate: while not implemented in _this_ PR, a future PR will use this pattern to implement a DSL that can specify CST nodes as graph nodes and edges as taint flow. For example, for the given Java code:
```java
String value = request.getHeader("abc");
System.out.println(value);
```
A test writer will be able to specify that they expect at line 1, "`value` is dependent on `request`, and that the `println` function is a sink for that value by simply writing:

```dot
digraph expected {
    value1 [text=value,line=1]
    value2 [text=value,line=2]
    request [line=1]
    println [line=2]

    value1 -> request [kind=dependence]  
    value2 -> value1 [kind=dependence]
    println -> value2 [kind=dependence]
}
```

* New third party dependency `graphviz-rust` is used to parse (and eventually generate) DOT language graphs used in unit tests.

### DIFF
--- a/crates/static-analysis-kernel/src/analysis/ddsa_lib/js/flow/graph.rs
+++ b/crates/static-analysis-kernel/src/analysis/ddsa_lib/js/flow/graph.rs
@@ -1,0 +1,602 @@
+// Unless explicitly stated otherwise all files in this repository are licensed under the Apache License, Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2024 Datadog, Inc.
+
+use deno_core::v8;
+use graphviz_rust::dot_structures;
+use std::borrow::{Borrow, Cow};
+use std::collections::{HashMap, HashSet};
+
+/// A wrapper around a [`graphviz_rust::Graph::DiGraph`]. This is only used in unit tests to specify
+/// and assert on the structure of a graph.
+#[derive(Debug, Clone)]
+pub struct Digraph(dot_structures::Graph);
+
+impl Digraph {
+    /// Constructs a [`Self`] from the provided `input`.
+    /// # Panics
+    /// * Panics if `input` is not a strict digraph.
+    /// * Panics if any [`dot_structures::Stmt`] in `input` isn't a node or an edge.
+    pub fn new(input: dot_structures::Graph) -> Self {
+        use dot_structures::*;
+        let Graph::DiGraph { strict, stmts, .. } = &input else {
+            panic!("graph should be a digraph");
+        };
+        assert!(strict, "should be \"strict\"");
+        let all_valid_stmts = stmts
+            .iter()
+            .all(|stmt| matches!(stmt, Stmt::Node(..) | Stmt::Edge(..)));
+        assert!(all_valid_stmts, "should only contain `node`s and `edge`s");
+        Self(input)
+    }
+
+    /// Returns `true` if this graph is a subgraph of the provided `other`. A graph is considered
+    /// a subgraph if the `other` has identical nodes and edges as this graph.
+    pub fn is_subgraph_of(&self, other: impl Borrow<Self>) -> bool {
+        self.difference(other).is_empty()
+    }
+
+    /// Returns the nodes and edges that are in `self`, but not in `other`.
+    pub fn difference(&self, other: impl Borrow<Self>) -> Vec<dot_structures::Stmt> {
+        use dot_structures::*;
+
+        /// Returns a hash map of nodes and edges, using an arbitrary (but stable) key.
+        fn normalize_stmts(graph: &Graph) -> HashMap<String, &Stmt> {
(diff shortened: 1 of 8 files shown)


crates/static-analysis-kernel/src/analysis/ddsa_lib/js/flow/graph.rs
```diff
@@ -0,0 +1,799 @@
...
+
+impl FlowDigraph {
+    /// Returns `true` if this graph is a subgraph of the provided `other`. A graph is considered
+    /// a subgraph if the `other` has identical nodes and edges as this graph.
+    pub fn is_subgraph_of(&self, other: &Self) -> bool {
+        self.difference(other).is_empty()
+    }
+
+    /// Returns the nodes and edges that are in `self`, but not in `other`.
+    pub fn difference(&self, other: &Self) -> Vec<dot_structures::Stmt> {
```
juli1, line 28: Can we explain in the comment why we need this?

crates/static-analysis-kernel/src/analysis/ddsa_lib/js/flow/graph.rs
```diff
@@ -0,0 +1,799 @@
...
+    pub fn target(&self) -> ddsa_lib::common::NodeId {
+        self.0 >> Self::KIND_BITS
+    }
+
+    /// The type of edge this is.
+    pub fn kind(&self) -> EdgeKind {
+        EdgeKind::try_from_id((self.0 & Self::KIND_BIT_MASK) as usize)
+            .expect("js should serialize PackedEdge correctly")
+    }
+}
```
juli1, line 138: Why this? This adds more complexity than anything else.

Why not a struct with a nodeid and a kind?

juli1 (commented):
I really like the graphviz approach for testing. We should also be able to export the graphviz dot representation into a file for debugging purposes.

I am not sure about the `PackedEdge`. It seems early optimization to me and not needed.

juli1 (approved):

PR #489, opened 2024-08-13
[STAL-2687] Fetch remote config

juli1:

We want to be able to fetch the configuration remotely.

## Solution

We attempt to read the configuration file

1. If there is a configuration file, we attempt to get the remote configuration by passing the existing config file and the backend returns the final config
2. If there is no config, we retrieve the remote configuration by passing a `null` configuration to the API

## Notes

1. We fetch the remote configuration ONLY if we have some API keys defined. Otherwise, we do not even try
2. We added a lot of debug information so that customer will be able to troubleshoot

## Testing

Tested locally, existing tests should pass

```
cargo run --bin datadog-static-analyzer -- --staging --directory /Users/julien.delange/git/datadog-ci --format sarif --output ~/plop.json  --debug yes 
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.16s
     Running `target/debug/datadog-static-analyzer --staging --directory /Users/julien.delange/git/datadog-ci --format sarif --output /Users/julien.delange/plop.json --debug yes`
Remote config (base64): "cnVsZXNldHM6CiAgLSBvcmctcnVsZXNldAogIC0ga2Rza2xmanNsZGYK"
Error: ruleset org-ruleset not found
Error: error when reading rules from API

Caused by:
    Error: ruleset org-ruleset not found
```

### DIFF
--- a/crates/bins/src/bin/datadog-static-analyzer.rs
+++ b/crates/bins/src/bin/datadog-static-analyzer.rs
@@ -233,7 +233,7 @@ fn main() -> Result<()> {
         };
 
     if configuration_file.is_none() && use_debug {
-        eprintln!("INFO: no configuration detected (either local or remote)")
+        eprintln!("INFO: no configuration detected locally or remotely")
     }
 
     let rule_config_provider = configuration_file
--- a/crates/bins/src/bin/datadog-static-analyzer.rs
+++ b/crates/bins/src/bin/datadog-static-analyzer.rs
@@ -11,7 +11,7 @@ use std::sync::Arc;
 use std::time::{Duration, Instant, SystemTime};
 use std::{env, fs};
 
-use cli::config_file::read_config_file;
+use cli::config_file::get_config;
 use cli::constants::{DEFAULT_MAX_CPUS, DEFAULT_MAX_FILE_SIZE_KB};
 use cli::csv;
 use cli::datadog_utils::{
@@ -221,7 +221,7 @@ fn main() -> Result<()> {
     }
 
     let configuration_file: Option<ConfigFile> =
-        match read_config_file(directory_to_analyze.as_str()) {
+        match get_config(directory_to_analyze.as_str(), use_debug) {
             Ok(cfg) => cfg,
             Err(err) => {
                 eprintln!(
@@ -231,6 +231,11 @@ fn main() -> Result<()> {
                 exit(1)
             }
         };
+
+    if configuration_file.is_none() && use_debug {
+        eprintln!("INFO: no configuration detected (either local or remote)")
+    }
+
     let rule_config_provider = configuration_file
         .as_ref()
         .map(RuleConfigProvider::from_config)
--- a/crates/cli/src/config_file.rs
+++ b/crates/cli/src/config_file.rs
@@ -50,7 +50,7 @@ pub fn read_config_file(path: &str) -> Result<Option<ConfigFile>> {
                 if size_read == 0 {
                     return Err(anyhow!("the config file is empty"));
                 }
-                parse_config_file(&contents).map(|c| Some(c))
+                parse_config_file(&contents).map(Some)
             } else {
                 Ok(None)
             }
--- a/crates/cli/src/config_file.rs
+++ b/crates/cli/src/config_file.rs
@@ -39,45 +39,35 @@ fn get_config_file(path: &str) -> Result<Option<File>> {
 // If the file does not exist, we return a Ok(None).
 // If there is an error reading the file, we return a failure
 pub fn read_config_file(path: &str) -> Result<Option<ConfigFile>> {
-    match get_config_file(path) {
-        Ok(file_opt) => {
-            if let Some(mut file) = file_opt {
-                let mut contents = String::new();
+    if let Some(mut file) = get_config_file(path)? {
+        let mut contents = String::new();
 
-                let size_read = file
-                    .read_to_string(&mut contents)
-                    .context("error when reading the configration file")?;
-                if size_read == 0 {
-                    return Err(anyhow!("the config file is empty"));
-                }
-                parse_config_file(&contents).map(Some)
-            } else {
-                Ok(None)
-            }
+        let size_read = file
+            .read_to_string(&mut contents)
+            .context("error when reading the configration file")?;
+        if size_read == 0 {
+            return Err(anyhow!("the config file is empty"));
         }
-        Err(e) => Err(e),
+        parse_config_file(&contents).map(Some)
+    } else {
+        Ok(None)
     }
 }
 
 // Read the config file in base64
 pub fn read_config_file_in_base64(path: &str) -> Result<Option<String>> {
-    match get_config_file(path) {
-        Ok(file_opt) => {
-            if let Some(mut file) = file_opt {
-                let mut contents = String::new();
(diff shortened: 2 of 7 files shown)


crates/cli/src/config_file.rs
```diff
@@ -22,26 +20,117 @@ pub fn read_config_file(path: &str) -> Result<Option<ConfigFile>> {
...
+                    return Err(anyhow!("the config file is empty"));
+                }
+                match parse_config_file(&contents) {
+                    Ok(c) => Ok(Some(c)),
+                    Err(e) => Err(e),
+                }
+            }
+            None => Ok(None),
+        },
+        Err(e) => Err(e),
```
amaanq, line 60: the Err case in the outer match just returns the error, so why not just use `?` to propagate the error if any, and use a `let ... Some` to bail early when it's None as well: 
```rs
let Some(mut file) = get_config_file(path)? else {
    return Ok(None);
};
```

crates/cli/src/config_file.rs
```diff
@@ -22,26 +20,117 @@ pub fn read_config_file(path: &str) -> Result<Option<ConfigFile>> {
...
+
+                let size_read = file
+                    .read_to_string(&mut contents)
+                    .context("error when reading the configration file")?;
+                if size_read == 0 {
+                    return Err(anyhow!("the config file is empty"));
+                }
+                match parse_config_file(&contents) {
+                    Ok(c) => Ok(Some(c)),
+                    Err(e) => Err(e),
```
amaanq, line 55:
```suggestion
                parse_config_file(&contents).map(|c| Some(c))
```

crates/cli/src/config_file.rs
```diff
@@ -22,26 +20,117 @@ pub fn read_config_file(path: &str) -> Result<Option<ConfigFile>> {
...
+                    .context("error when reading the configration file")?;
+                if size_read == 0 {
+                    return Err(anyhow!("the config file is empty"));
+                }
+                Ok(Some(encode_base64_string(contents)))
+            }
+            None => Ok(None),
+        },
+        Err(e) => Err(e),
+    }
```
amaanq, line 82: Same as above, the double match can be reduced with a `let Some(...) else {}` statement

crates/cli/src/config_file.rs
```diff
@@ -22,26 +20,117 @@ pub fn read_config_file(path: &str) -> Result<Option<ConfigFile>> {
...
+                            eprintln!("Error when parsing remote config: {:?}", e);
+                            eprintln!("Proceeding with local config");
+                            cf
+                        }
+                    }
+                }
+                Err(e) => {
+                    if debug {
+                        eprintln!("error when trying to get remote config: {:?}", e);
+                        eprintln!("will use the local configuration if any")
```
amaanq, line 124:
```suggestion
                        eprintln!("Falling back to the local configuration, if any")
```

crates/cli/src/config_file.rs
```diff
@@ -22,26 +20,117 @@ pub fn read_config_file(path: &str) -> Result<Option<ConfigFile>> {
...
+                        Err(e) => {
+                            eprintln!("Error when parsing remote config: {:?}", e);
+                            eprintln!("Proceeding with local config");
+                            cf
+                        }
+                    }
+                }
+                Err(e) => {
+                    if debug {
+                        eprintln!("error when trying to get remote config: {:?}", e);
```
amaanq, line 123:
```suggestion
                        eprintln!("Error when attempting to fetch the remote config: {:?}", e);
```

crates/cli/src/datadog_utils.rs
```diff
@@ -343,6 +349,44 @@ pub fn get_diff_aware_information(
...
+    let path = "config/client";
+    let req = make_request(RequestMethod::Post, path, false, true)?.json(&request_payload);
+    let server_response = perform_request(req, path, debug)?;
+
+    let status_code = server_response.status();
+    let response_text = &server_response.text()?;
+
+    let api_response = serde_json::from_str::<ConfigResponse>(response_text);
+
+    if !&status_code.is_success() {
```
amaanq, line 377:
```suggestion
    if !status_code.is_success() {
```

crates/cli/src/datadog_utils.rs
```diff
@@ -343,6 +349,44 @@ pub fn get_diff_aware_information(
...
+        return Err(anyhow!("server returned error {}", &status_code.as_u16()));
```
amaanq, line 378:
```suggestion
        return Err(anyhow!("server returned error {}", status_code.as_u16()));
```

crates/cli/src/git_utils.rs
```diff
@@ -103,6 +103,18 @@ fn get_changed_files(diff: &Diff) -> anyhow::Result<HashMap<PathBuf, Vec<u32>>>
     Ok(res)
 }
 
+pub fn get_repository_url(path: &str) -> anyhow::Result<String> {
+    let repository_opt = Repository::init(path);
+    match repository_opt {
+        Ok(repository) => Ok(repository
+            .find_remote("origin")?
```
amaanq, line 110: I'm a little wary of hardcoding origin here, it'd be nice if we could perform some logic to see if there's only one remote & use that, otherwise use origin, and *then* error out

crates/cli/src/model/datadog_api.rs
```diff
@@ -26,6 +27,26 @@ pub struct DiffAwareResponse {
...
 
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct ConfigResponseDataAttributes {
+    #[serde(rename = "config_base64")]
+    pub config_base64: String,
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct ConfigResponseData {
+    #[serde(rename = "id")]
```
amaanq, line 38: Not asking you to remove, but as an FYI renames are not needed for field names that are the exact same as the json.

crates/bins/src/bin/datadog-static-analyzer.rs
```diff
@@ -231,6 +231,11 @@ fn main() -> Result<()> {
                 exit(1)
             }
         };
+
+    if configuration_file.is_none() && use_debug {
+        eprintln!("INFO: no configuration detected (either local or remote)")
```
amaanq, line 236:
```suggestion
        eprintln!("INFO: no configuration detected locally or remotely")
```

crates/cli/src/git_utils.rs
juli1: This is already how we do it for diff-aware, it's just taking the same approach.

crates/cli/src/config_file.rs
```diff
@@ -22,26 +20,106 @@ pub fn read_config_file(path: &str) -> Result<Option<ConfigFile>> {
...
+                debug,
+            );
+
+            match remote_config {
+                Ok(rc) => {
+                    if debug {
+                        eprintln!("Remote config (base64): {:?}", rc);
+                    }
+
+                    let remote_config_base64 =
```
amaanq, line 99: I think you can swap the names of remote_config and remote_config_base64, right now remote_config is the base64 encoded version, and remote_config_base64 is normal text which is weird

amaanq (approved):

PR #490, opened 2024-08-14
Add tests for JS/TS

juli1:

Making our test suite stronger

## What is your solution?

Checking we catch violations at each commit/push

### DIFF
--- a/misc/integration-test-js-ts.sh
+++ b/misc/integration-test-js-ts.sh
@@ -26,6 +26,13 @@ if [ $? -ne 0 ]; then
   exit 1
 fi
 
+FINDINGS=`jq '.runs[0].results|length' ${REPO_DIR}/results.json`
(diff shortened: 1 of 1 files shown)


amaanq (approved):

misc/integration-test-js-ts.sh
```diff
@@ -26,6 +26,13 @@ if [ $? -ne 0 ]; then
   exit 1
 fi
 
+FINDINGS=`jq '.runs[0].results|length' ${REPO_DIR}/results.json`
+echo "Found $FINDINGS violations"
+if [ $FINDINGS -lt 10 ]; then
+  echo "only $FINDINGS found, expecting at least 10 findings"
```
amaanq, line 32: Keep tense consistent

```suggestion
  echo "only $FINDINGS found, expected at least 10 findings"
```

PR #499, opened 2024-08-26
Do not follow symlinks while scanning the repo for files to analyze

jacobotb:
We want to be able to tell, just by inspecting the code, that the scanner is not following symlinks.

## What is your solution?
Set the configuration values, even if they are the defaults, to make it explicit.

### DIFF
--- a/crates/cli/src/file_utils.rs
+++ b/crates/cli/src/file_utils.rs
@@ -128,7 +128,10 @@ pub fn get_files(
     };
 
     for directory_to_walk in directories_to_walk {
-        for entry in WalkDir::new(directory_to_walk.as_str()) {
+        for entry in WalkDir::new(directory_to_walk.as_str())
+            .follow_links(false)
+            .follow_root_links(true)
+        {
             let dir_entry = entry?;
             let entry = dir_entry.path();
 


crates/cli/src/file_utils.rs
```diff
@@ -128,7 +128,10 @@ pub fn get_files(
     };
 
     for directory_to_walk in directories_to_walk {
-        for entry in WalkDir::new(directory_to_walk.as_str()) {
+        for entry in WalkDir::new(directory_to_walk.as_str())
+            .follow_links(false)
+            .follow_root_links(true)
```
juli1, line 133: Why `true` on this one? Could we just have it `false` as well or it's breaking some behavior?

jacobotb: I don't know for sure whether anyone is doing it, but it is plausible that someone could check out the code to a temporary location, then set a soft link to that location, then execute the static analyzer on that soft link.

The dangerous case is when there is a soft link inside the repo pointing outside the repo. I'm not worried about a soft link pointing to the root of the repo, because by definition it points inside the repo :-)

jacobotb: In any case, this code is not changing the behavior of the code, since those are the default values ([one](https://github.com/BurntSushi/walkdir/blob/4f26be4d450910916ea11533b2efc52b9a6483bc/src/lib.rs#L351-L352), [two](https://github.com/BurntSushi/walkdir/blob/4f26be4d450910916ea11533b2efc52b9a6483bc/src/lib.rs#L335).) It's just making them explicit so we can tell, just by looking at the source code and nothing else, that we are not following symlinks, and nobody else has to spend ten minutes digging up the documentation to make sure :-)

juli1 (approved):

PR #517, opened 2024-09-27
[STAL-2643] Export taint analysis violations to SARIF

jasonforal:
# Notes
* Ignore the "[Clean up serialization of Violation on the server.](https://github.com/DataDog/datadog-static-analyzer/pull/517/commits/744d002558ef77edf53b3d4fd3b53d3753616ee5)" commit -- this will be merged via https://github.com/DataDog/datadog-static-analyzer/pull/514
* The CI failure is expected and unrelated to this PR. It will be fixed automatically when this branch is rebased on main.=

Violations from our [Java taint analysis implementation](https://github.com/DataDog/datadog-static-analyzer/pull/493) are represented by a flow of taint from a source variable to a sink method (and thus it has multiple "code regions" associated with it). This is distinctly different than a "standard" static analysis violation, which only has one code region.

So while we can detect and flag taint flow violations within rules, we can't currently export the flows via SARIF. 

### codeFlows in SARIF
This PR modifies our SARIF output to include a ["codeFlows"](https://docs.oasis-open.org/sarif/sarif/v2.1.0/errata01/os/sarif-v2.1.0-errata01-os-complete.html#_Toc141790906) field in an analysis result.
> The `codeFlows` property is intended for use by analysis tools that provide execution path details that illustrate a possible problem in the code.

[You can see example JSON output here](https://github.com/DataDog/datadog-static-analyzer/blob/1fef7820c98c69de127b7b5b3f79a980088470e4/crates/cli/src/sarif/sarif_utils.rs#L1035-L1063).

### JavaScript Violation API
As part of this, the `Violation` class constructor [has been changed](https://github.com/DataDog/datadog-static-analyzer/blob/1fef7820c98c69de127b7b5b3f79a980088470e4/crates/static-analysis-kernel/src/analysis/ddsa_lib/js/violation.js#L27) to allow for more permissive passing in of objects/parameters to be interpreted as code regions in violations.

Before, a violation had to be created like
```javascript
const v = new Violation(node.start.line, node.start.col, node.end.line, node.end.col, "message here");
```

Now, the constructor accepts a string message and is variadic for the rest of the properties. All of the following now work (and are equivalent in terms of the resulting region):
```javascript
// "old" style
const v = new Violation("message here", node.start.line, node.start.col, node.end.line, node.end.col);
// Passing in a TreeSitterNode
const v = new Violation("message here", node);
// Passing in positions
const v = new Violation("message here", node.start, { line: node.end + 1, col: node.col });
```
Additionally, a `TaintFlow` can be passed in:
```javascript
const flows = ddsa.getTaintSources(node);
const flow = flows[0];
const v = new Violation("message here", flow);
```

Note that this doesn't affect the old stella [`buildError`](https://github.com/DataDog/datadog-static-analyzer/blob/1fef7820c98c69de127b7b5b3f79a980088470e4/crates/static-analysis-kernel/src/analysis/ddsa_lib/js/stella_compat.js#L5) API.

### Small details:
* A taint flow's fingerprint is a combination of all of the code regions it has in the flow. This uses the same methodology as our existing fingerprint.
* A taint flow is ignored if any of its regions would be ignored (regardless of which specific line has the "base" violation).

### DIFF
--- a/crates/static-analysis-kernel/src/analysis/ddsa_lib/js/violation.rs
+++ b/crates/static-analysis-kernel/src/analysis/ddsa_lib/js/violation.rs
@@ -22,6 +22,7 @@ pub struct Violation<T> {
     pub fixes: Option<Vec<Fix<T>>>,
     pub base_region: CodeRegion<T>,
     pub taint_flow_regions: Option<Vec<CodeRegion<T>>>,
+    /// (See documentation on [`Instance`]).
     pub _pd: PhantomData<T>,
 }
 
--- a/crates/static-analysis-kernel/src/analysis/ddsa_lib/js/context_file.rs
+++ b/crates/static-analysis-kernel/src/analysis/ddsa_lib/js/context_file.rs
@@ -15,6 +15,7 @@ use std::marker::PhantomData;
 #[derive(Debug)]
 pub struct FileContext<T> {
     v8_object: v8::Global<v8::Object>,
+    /// (See documentation on [`Instance`]).
     _pd: PhantomData<T>,
     // Individual language support:
     pub ctx_go: Option<js::FileContextGo<Instance>>,
--- a/crates/static-analysis-kernel/src/analysis/ddsa_lib/js/context_file_go.rs
+++ b/crates/static-analysis-kernel/src/analysis/ddsa_lib/js/context_file_go.rs
@@ -16,6 +16,7 @@ pub struct FileContextGo<T> {
     // Cached keys
     s_alias_map: v8::Global<v8::String>,
     s_is_cached: v8::Global<v8::String>,
+    /// (See documentation on [`Instance`]).
     _pd: PhantomData<T>,
 }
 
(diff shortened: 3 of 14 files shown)


crates/static-analysis-kernel/src/analysis/ddsa_lib/js/violation.rs
```diff
@@ -6,23 +6,22 @@ use crate::analysis::ddsa_lib::common::{
...
 use deno_core::v8::HandleScope;
 use std::marker::PhantomData;
 
 /// A representation of a JavaScript `Violation` class instance.
 #[derive(Debug, Clone, Eq, PartialEq, Hash)]
 pub struct Violation<T> {
-    pub start_line: u32,
-    pub start_col: u32,
-    pub end_line: u32,
-    pub end_col: u32,
```
juli1, line 23: Why removing this? It's clear - it's the section that is annotated and shown. The "regions" should just be something that contains the whole annotations but we should not touch the existing part.

juli1 (commented):
reviewing the code, it seems to replace the existing mechanism to handle position to annotate and integrate this with data flow. we should have a position that is used for annotation. and if there is data flow, we add this information on top. we should not merge these two concepts.

juli1 (approved):
Can we at least document the need for `PhantomData`?

PR #541, opened 2024-10-30
feat(server): add option to save logs to a file

robertohuertasm:

We have been observing several issues when using the server in the IDE:
- intelliJ reports having multiple instances of the server that won't be automatically shutdown
- a user in VS Code reported a case where multiple processes where present at the same time after leaving his computer idle all the night (unable to reproduce on our side)

## What is your solution?

In order to better understand what's going on it would be nice to have the generated logs of the server saved into a file.

This PR adds a new option in the CLI to roll all the logs to a file. The user can decide the cadence: minutely, hourly or daily. These logs will be using JSON format for easy parsing.

These logs are saved into the `/tmp` folder in UNIX and will default to the `TEMP` env var in windows, which is usually pointing to `C:\Users\<user_name>\AppData\Local\Temp`. If not present, for whatever reason, then it will fallback to `c:\Temp`.

The idea is to leverage this option to enable a setting in the IDEs that will allow users and the dev team to enable this feature in the binary and then easily debug the static analysis server processes. 

Tested it on Windows 11 and it works, thought the compilation process didn't work out of the box due to some UNIX commands being used in the kernel build.rs file and a wrong dependency version. I managed to fix all that. I'll prepare another PR for that.

IDE-3768

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -19,7 +19,7 @@
           "kind": "bin"
         }
       },
-      "args": ["-p", "49159", "-e"],
+      "args": ["-p", "49159", "-e", "-l"],
       "cwd": "${workspaceFolder}"
     },
     {
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -19,7 +19,10 @@
           "kind": "bin"
         }
       },
-      "args": ["-p", "49159", "-e", "-l"],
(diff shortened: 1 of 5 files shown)


robertohuertasm:
I think I didn't provide a good way of flushing the messages. I'll merge this PR anyway and provide the fix in another.

PR #543, opened 2024-10-31
fix(server): better tracing management

robertohuertasm:

There were different threads using process::exit. This was making it difficult to correctly flush the tracing messages.

On top of that, the `WorkerGuard` was shared but this was introducing other issues, as multiple references would avoid flushing the traces.

In conclusion, the mixture of these two facts was making the tracing messages not being correctly flushed

## What is your solution?

Centralized all the panics into the main thread so we can guarantee that the tracing messages are properly flushed by dropping the `WorkerGuard` if needed.

There's only one panic left in the keep alive thread. This is a last resource mechanism, and it's only being used after having tried 2 different alternatives before.

While at the task I added errors 100, 102 and 103 so we can have more useful information if we find those codes in the IDE and handle those cases accordingly.

IDE-3768


crates/bins/src/bin/datadog_static_analyzer_server/cli.rs
```diff
@@ -213,43 +211,46 @@ pub fn prepare_rocket() -> Result<RocketPreparation, CliError> {
...
+                            tracing::info!("Exiting because of timeout. Trying to exit gracefully");
+                            shutdown_handle.notify();
+                            // we give 10 seconds for the process to terminate
+                            // if it does not, we abort
+                            sleep(Duration::from_secs(10)).await;
+                            tracing::error!(
+                                "No graceful exit on first attempt. Trying another channel"
+                            );
+                            // signal the main thread to kill itself. wait for 10 secs and use abort if needed.
+                            let _ = tx_keep_alive_error.send(101).await;
```
tanishq-dubey, line 240: We should probably move the `101` here, `100` below, and other exit codes to a constant:

```rust
const ERROR_CHANNEL_DISCONNECTED: i32 = 100;
const ERROR_SHUTDOWN_FAILURE: i32 = 101;
const ERROR_UNKNOWN: i32 = 102;
const ERROR_BAD_ADDRESS: i32 = 14;
```

robertohuertasm: Good suggestion! Done!

amaanq (approved):

PR #545, opened 2024-11-02
[STAL-3099] Remove env and service

juli1:

Remove `env` and `service` from our GitHub action and hardcode it.

## What is your solution?

Hardcode values to `datadog-static-analyzer` for the service and `ci` for the env.

## Notes

Should be merged before [this PR](https://github.com/DataDog/datadog-static-analyzer-github-action/pull/47).


amaanq (approved):
approved from 30000 ft in the air

PR #553, opened 2024-11-05
[STAL-3020] Support validation

juli1:

We want to validate secrets

## What is your solution?

1. Support new secret rules formats with validation
2. Build SDS rules with validation

## Testing

Add unit. and integration test with validation (test only unvalidated secrets)

## Validation

```
Configuration
=============
version                : 0.4.5
revision               : development
config method          : remote configuration
cores available        : 10
cores used             : 8
#static analysis rules : 58
#secrets rules loaded  : 21
source directory       : /var/folders/8v/d16566g17b91b7zq6t5wwp000000gn/T/tmp.qTPdPIU5sI/
subdirectories         :
output file            : /tmp/bahar.json
secrets enabled        : true
output format          : sarif
ignore paths           : **/node_modules/**/*,**/jspm_packages/**/*,**/.next/**/*,**/.vuepress/**/*,**/venv/**/*,**/__pycache__/**/*,**/_vendor/bundle/ruby/**/*,**/.vendor/bundle/ruby/**/*,**/.bundle/**/*,**/.gradle/**/*,**/TemporaryGeneratedFile_.*.cs,**/*.designer.cs,**/*.generated.cs,**/*.g.cs,**/*.g.i.cs
only paths             : all paths
ignore gitignore       : false
use debug              : false
use staging            : false
ignore gen files       : true
rules languages        : python
max file size          : 200 kb
Found 0 violation(s) in 0 file(s) using 0 rule(s) within 0 sec(s)
███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████ 6/6Found 4 secret(s) (including 1 valid) in 6 file(s) using 21 rule(s) within 0 sec(s)
```

### DIFF
--- a/crates/cli/src/datadog_utils.rs
+++ b/crates/cli/src/datadog_utils.rs
@@ -59,7 +59,11 @@ pub fn get_secrets_rules(use_staging: bool) -> Result<Vec<SecretRule>> {
     let api_response = serde_json::from_str::<StaticAnalysisSecretsAPIResponse>(response_text);
 
     match api_response {
-        Ok(d) => Ok(d.data.iter().map(|v| v.clone().into()).collect()),
+        Ok(d) => Ok(d
+            .data
+            .iter()
+            .map(|v| v.clone().try_into().expect("cannot convert rule"))
+            .collect()),
         Err(e) => {
             eprintln!("Error when parsing the secret rules {e:?}");
             eprintln!("{response_text}");
--- a/crates/cli/src/datadog_utils.rs
+++ b/crates/cli/src/datadog_utils.rs
@@ -62,17 +62,7 @@ pub fn get_secrets_rules(use_staging: bool) -> Result<Vec<SecretRule>> {
         Ok(d) => Ok(d
             .data
             .iter()
-            .map(|v| SecretRule {
-                id: v.id.clone(),
-                name: v.attributes.name.clone(),
-                description: v.attributes.description.clone(),
-                pattern: v.attributes.pattern.clone(),
-                default_included_keywords: v
-                    .attributes
-                    .default_included_keywords
-                    .clone()
-                    .unwrap_or_default(),
-            })
+            .map(|v| v.clone().into())
             .collect()),
         Err(e) => {
             eprintln!("Error when parsing the secret rules {e:?}");
--- a/crates/cli/src/datadog_utils.rs
+++ b/crates/cli/src/datadog_utils.rs
@@ -59,11 +59,7 @@ pub fn get_secrets_rules(use_staging: bool) -> Result<Vec<SecretRule>> {
     let api_response = serde_json::from_str::<StaticAnalysisSecretsAPIResponse>(response_text);
 
     match api_response {
-        Ok(d) => Ok(d
-            .data
-            .iter()
-            .map(|v| v.clone().into())
-            .collect()),
+        Ok(d) => Ok(d.data.iter().map(|v| v.clone().into()).collect()),
         Err(e) => {
             eprintln!("Error when parsing the secret rules {e:?}");
             eprintln!("{response_text}");
--- a/crates/cli/src/datadog_utils.rs
+++ b/crates/cli/src/datadog_utils.rs
@@ -61,8 +61,8 @@ pub fn get_secrets_rules(use_staging: bool) -> Result<Vec<SecretRule>> {
     match api_response {
         Ok(d) => Ok(d
             .data
-            .iter()
-            .map(|v| v.clone().try_into().expect("cannot convert rule"))
+            .into_iter()
+            .map(|v| v.try_into().expect("cannot convert rule"))
             .collect()),
         Err(e) => {
             eprintln!("Error when parsing the secret rules {e:?}");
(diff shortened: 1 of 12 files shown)


crates/cli/src/model/datadog_api.rs
```diff
@@ -258,6 +344,24 @@ pub struct SecretRuleApiType {
...
+            name: val.attributes.name.clone(),
+            description: val.attributes.description.clone(),
+            pattern: val.attributes.pattern.clone(),
+            default_included_keywords: val
+                .attributes
+                .default_included_keywords
+                .clone()
+                .unwrap_or_default(),
+            validators: val.attributes.validators.clone(),
+            match_validation: val.attributes.match_validation.map(|v| v.into()),
```
amaanq, line 360: Like Jason said, these clones are not necessary.

```suggestion
            id: val.id,
            name: val.attributes.name,
            description: val.attributes.description,
            pattern: val.attributes.pattern,
            default_included_keywords: val
                .attributes
                .default_included_keywords
                .unwrap_or_default(),
            validators: val.attributes.validators,
            match_validation: val.attributes.match_validation.map(|v| v.into()),
```

crates/secrets/src/model/secret_rule.rs
```diff
@@ -3,10 +3,129 @@
...
+                        .set_valid_http_status_code(valid_ports)
+                        .set_method(value.http_method.unwrap().into())
+                        .build()
+                        .unwrap(),
+                ))
+            }
+
+            _ => Err("invalid type"),
+        }
+    }
```
amaanq, line 127: We're cloning in spots we don't need to, because you're using part of an Option that isn't used elsewhere, so why not just `take()` it?

```suggestion
    fn try_from(mut value: SecretRuleMatchValidation) -> Result<Self, Self::Error> {
        match value.r#type.as_str() {
            AWS_ID_STRING => Ok(MatchValidationType::Aws(AwsId)),
            AWS_SECRET_STRING => Ok(MatchValidationType::Aws(AwsSecret(AwsConfig::default()))),
            AWS_SESSION_STRING => Ok(MatchValidationType::Aws(AwsSession)),
            CUSTOM_HTTP_STRING => {
                let invalid_ports: Vec<Range<u16>> = value
                    .invalid_http_status_code
                    .take()
                    .unwrap_or_default()
                    .iter()
                    .map(|v| Range {
                        start: v.start,
                        end: v.end,
                    })
                    .collect();
                let valid_ports: Vec<Range<u16>> = value
                    .valid_http_status_code
                    .take()
                    .unwrap_or_default()
                    .iter()
                    .map(|v| Range {
                        start: v.start,
                        end: v.end,
                    })
                    .collect();

                Ok(MatchValidationType::CustomHttp(
                    HttpValidatorConfigBuilder::new(value.endpoint.clone().unwrap())
                        .set_hosts(value.hosts.take().unwrap_or_default())
                        .set_invalid_http_status_code(invalid_ports)
                        .set_request_header(value.get_request_headers())
                        .set_valid_http_status_code(valid_ports)
                        .set_method(value.http_method.unwrap().into())
                        .build()
                        .unwrap(),
                ))
            }

            _ => Err("invalid type"),
        }
    }
```

crates/cli/src/datadog_utils.rs
```diff
@@ -59,21 +59,7 @@ pub fn get_secrets_rules(use_staging: bool) -> Result<Vec<SecretRule>> {
...
-                description: v.attributes.description.clone(),
-                pattern: v.attributes.pattern.clone(),
-                default_included_keywords: v
-                    .attributes
-                    .default_included_keywords
-                    .clone()
-                    .unwrap_or_default(),
-            })
-            .collect()),
+        Ok(d) => Ok(d.data.iter().map(|v| v.clone().into()).collect()),
```
amaanq, line 62: just consume it, d isn't used elsewhere

```suggestion
        Ok(d) => Ok(d.data.into_iter().map(|v| v.into()).collect()),
```

crates/cli/src/model/datadog_api.rs
```diff
@@ -242,12 +247,93 @@ pub struct StaticAnalysisRulesAPIResponse {
...
+            request_headers: val.request_headers,
+            http_method: val.http_method.map(|v| v.into()),
+            timeout_seconds: val.timeout_seconds,
+            valid_http_status_code: val
+                .valid_http_status_code
+                .map(|v| v.iter().map(|w| w.clone().into()).collect()),
+            invalid_http_status_code: val
+                .invalid_http_status_code
+                .map(|v| v.iter().map(|w| w.clone().into()).collect()),
+        }
```
amaanq, line 325: no need to clone owned values:

```suggestion
        SecretRuleMatchValidation {
            r#type: val.r#type,
            endpoint: val.endpoint,
            hosts: val.hosts,
            request_headers: val.request_headers,
            http_method: val.http_method.map(|v| v.into()),
            timeout_seconds: val.timeout_seconds,
            valid_http_status_code: val
                .valid_http_status_code
                .map(|v| v.into_iter().map(|w| w.into()).collect()),
            invalid_http_status_code: val
                .invalid_http_status_code
                .map(|v| v.into_iter().map(|w| w.into()).collect()),
        }
```

amaanq (approved):

PR #554, opened 2024-11-06
Use `Result` for `process_analysis_request`.

jasonforal:
I'm going to be adding optional support for caching within [process_analysis_request](https://github.com/DataDog/datadog-static-analyzer/blob/b5bdd4b86bceb9caac5d43caa885050cc9281e63/crates/static-analysis-server/src/request.rs#L18), and part of this will introduce control flow that is much more readable using the idiomatic `?` operator. It's a good time to clean this function up a bit.

## What is your solution?
Refactor `process_analysis_request` to use a Result.

This is a code cleanup. The serialized response remains the exact same.

* Some `tracing` logs were deleted while refactoring because they were redundant (for example, we were logging the info message ["Validation error: configuration is not a base64 string"](https://github.com/DataDog/datadog-static-analyzer/blob/b5bdd4b86bceb9caac5d43caa885050cc9281e63/crates/static-analysis-server/src/request.rs#L24) despite our error message also being "configuration-not-base64"). The human friendly version isn't particularly necessary here.
    * No internal services rely on these strings, so they are safe to change.

### DIFF
--- a/crates/bins/src/bin/datadog_static_analyzer_server/endpoints.rs
+++ b/crates/bins/src/bin/datadog_static_analyzer_server/endpoints.rs
@@ -11,6 +11,7 @@ use rocket::{
     serde::json::{json, Json, Value},
     Build, Rocket, Shutdown, State,
 };
+use server::model::analysis_response::AnalysisResponse;
 use server::model::{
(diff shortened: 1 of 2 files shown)


amaanq (approved):
nice!

juli1 (approved):

PR #555, opened 2024-11-11
[STAL-2846] feat: add a timeout flag, and use it in the rule execution

amaanq:

Currently, we have no way to halt the analyzer during tree-sitter query execution. This is problematic because a customer or internal user could write a rule that has a query that is very slow (suffers from combinatorial explosion).

## What is your solution?

Upgrading to tree-sitter `0.24` allows us to set a timeout on query execution. This upgrade does bring in some other breaking changes that we have to adapt, including using `streaming-iterator` to iterate over the query cursor's matches or captures, which avoids UB upstream since the iterators are stateful. Another change with the upgrade is that `Node::child_containing_descendant` was deprecated in favor of `Node::child_with_descendant`, which is almost the same, except that it can return the `self` node. So, we add a check in the loop to ensure we don't process the `self` node.

With this upgrade, we add a `timeout` option to the CLI (analyzer and git hook) which takes in a number that sets the timeout for rule execution in milliseconds. This timeout applies to both tree-sitter query execution **AND** rule execution. After tree-sitter query execution finished, the timeout is subtracted by the time it took to execute the query, and this remaining duration for the timeout is passed into the rule execution. If the remainder is zero, that means we timed out executing the tree-sitter query, and we return this error, canceling the rule execution.

A test has been added that tests this timeout, by executing a quadratic query on a piece of code that is the same JavaScript function repeated 1000 times. The query wants to find two functions following one another, and given that we have 1000 of them, every single combination of these functions has to be checked during query execution, which will take a long time. As such, if we pass in a timeout of 1 second, we can observe that the tree-sitter timeout was reached, our program doesn't hang, and we get an error letting us know the query timed out.

### DIFF
--- a/crates/bins/src/bin/datadog-static-analyzer-git-hook.rs
+++ b/crates/bins/src/bin/datadog-static-analyzer-git-hook.rs
@@ -159,6 +159,12 @@ fn main() -> Result<()> {
     opts.optflag("t", "include-testing-rules", "include testing rules");
     opts.optflag("", "secrets", "enable secrets detection (BETA)");
     opts.optflag("", "static-analysis", "enable static-analysis");
+    opts.optopt(
+        "",
+        "rule-timeout-ms",
+        "how long a rule can run before being killed, in milliseconds",
+        "1000",
+    );
 
     let matches = match opts.parse(&args[1..]) {
         Ok(m) => m,
@@ -329,10 +335,19 @@ fn main() -> Result<()> {
         print_configuration(&configuration);
     }
 
+    let timeout = matches
+        .opt_str("rule-timeout-ms")
+        .map(|val| {
+            val.parse::<u64>()
+                .context("unable to parse `rule-timeout-ms` flag as integer")
+        })
+        .transpose()?;
+
     let analysis_options = AnalysisOptions {
         log_output: true,
         use_debug,
         ignore_generated_files,
+        timeout,
     };
 
     if should_verify_checksum {
--- a/crates/bins/src/bin/datadog-static-analyzer.rs
+++ b/crates/bins/src/bin/datadog-static-analyzer.rs
@@ -131,6 +131,12 @@ fn main() -> Result<()> {
     );
     // TODO (JF): Remove this when releasing 0.3.8
     opts.optflag("", "ddsa-runtime", "(deprecated)");
+    opts.optopt(
+        "",
+        "rule-timeout-ms",
+        "how long a rule can run before being killed, in milliseconds",
+        "1000",
+    );
 
     let matches = match opts.parse(&args[1..]) {
         Ok(m) => m,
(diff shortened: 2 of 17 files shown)


crates/bins/src/bin/datadog-static-analyzer-git-hook.rs
```diff
@@ -159,6 +159,12 @@ fn main() -> Result<()> {
     opts.optflag("t", "include-testing-rules", "include testing rules");
     opts.optflag("", "secrets", "enable secrets detection (BETA)");
     opts.optflag("", "static-analysis", "enable static-analysis");
+    opts.optopt(
+        "",
+        "timeout",
```
juli1, line 164: an option should give a meaning of what is does. with the unit

`rule-exec-timeout-ms` is better for example: it include the what (e.g. `rule execution`) and includes the unit (`millisecond`).

crates/bins/src/bin/datadog-static-analyzer-git-hook.rs
```diff
@@ -329,10 +335,19 @@ fn main() -> Result<()> {
         print_configuration(&configuration);
     }
 
+    let timeout = matches
+        .opt_str("timeout")
+        .map(|val| {
+            val.parse::<u64>()
+                .context("unable to parse `timeout` flag as integer")
```
juli1, line 342: update flag name

crates/bins/src/bin/datadog-static-analyzer.rs
```diff
@@ -131,6 +131,12 @@ fn main() -> Result<()> {
     );
     // TODO (JF): Remove this when releasing 0.3.8
     opts.optflag("", "ddsa-runtime", "(deprecated)");
+    opts.optopt(
+        "",
+        "timeout",
```
juli1, line 136: same comment

crates/bins/src/bin/datadog-static-analyzer.rs
```diff
@@ -368,10 +374,19 @@ fn main() -> Result<()> {
     if matches.opt_present("ddsa-runtime") {
         println!("[WARNING] the --ddsa-runtime flag is deprecated and will be removed in the next version");
     }
+    let timeout = matches
+        .opt_str("timeout")
```
juli1, line 378: same

juli1 (changes requested):
Add integration test for both the server and the SARIF side to show we return a timeout

crates/bins/src/bin/datadog-static-analyzer-git-hook.rs
amaanq: improved the name, but I disagree with the unit being in the flag name itself.

crates/static-analysis-server/src/request.rs
```diff
@@ -922,4 +924,44 @@ rulesets:
         // We should've logged the `int_node`
         assert_eq!(response.rule_responses[0].output, Some("123".to_string()));
     }
+
+    #[test]
+    fn test_query_execution_timeout() {
```
juli1, line 929: put a comment that this timeout is because of tree-sitter.

crates/bins/src/bin/datadog-static-analyzer.rs
```diff
@@ -131,6 +131,12 @@ fn main() -> Result<()> {
     );
     // TODO (JF): Remove this when releasing 0.3.8
     opts.optflag("", "ddsa-runtime", "(deprecated)");
+    opts.optopt(
+        "",
+        "max-rule-time",
```
juli1, line 136: said previously, put a name that is meaningful **AND** contains the UNIT

`rule-timeout-ms`

juli1 (changes requested):

juli1 (approved):

PR #565, opened 2024-12-09
[K9VULN-2229] Various SARIF mod code cleanups

jasonforal:
(For better organization, this PR is split off from one that implement SARIF artifacts).

We'll be adding the [`run/artifacts` property](https://docs.oasis-open.org/sarif/sarif/v2.1.0/errata01/os/sarif-v2.1.0-errata01-os-complete.html#_Toc141790749) from the SARIF spec in an upcoming PR.

While implementing that, I ran into some minor annoyances:
* Adding a new property to output broke unrelated tests because those tests hardcode the entire expected SARIF. Ideally, each unit test only targets a specific behavior.
* For each file, we currently have three different serialized forms (all opaquely using the `String` type).
    * The path as reported by the `walkdir` (which will be the OS-native form)
    * The unix path (i.e. on Windows systems, the path from above will be converted to unix format)
    * A percent-encoded path (e.g. converting `[` to `%5B`)
 As-is, these transformations are performed by functions with vague names.

## What is your solution?
* Narrow down each unit test to only test specific behavior.
* Rename functions to be more descriptive (and a small nit to reduce allocations)

* A better solution for the path issue would be to [newtype](https://doc.rust-lang.org/rust-by-example/generics/new_types.html) each of the 3 formats for paths so we can have compile-time safety. However, this is foregone to keep changes as minimal as possible.

### DIFF
--- a/crates/cli/src/sarif/sarif_utils.rs
+++ b/crates/cli/src/sarif/sarif_utils.rs
@@ -193,17 +193,12 @@ impl SarifRuleResult {
         }
     }
 
-    fn file_path(&self) -> String {
-        match self {
-            SarifRuleResult::StaticAnalysis(r) => Path::new(r.filename.as_str())
-                .to_slash()
-                .unwrap()
-                .to_string(),
-            SarifRuleResult::Secret(r) => Path::new(r.filename.as_str())
-                .to_slash()
-                .unwrap()
-                .to_string(),
-        }
+    /// Returns the file path for this result as a slash path. See [`as_slash_path`] for documentation.
+    fn slash_path_str(&self) -> std::borrow::Cow<'_, str> {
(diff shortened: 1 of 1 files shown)


crates/cli/src/sarif/sarif_utils.rs
```diff
@@ -1279,91 +1296,175 @@ mod tests {
...
+                            "text": "secret-rule"
+                          }
+                        }
+                      ],
+                    }
+                  }
+                }
+              ],
+            }
+                    );
```
amaanq, line 1457: Not a blocker, but I'm not sure if the indentation for the closing paren was expected (if it is for visual distinction from the json punctuation, that's totally fine)

amaanq (approved):

PR #569, opened 2024-12-10
[K9VULN-2229] Add `is_test_file` with path-based classification and SARIF output.

jasonforal:
In https://github.com/DataDog/datadog-static-analyzer/pull/568, we added the run [artifacts property](https://docs.oasis-open.org/sarif/sarif/v2.1.0/errata01/os/sarif-v2.1.0-errata01-os-complete.html#_Toc141790749) to our SARIF output, with the intention of using an artifact's property bag to tag "test file" artifacts.

We still need to implement that logic of adding the per-artifact tags.

## What is your solution?
The test file classifier will have two strategies that are unique to each language:
1. Classify based on file path.
2. Classify based on file contents (currently planned: analyzing package/module/etc imports).

This PR only implements no.1, but includes the structure for eventually implementing no.2.

### Summary
By default, if a file is contained in a folder with a specific name ([examples here](https://github.com/DataDog/datadog-static-analyzer/blob/b2ffbfd7a315b9bd9b512068e3104fa0e9f4cfad/crates/static-analysis-kernel/src/classifiers/tests.rs#L28-L37)) or the filename matches a specific pattern ([examples here](https://github.com/DataDog/datadog-static-analyzer/blob/b2ffbfd7a315b9bd9b512068e3104fa0e9f4cfad/crates/static-analysis-kernel/src/classifiers/tests.rs#L40)), the SARIF output for that file will contain the tag `"DATADOG_ARTIFACT_IS_TEST_FILE"`.

**Before (runs/0/artifacts)**
```json
[
  {
    "location": {
      "uri": "src/core/some_file.js"
    }
  },
  {
    "location": {
      "uri": "src/tests/app.router.js"
    }
  }
]
```

**After (runs/0/artifacts)**
```json
[
  {
    "location": {
      "uri": "src/core/some_file.js"
    }
  },
  {
    "location": {
      "uri": "src/tests/app.router.js"
    },
    "properties": {
      "tags": [
        "DATADOG_ARTIFACT_IS_TEST_FILE"
      ]
    }
  }
]
```

---

#### Classifiers module
`static-analysis-kernel::classifiers` has been introduced, and it will serve as the place to put all logic that analyzes file contents (or parse trees) to attach metadata. Eventually, our [generated_content.rs](https://github.com/DataDog/datadog-static-analyzer/blob/80b97f5567d7ae433292845a97b4d6a86f4c8997/crates/static-analysis-kernel/src/analysis/generated_content.rs) module should be moved here.

[`ArtifactClassification`](https://github.com/DataDog/datadog-static-analyzer/blob/b2ffbfd7a315b9bd9b512068e3104fa0e9f4cfad/crates/static-analysis-kernel/src/classifiers.rs#L10) is a struct containing metadata associated with an artifact. Right now, we only implement "is_test_file", however, this struct can be expanded to accommodate more "classifications", e.g. generated content, or minified file.

#### Hooks for classifications per-language
This PR establishes the pattern for customizing the classifiers for each language:
* Path-based classifiers can be modified [here](https://github.com/DataDog/datadog-static-analyzer/blob/b2ffbfd7a315b9bd9b512068e3104fa0e9f4cfad/crates/static-analysis-kernel/src/classifiers/tests.rs#L81-L83). For example, for Go, all we need to detect is a file ending in `_test.go`, so we don't need to perform the same scans that other languages may need.
* File content (and/or parse tree) classifiers can be modified [here](https://github.com/DataDog/datadog-static-analyzer/blob/b2ffbfd7a315b9bd9b512068e3104fa0e9f4cfad/crates/static-analysis-kernel/src/classifiers/tests.rs#L87-L110).

#### Classification executes once-per-file
Each rayon thread keeps a local HashMap that contains a mapping from file paths to classifications, allowing us to only [perform the classification once per file](https://github.com/DataDog/datadog-static-analyzer/blob/b2ffbfd7a315b9bd9b512068e3104fa0e9f4cfad/crates/bins/src/bin/datadog-static-analyzer.rs#L616-L631). This is important because classification is (and should be) considered an expensive operation, and there could be multiple violations within a single file.

### Testing
* [Unit test](https://github.com/DataDog/datadog-static-analyzer/blob/b2ffbfd7a315b9bd9b512068e3104fa0e9f4cfad/crates/cli/src/sarif/sarif_utils.rs#L1817-L1824) confirms that `generate_sarif_report` correctly maps an artifact with `ArtifactClassification { is_test_file: true }` to a SARIF property tag `"DATADOG_ARTIFACT_IS_TEST_FILE"`
* [Integration test](https://github.com/DataDog/datadog-static-analyzer/blob/b2ffbfd7a315b9bd9b512068e3104fa0e9f4cfad/misc/integration-test-classification.sh) confirms the analyzer binary outputs SARIF with all artifacts tagged properly.


crates/bins/src/bin/datadog-static-analyzer.rs
```diff
@@ -610,6 +613,25 @@ fn main() -> Result<()> {
...
+                                std::path::Path::new(&cloned_path_str),
+                                None,
+                            ) {
+                                Some(ArtifactClassification { is_test_file: true })
+                            } else {
+                                None
+                            };
+                            path_metadata.insert(cloned_path_str, metadata);
+                        }
+
```
juli1, line 634: 1. Do not think it should be sarif-specific.
2. Can we try to extract this into a function that can be reused by the git hooks?

crates/static-analysis-kernel/src/classifiers.rs
```diff
@@ -0,0 +1,14 @@
...
+mod tests;
+pub use tests::is_test_file;
+
+/// Metadata associated with an artifact that has been analyzed.
+#[derive(Debug, Clone, Eq, PartialEq, Default)]
+pub struct ArtifactClassification {
+    /// Whether the artifact is considered a "test file" or not. A test file is
+    /// language-dependent and either contains unit tests or is associated with unit test frameworks.
+    pub is_test_file: bool,
+}
```
juli1, line 14: I like this idea of classifying files with attributes. It's a very nice idea.

crates/static-analysis-kernel/src/classifiers/tests.rs
```diff
@@ -0,0 +1,176 @@
...
+pub fn is_test_file(
+    language: Language,
+    code: &str,
+    file_path: &Path,
+    pre_parsed_tree: Option<&tree_sitter::Tree>,
+) -> bool {
+    // This function implements a tiered classification:
+    //
+    // 1. File path heuristic:
+    if FILENAME_CONVENTION_GLOB.is_match(file_path) {
```
juli1, line 21: Will be more `getConventionGlob(language).is_match(file_path)`

misc/_test-classification.sh
```diff
@@ -0,0 +1,107 @@
...
+#
+# It tests that:
+# * Each `js` file has exactly one detection.
+# * The SARIF `artifacts` array contains exactly one object for each file.
+# * A file name that starts with "should_be_t_file" should have its SARIF artifact tagged as a test file
+#   (Note that the analyzer itself does not use this specific name as a detection -- this naming schema
+#   is used as a way to declaratively and robustly define the test corpus).
+# * A file name not starting with that string should not have its SARIF artifact tagged as a test file.
+
+# The SARIF property bag tag that a test file artifact should have.
```
juli1, line 19: Why not merging this in `integration-test-classification.sh`?

juli1 (approved):
We should probably combine the test scripts

PR #573, opened 2024-12-11
Specify error code

juli1:

We want to distinguish each error case so that we can surface in the datadog UI why a run failed.

## What is your solution?

Put a specific exit code for each failure.

 - User-specific failures are more than 50
 - Program failures are less than 50


crates/cli/src/constants.rs
jacobotb, line 1: We don't need that many exit statuses; just two: one for "the error is in the data" and another for "the error is in the program". More codes just make it more complicated.

juli1: Anything more than 50 is user, lower is specific to the program itself

jacobotb (approved):

PR #586, opened 2024-12-18
[K9VULN-2350] Implement test file detection for C#

jasonforal:
The analyzer doesn't accurately classify C# files that contain or are related to tests.

## What is your solution?
* Implement a parser for C# using directives. ([Language reference here](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/using-directive))
* Add detections for common C# test package namespaces ([here](https://github.com/DataDog/datadog-static-analyzer/blob/dbe1bd144eed0a01c05ca81bf8f8588223a5f023/crates/static-analysis-kernel/src/classifiers/tests.rs#L145-L195)).
* Add file-path detections for common C# naming conventions ([here](https://github.com/DataDog/datadog-static-analyzer/blob/dbe1bd144eed0a01c05ca81bf8f8588223a5f023/crates/static-analysis-kernel/src/classifiers/tests.rs#L108))

* Parsing imports with the [namespace alias operator](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/namespace-alias-qualifier) is unimplemented because it adds complication to the logic and none of our current detections need it. This limitation is documented [here](https://github.com/DataDog/datadog-static-analyzer/blob/dbe1bd144eed0a01c05ca81bf8f8588223a5f023/crates/static-analysis-kernel/src/analysis/languages/csharp/using_directives.rs#L183), [here](https://github.com/DataDog/datadog-static-analyzer/blob/dbe1bd144eed0a01c05ca81bf8f8588223a5f023/crates/static-analysis-kernel/src/analysis/languages/csharp/using_directives.rs#L82-L90), and [here](https://github.com/DataDog/datadog-static-analyzer/blob/dbe1bd144eed0a01c05ca81bf8f8588223a5f023/crates/static-analysis-kernel/src/analysis/languages/csharp/using_directives.rs#L132-L133).
* The "Software Composition Analysis" [CI job](https://github.com/DataDog/datadog-static-analyzer/actions/runs/12398382985/job/34610934504) is expected to fail due to upstream issues unrelated to this PR.

### DIFF
--- a/crates/static-analysis-kernel/src/classifiers/tests.rs
+++ b/crates/static-analysis-kernel/src/classifiers/tests.rs
@@ -105,6 +105,7 @@ macro_rules! trie_from {
 fn has_test_like_path(language: Language, path: &Path) -> bool {
     use Language::*;
     let globset = match language {
+        Csharp => globset_from!([DEFAULT_PATHS, &["**/*Test.cs", "**/*Tests.cs"]]),
         Go => globset_from!([&[
             // `go test` required filename
             "**/*_test.go",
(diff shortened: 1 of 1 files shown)


crates/static-analysis-kernel/src/classifiers/tests.rs
```diff
@@ -303,6 +369,26 @@ mod cfg_test_tests {
         assert!(!trie_has_prefix(&trie, "no.sequence.match".split(".")));
     }
 
+    #[test]
+    fn language_csharp() {
+        let shoulds = &[
+            "using Xunit;",
+            "using UT = Microsoft.VisualStudio.TestTools.UnitTesting;",
+        ];
+        let should_nots = &["using System.Diagnostics;"];
```
amaanq, line 378: nit(non-blocking): a good test case might be to ensure we never consider the alias, e.g:

```cs
using Xunit = CustomFramework;
```

though it's not necessary at all

amaanq (approved):

amaanq (approved):

PR #588, opened 2024-12-19
[K9VULN-1926] feat: add query timeout functionality to the server

amaanq:

In https://github.com/DataDog/datadog-static-analyzer/pull/555 we added support for timing out tree-sitter query executions, as this was necessary to prevent the analyzing spinning forever in case a query suffered from combinatorial explosion. However, we did not add the CLI flag added to the main analyzer binary to the server, when the server could leverage this timeout flag as well for requests made to the `/analyze` endpoint. It would make sense for both the CLI and the server to have the same flag to control the duration of the timeout.

## What is your solution?

This PR adds a flag, `rule-timeout-ms`, to the server binary, which is the exact same flag the CLI contains [here](https://github.com/DataDog/datadog-static-analyzer/blob/bce3e2d1cd9f5592299f195cad8b9e4422a54d36/crates/bins/src/bin/datadog-static-analyzer.rs#L143). The parsing logic for the timeout flag is the same as it is for the CLI binary as well, and if parsing fails, we surface this to the user with a helpful error message. The resulting number parsed from this timeout flag is then stored in the server's [`ServerState`](https://github.com/DataDog/datadog-static-analyzer/blob/15617a14a79e22e7e64ec1d7531e90b2eae8e49e/crates/bins/src/bin/datadog_static_analyzer_server/state.rs#L5), which can be passed into the handler for `/analyze`. Inside this handler, we pass the timeout into `process_analysis_request`, which is then attached to the `AnalysisOptions` of the "main" analysis function `analyze_with`.

This PR doesn't need a new test, as there is currently a test that checks the timeout works for the server [here](https://github.com/DataDog/datadog-static-analyzer/blob/15617a14a79e22e7e64ec1d7531e90b2eae8e49e/crates/static-analysis-server/src/request.rs#L944), but the code snippet used in here has an extraneous equals sign, so I removed it. The intention is that the code snippet is a function in javascript repeated `100000` times, as seen in [this](https://github.com/DataDog/datadog-static-analyzer/blob/15617a14a79e22e7e64ec1d7531e90b2eae8e49e/crates/static-analysis-server/src/request.rs#L944) test. 

Additionally, as tokio requires that the paramter `F` passed into `spawn_blocking` be `'static`, we have to expliclty move the timeout variable into the closure, as the closure could outlive the `analyze` function.


juli1 (changes requested):
Let's use a duration. Ideally, we should have used this all along the codebase.

juli1 (approved):

PR #593, opened 2024-12-26
[K9VULN-2162] Add rule caching to datadog-static-analyzer-server binary

jasonforal:
Our IDE plugin can run batch analyses (e.g on an entire workspace). Under the hood, this uses the [datadog-static-analyzer-server](https://github.com/DataDog/datadog-static-analyzer/blob/main/crates/bins/src/bin/datadog-static-analyzer-server.rs) binary, which runs as a local HTTP server. For each file, the IDE plugin makes an HTTP request containing the file contents, metadata, and a JSON serialization of every rule to run against the file.

The server is currently stateless, and so for every HTTP request, each rule needs to be compiled (from a JSON string into a tree-sitter query and a v8 script). This gives us `O(n x m)` time complexity for rule compilation, where `n` is number of files, and `m` is number of rules.

Rule compilation is expensive (some taking hundreds of milliseconds). A workspace scan of a C# repo with 6.7k files using 82 rules took over 13 minutes and had the following flamegraph:

**Before: 6,700 files, 82 rules: 793 seconds**
image

As can be seen, 96% of the trace samples were from compiling rules: 6,700 x 82 == 549,400 compilations, whereas the algorithmic optimum—`O(m)`—is only 82 times.

## What is your solution?

Add a caching layer to the server to store compiled rules.

**After: 6,700 files, 82 rules: 29.7 seconds**
image

Now, only 10.5% of the trace samples came from compiling rules and the scan took 30 seconds—a 25x speedup.

This functionality is enabled by passing CLI flag `--use-rules-cache` to the `datadog-static-analyzer-server` binary.

### Server library unaware of caching
`process_analysis_request` in the server library [has been changed](https://github.com/DataDog/datadog-static-analyzer/blob/22b3e57db43d525b96fda49a39ede23c100ed570/crates/static-analysis-server/src/request.rs#L19-L20) to expect a list of `T: Borrow<RuleInternal>` instead of `ServerRule` (previously, this function would handle the conversion from ServerRule to `RuleInternal`).

This allows callers (in this case, the `datadog-static-analyzer-server`) to implement, e.g. their own caching layer without concerning the library with such details.

### Security
The caching layer introduces a vector for resource denial. For example, the custom rule editor within the Datadog app is powered by the `datadog-static-analyzer-server` binary. Because the cache is keyed by rule name, an adversary could flood the cache with arbitrarily long (and/or random) key names in order to persist unbounded bytes of memory.

Rather than solve this with more complexity (e.g. an LRU cache), it's the most straightforward to just [enable caching via a CLI flag](https://github.com/DataDog/datadog-static-analyzer/blob/459490dd6cfa51997edc21f51dd1f8d311ed181a/crates/bins/src/bin/datadog_static_analyzer_server/cli.rs#L180-L182) so that the IDE can turn it on and the custom rule editor can turn it off.

### Concurrency
**Memory Footprint**
To reduce the memory footprint for IDEs, we use a single cache (versus, e.g. multiple thread-local caches). This requires synchronization considerations to preserve a high throughput. We thus add a new 3rd party dependency ([DashMap](https://github.com/xacrimon/dashmap)) as a superior alternative to a `RwLock<HashMap<K, V>>`.

**Why `Arc<CompiledServerRule>` instead of `CompiledServerRule`?**
For memory efficiency, the cache is keyed by the rule name (which can have collisions), and the value is an [Arc\<CompiledServerRule\>](https://github.com/DataDog/datadog-static-analyzer/blob/22b3e57db43d525b96fda49a39ede23c100ed570/crates/bins/src/bin/datadog_static_analyzer_server/rule_cache.rs#L32).

Consider a pathological case where:
* "Rule A" and "Rule B" have the same rule name but different rule logic.
* "Rule A" takes 10ms to execute and "Rule B" takes 20ms to execute
* There are only two threads and requests arrive in the sequence "Rule A, Rule B, Rule A, Rule B, Rule A..."

A `DashMap<String, CompiledServerRule>` yields borrowed values (`&CompiledServerRule`). Each thread must hold the lock for the entire rule compilation and rule execution. This devolves into serial (single-threaded) execution.

Instead, a `DashMap<String, Arc<CompiledServerRule>>` yields owned values (`Arc<CompiledServerRule>`). Each thread only needs to hold the lock long enough to increment an `Arc`. Additionally, this also means that the cache can be updated at any time, even if other threads are still processing a request for "Rule A" or "Rule B".

```
Legend:
...   Waiting for Lock
[*]   Lock Acquired
A-A   Execution of Rule A
B-B   Execution of Rule B

// With &CompiledServerRule
Thread 1: [*]A--------A . . . . . . . . . . . [*]A--------A . . . . . . . . . . . [*]A--------A
Thread 2: . . . . . . .[*]B------------------B . . . . . . [*]B------------------B

// With Arc<CompiledServerRule>
Thread 1: [*]A--------A[*]A--------A[*]B------------------B
Thread 2: ...[*]B------------------B...[*]A--------A
```

In the above pathological case, the server at worst "devolves" to the current (cacheless) status quo, as each execution ends up compiling and inserting either "Rule A" or "Rule B" into the cache every time.

## Alternatives considered
* Keep the server stateless, but introduce a batch analysis endpoint (so we'd still have `O(n x m)` complexity, but in practice, `n` would be much smaller).
* Change the server library's `process_analysis_request` to accept an `Option<RuleCache>`.

* This will significantly reduce latency in ad-hoc analyses (e.g. when a user is typing in the IDE)

### DIFF
--- a/crates/bins/src/bin/datadog_static_analyzer_server/cli.rs
+++ b/crates/bins/src/bin/datadog_static_analyzer_server/cli.rs
@@ -14,6 +14,8 @@ use tracing_subscriber::EnvFilter;
 use crate::datadog_static_analyzer_server::error_codes::{
     ERROR_CHANNEL_DISCONNECTED, ERROR_SHUTDOWN_FAILURE,
 };
+use crate::datadog_static_analyzer_server::rule_cache::RuleCache;
+use crate::RULE_CACHE;
 
 use super::state::ServerState;
 use super::utils::get_current_timestamp_ms;
@@ -42,6 +44,7 @@ fn get_opts() -> Options {
     opts.optflag("e", "enable-shutdown", "Enables the shutdown endpoint");
     opts.optflag("h", "help", "Print this help");
     opts.optflag("v", "version", "Shows the tool version");
+    opts.optflag("c", "use-rules-cache", "Pre-compile and cache rules");
 
     opts.optopt(
         "l",
@@ -171,6 +174,11 @@ pub fn prepare_rocket(tx_keep_alive_error: Sender<i32>) -> Result<RocketPreparat
     // server state
     tracing::debug!("Preparing the server state and rocket configuration");
     let mut server_state = ServerState::new(matches.opt_str("s"), matches.opt_present("e"));
+    // Create a cache for rules, if requested
+    if matches.opt_present("use-rules-cache") {
+        RULE_CACHE.set(RuleCache::new()).expect("should init once");
+    }
+
     let mut rocket_configuration = rocket::config::Config {
         // disable rocket colors and emojis if we're logging to a file as we will be using json format
         cli_colors: !matches.opt_present("l"),
--- a/crates/bins/src/bin/datadog-static-analyzer-server.rs
+++ b/crates/bins/src/bin/datadog-static-analyzer-server.rs
@@ -2,12 +2,15 @@
 // This product includes software developed at Datadog (https://www.datadoghq.com/).
 // Copyright 2024 Datadog, Inc.
 
+use crate::datadog_static_analyzer_server::rule_cache::RuleCache;
 use kernel::analysis::ddsa_lib::v8_platform::{initialize_v8, Initialized, V8Platform};
 use std::sync::OnceLock;
+
 mod datadog_static_analyzer_server;
 
 pub(crate) static V8_PLATFORM: OnceLock<V8Platform<Initialized>> = OnceLock::new();
 pub(crate) static RAYON_POOL: OnceLock<rayon::ThreadPool> = OnceLock::new();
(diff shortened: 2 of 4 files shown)


crates/bins/src/bin/datadog_static_analyzer_server/cli.rs
```diff
@@ -42,6 +44,7 @@ fn get_opts() -> Options {
     opts.optflag("e", "enable-shutdown", "Enables the shutdown endpoint");
     opts.optflag("h", "help", "Print this help");
     opts.optflag("v", "version", "Shows the tool version");
+    opts.optflag("", "use-rules-cache", "Pre-compile and cache rules");
```
juli1, line 47: Why not a `-c` for a short option?

juli1 (approved):
Let's add the short-option flavor and ship

PR #596, opened 2025-01-04
[K9VULN-2502] Decouple timeout watchdog from `JsRuntime`

jasonforal:
We will be adding the ability to gracefully recover from a JavaScript execution that causes v8 to run out of memory.

To do this, we need to provide a callback ([NearHeapLimitCallback](https://v8.github.io/api/head/namespacev8.html#aa1fc7d6f7736e8256fc03c4b3ebaf6eb)) to v8, which is a fundamentally different model from the condvar/mutex we use for our timeout watchdog.

We should thus decouple the timeout logic from the `JsRuntime` so that we can handle the increased complexity in an encapsulated way. (But in my opinion, this would be a good refactor irrespective--it's a design smell that our [scoped_execute](https://github.com/DataDog/datadog-static-analyzer/blob/8f77360139fa8f94a00473047b4ab6a00983dc66/crates/static-analysis-kernel/src/analysis/ddsa_lib/runtime.rs#L334-L342) function needs to manually deal with internal implementation details like notifying condvars).

## What is your solution?
Remove all timeout logic from the ddsa `JsRuntime` and introduce a new, more generic "ResourceWatchdog". This will eventually contain the out-of-memory v8 callback implementation.

> [!NOTE]
> **There is no change to runtime behavior or thread synchronization behavior**
> This PR is just a structural refactor

### Notable Implementation Details
**Watchdog state now stores termination reason**
```rs
struct WatchdogState {
    timeout: TimeoutState,
    /// This will be `Some` if there was a termination. Otherwise, it will be `None`.
    termination_err: Option<DDSAJsRuntimeError>,
}
```
We need to distinguish between a termination from the v8 callback (out of memory) and our thread (timeout). This requires a mutex to both perform the termination and signal the reason for termination. While channels are probably a better way to encapsulate that communication, I'd like to keep using a Mutex to take advantage of Condvar's `wait_timeout` api.

We now need to make sure that this `WatchdogState` is manually cleared after each execution. This is extensively tested with [all possible state transition permutations here](https://github.com/DataDog/datadog-static-analyzer/commit/21d59181b1a28b7a79f7e415a15662c1cfcf31ad#diff-1a31fba10caf14d595b6f377e5f7e9de7d7c1b911ece7e7800573d783ef76d87R216-R224).

**Caller is no longer concerned with implementation details**
Calls needing resource limitation just need make a call through the watchdog. Simple:
```rs
let execution_result = self
    .v8_resource_watchdog
    .execute(timeout, tc_ctx_scope, |sc| bound_script.run(sc))?;
```
(See `execute` function signature [here](https://github.com/DataDog/datadog-static-analyzer/blob/21d59181b1a28b7a79f7e415a15662c1cfcf31ad/crates/static-analysis-kernel/src/analysis/ddsa_lib/resource_watchdog.rs#L41-L53))

* For slightly less-intimidating diff, see https://github.com/DataDog/datadog-static-analyzer/commit/d1836d27b255fa3775ab0604033639d03826c5b3 to see how "JsExecutionState" was simplified.
* [`spawn_timeout_thread`](https://github.com/DataDog/datadog-static-analyzer/commit/21d59181b1a28b7a79f7e415a15662c1cfcf31ad#diff-1a31fba10caf14d595b6f377e5f7e9de7d7c1b911ece7e7800573d783ef76d87R94-R151) is a copy/paste of the prior implementation with no behavior change.

### DIFF
--- a/crates/static-analysis-kernel/src/analysis/ddsa_lib.rs
+++ b/crates/static-analysis-kernel/src/analysis/ddsa_lib.rs
@@ -9,6 +9,7 @@ pub use context::*;
 pub mod extension;
 pub(crate) mod js;
 pub(crate) mod ops;
+pub(crate) mod resource_watchdog;
 pub(crate) mod runtime;
 pub use runtime::JsRuntime;
 #[allow(dead_code)]
--- a/crates/static-analysis-kernel/src/analysis/ddsa_lib/resource_watchdog.rs
+++ b/crates/static-analysis-kernel/src/analysis/ddsa_lib/resource_watchdog.rs
@@ -1,0 +1,264 @@
+// Unless explicitly stated otherwise all files in this repository are licensed under the Apache License, Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2025 Datadog, Inc.
+
+use crate::analysis::ddsa_lib::common::DDSAJsRuntimeError;
+use deno_core::v8;
+use std::sync::{Arc, Condvar, Mutex};
+use std::time::{Duration, Instant};
+
+/// A persistent thread that can track execution on a [`v8::Isolate`] and terminate it if
+/// a per-execution configurable duration has been exceeded.
+#[derive(Debug)]
+pub(crate) struct V8ResourceWatchdog {
+    state: Arc<Mutex<WatchdogState>>,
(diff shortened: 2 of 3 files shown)


juli1 (approved):
please review the comment below, fix if you think it's appropriate and ship

PR #599, opened 2025-01-07
[K9VULN-2580] Add Elixir support

juli1:

Add Elixir support

## What is your solution?

Add the [tree-sitter grammar](https://github.com/elixir-lang/tree-sitter-elixir/tree/main) from the `elixir-lang` GitHub account.

Added unit tests to ensure parsing works correctly.

### DIFF
--- a/crates/cli/src/file_utils.rs
+++ b/crates/cli/src/file_utils.rs
@@ -16,7 +16,7 @@ use kernel::model::violation::Violation;
 static FILE_EXTENSIONS_PER_LANGUAGE_LIST: &[(Language, &[&str])] = &[
     (Language::Csharp, &["cs"]),
     (Language::Dockerfile, &["docker", "dockerfile"]),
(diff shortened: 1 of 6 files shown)


crates/cli/src/file_utils.rs
```diff
@@ -16,6 +16,7 @@ use kernel::model::violation::Violation;
 static FILE_EXTENSIONS_PER_LANGUAGE_LIST: &[(Language, &[&str])] = &[
     (Language::Csharp, &["cs"]),
     (Language::Dockerfile, &["docker", "dockerfile"]),
+    (Language::Elixir, &["ex"]),
```
amaanq, line 19: Add .exs as well, which is the scripting file extension - https://hexdocs.pm/elixir/modules-and-functions.html#scripting-mode

```suggestion
    (Language::Elixir, &["ex", "exs"]),
```

amaanq (approved):

PR #602, opened 2025-01-10
Handle `no-dd-sa` with comment blocks

juli1:

We do have customers who want to specify the `no-dd-sa` in block comments.

For example, code such as

```
# no-dd-sa
# another comment
def line_with_violation(bla):
  pass
```

The `no-dd-sa` should then mute violation for line `def line_with_violation(bla):`

This happens especially when customers are running multiple linters (e.g. eslint) and ignore also eslint findings. For example, in the following case, we want to disable the violation for `doSomethingThatTriggerAViolation();`

```
// no-dd-sa
// eslint-disable-next-line
doSomethingThatTriggerAViolation();
```

## What is your solution?

We see if the `no-dd-sa` directive is in a comment block. When we find a `no-dd-sa` comment, we check is there was comments below and if yes, ignore the next line until it's not a comment.

### DIFF
--- a/crates/static-analysis-kernel/src/analysis/analyze.rs
+++ b/crates/static-analysis-kernel/src/analysis/analyze.rs
@@ -20,6 +20,7 @@ use std::time::{Duration, Instant};
 /// This includes the time it takes for the tree-sitter query to collect its matches, as well as
 /// the time it takes for the JavaScript rule to execute.
 const RULE_EXECUTION_TIMEOUT: Duration = Duration::from_millis(2000);
+const DISABLE_KEYWORDS: [&str; 2] = ["no-dd-sa", "datadog-disable"];
 
 /// Split the code and extract all the logic that reports to lines to ignore.
 /// If a no-dd-sa statement occurs on the first line, it applies to the whole file.
@@ -29,7 +30,7 @@ fn get_lines_to_ignore(code: &str, language: &Language) -> LinesToIgnore {
     let mut lines_to_ignore_per_rules: HashMap<u32, Vec<String>> = HashMap::new();
 
     let mut line_number = 1u32;
-    let comment_starters = match language {
+    let comment_prefixes = match language {
         Language::Python
         | Language::Starlark
         | Language::Dockerfile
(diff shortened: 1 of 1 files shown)


crates/static-analysis-kernel/src/analysis/analyze.rs
```diff
@@ -29,6 +29,51 @@ fn get_lines_to_ignore(code: &str, language: &Language) -> LinesToIgnore {
...
+        Language::PHP => {
+            vec!["//", "/*", "#"]
+        }
+        Language::Markdown => {
+            vec!["<!-"]
+        }
+        Language::SQL => {
+            vec!["--", "/*"]
+        }
+    };
```
amaanq, line 62: IMO, this feels like we're repeating

amaanq, line 62: We're sorta duplicating information in two match blocks, one here and one later on when we compute "disabling_patterns" - why don't we have a look-up map for language -> comment prefix, and then have `disabling_patterns` be computed independent of the language (removing the "duplicate" match), e.g. something like

```rs
["no-dd-sa", "datadog-disable"]
    .iter()
    .map(|pattern| 
        comment_prefixes // e.g. ["//", "/*"]
            .iter()
            .map(|prefix| prefix + pattern) // gives us a 2x2 matrix of all possible patterns, so we'll flatten it
    ).flatten()
```

crates/static-analysis-kernel/src/analysis/analyze.rs
```diff
@@ -29,6 +29,51 @@ fn get_lines_to_ignore(code: &str, language: &Language) -> LinesToIgnore {
...
+    // comment blocks that contains no-dd-sa.
+    let comment_lines = code
+        .lines()
+        .enumerate()
+        .filter(|(_idx, line)| {
+            comment_starters
+                .iter()
+                .any(|m| line.replace(" ", "").starts_with(m))
+        })
+        .map(|(idx, _line)| (idx + 1) as u32)
```
amaanq, line 74: For unused variables, just use `_` to indicate we're discarding it.
```suggestion
        .filter(|(_, line)| {
            comment_starters
                .iter()
                .any(|m| line.replace(" ", "").starts_with(m))
        })
        .map(|(idx, _)| (idx + 1) as u32)
```

Or, you could use filter_map

```suggestion
        .filter_map(|(idx, line)| {
            comment_starters
                .iter()
                .any(|m| line.replace(" ", "").starts_with(m))
                .then_some((idx + 1) as u32)
        })
```

amaanq (approved):

PR #608, opened 2025-01-15
[K9VULN-2775] Fix gaps in default ruleset CI coverage

jasonforal:
If no configuration is specified, the analyzer CLI uses a default list of languages, specified in `datadog_utils.rs`:
```rs
const DEFAULT_RULESETS_LANGUAGES: &[&str] = &[
    "CSHARP",
    // ...and so on
];
```
([source](https://github.com/DataDog/datadog-static-analyzer/blob/f088246e22a46238556e625729e7d923847317e0/crates/cli/src/datadog_utils.rs#L30-L42))

In CI, we've been manually syncing that list of languages to ensure we test all default rulesets:
```sh
for language in go python typescript javascript csharp java ruby; do
    # ...test logic
done
```
([source](https://github.com/DataDog/datadog-static-analyzer/blob/f088246e22a46238556e625729e7d923847317e0/.github/workflows/test-rules.yaml#L29))

These two have drifted out of sync (presently, the CI job is missing "Dockerfile", "Kotlin", and "PHP")

## What is your solution?
Ensure CI coverage can't drift by dynamically parsing `datadog_utils.rs` and extracting the list of languages. That file is now the single source of truth.

### Robustness
This script always makes sure it can parse _something_, [otherwise it returns an error](https://github.com/DataDog/datadog-static-analyzer/commit/ccf5dc87be52a33e75ed171b45c687f3a14f6697#diff-b1a7cc3b15bf6b3765328ae0ceb6d1ed680656747bb629b3fb294973496ff959R45-R48). Thus, if the "DEFAULT_RULESETS_LANGUAGES" code is refactored, the "extract-languages" job will fail until it's updated (rather than looping over zero languages, causing the entire test to inaccurately be marked successful)

### Testing
You can confirm that the CI job correctly tests all of the languages by [viewing the logs of this successful run](https://github.com/DataDog/datadog-static-analyzer/actions/runs/12779832469/job/35625115665#step:4:175).

## Alternatives considered
* Writing the parse logic in a more universally understandable form--e.g a Python script--instead of `grep`, `tr`, and `sed`.

### DIFF
--- a/.github/workflows/test-rules.yaml
+++ b/.github/workflows/test-rules.yaml
@@ -33,7 +33,7 @@ jobs:
             # Strip the null byte added by -z
             tr '\0' '\n' |
             # Get any strings present
-            grep -Po '".*?"' |
+            grep -Po '"[^"]*"' |
             # Delete quotation marks
             tr -d '"' |
             # Convert each newline to a space


.github/workflows/test-rules.yaml
```diff
@@ -11,7 +11,45 @@ on:
...
+            exit 1
+          fi
+          
+          concat_languages=$(
+            # Find the definition of the slice
+            grep -Poz 'const DEFAULT_RULESETS_LANGUAGES: &\[&str\] = &\[\s*?(?:.*?\s*)+?\s*?\];' "$relative_path" |
+            # Strip the null byte added by -z
+            tr '\0' '\n' |
+            # Get any strings present
+            grep -Po '".*?"' |
```
amaanq, line 36: Maybe match the content by specifying anything besides a quote? That way we don't need to use the lazy operator, which would make it easier to grok

```suggestion
            grep -Po '"[^"]*"' |
```

.github/workflows/test-rules.yaml
```diff
@@ -11,7 +11,45 @@ on:
...
+      languages: ${{ steps.extract.outputs.languages }}
+    steps:
+      - uses: actions/checkout@v4
+      - id: extract
+        name: Extract languages from file
+        run: |
+          relative_path="crates/cli/src/datadog_utils.rs"
+          
+          if [ ! -f "$relative_path" ]; then
+            echo "Error: could not find target file"
```
amaanq, line 26: We could use the native error syntax to log errors, which makes them more easily visible in the workflow summary (see [here](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions#setting-an-error-message)). I also think we could log the file (in case we move it around in the future), because we'd know immediately what the path is in the workflow summary. This removes the need to diginto the workflow itself if this fails.

```suggestion
            echo "::error::Could not find file `$relative_path`"
```

.github/workflows/test-rules.yaml
```diff
@@ -11,7 +11,45 @@ on:
...
+            # Convert each newline to a space
+            tr '\n' ' ' |
+            # Strip a trailing space
+            sed 's/ $//'
+          )
+          
+          if [ -z "$concat_languages" ]; then
+          echo "Error: could not parse default ruleset languages"
+            exit 1
+          fi
```
amaanq, line 48: Similar to the above comment about setting an error message (also, indent the echo command)

```suggestion
          if [ -z "$concat_languages" ]; then
          	echo "::error::Could not parse default ruleset languages"
            exit 1
          fi
```

amaanq (approved):

PR #620, opened 2025-01-29
Split logs in stdout and stderr & also write there when logging to file

albertvaka:
When the static analyzer finishes with an error, from the IDE plugins we want to report the error logs to Datadog. 

## What is your solution?
- Send warning and error logs to `stderr` and the rest to `stdout`, so we can read `stderr` only from the IDE.
- Still print to the standard output even when logging to a file is enabled. Earlier, if logging to a file was enabled we would not log anything to the standard output. Note that logging to a file is still done in JSON format like earlier.

## Alternatives considered
I'm not sure whether the stderr/stdout writers can actually block and need the `non_blocking` wrapper (like the file writer does). I saw some open source projects wrapping them, so I also did it just in case. This means we can now have more than one guard.

### DIFF
--- a/crates/bins/src/bin/datadog_static_analyzer_server/cli.rs
+++ b/crates/bins/src/bin/datadog_static_analyzer_server/cli.rs
@@ -164,6 +164,10 @@ pub fn prepare_rocket(tx_keep_alive_error: Sender<i32>) -> Result<RocketPreparat
     let stdouterr_layer = tracing_subscriber::fmt::layer().with_writer(stdouterr);
 
     // initialize the tracing subscriber here as we're only interested in the server logs not the other CLI instructions
+    let builder = tracing_subscriber::registry()
+        .with(EnvFilter::from_default_env())
+        .with(stdouterr_layer);
+
     let guards = if let Some(log_rolling) = matches.opt_str("l") {
         // tracing with logs
         let log_dir = matches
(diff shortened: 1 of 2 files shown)


crates/bins/src/bin/datadog_static_analyzer_server/cli.rs
```diff
@@ -164,20 +174,26 @@ pub fn prepare_rocket(tx_keep_alive_error: Sender<i32>) -> Result<RocketPreparat
...
+
+        let file_layer = tracing_subscriber::fmt::layer()
+            .with_writer(non_blocking_file)
+            .json();
+
+        tracing_subscriber::registry()
+            .with(EnvFilter::from_default_env())
+            .with(stdouterr_layer)
+            .with(file_layer)
+            .try_init()
```
robertohuertasm, line 187: NIT: `init` does precisely the same as you did here with `try_init().expect`.

robertohuertasm (approved):
LGTM!

crates/bins/src/bin/datadog_static_analyzer_server/cli.rs
```diff
@@ -164,20 +174,26 @@ pub fn prepare_rocket(tx_keep_alive_error: Sender<i32>) -> Result<RocketPreparat
...
-            .with_writer(non_blocking)
-            .init();
-        Some(guard)
+        let (non_blocking_file, file_guard) = tracing_appender::non_blocking(file_appender);
+
+        let file_layer = tracing_subscriber::fmt::layer()
+            .with_writer(non_blocking_file)
+            .json();
+
+        tracing_subscriber::registry()
```
robertohuertasm, line 183: NIT: you could set the common layers outside the `if let` scope so you wouldn't have to repeat them.

robertohuertasm: For instance: 
```rust
    let builder = tracing_subscriber::registry()
        .with(EnvFilter::from_default_env())
        .with(stdouterr_layer);

// and then in the `if let` branches:
// ... if
builder.with(file_layer).init();
// ... else
builder.init();
        
```

PR #639, opened 2025-02-11
[K9VULN-3472] Enforce `RuleSet` validation upon struct creation

jasonforal:
When deserializing a `RuleSet` from the Datadog API, we apply special validation logic:

https://github.com/DataDog/datadog-static-analyzer/blob/a37a1c0200b29a823b3528b05b23898141a06751/crates/cli/src/model/datadog_api.rs#L201

https://github.com/DataDog/datadog-static-analyzer/blob/a37a1c0200b29a823b3528b05b23898141a06751/crates/cli/src/datadog_utils.rs#L233

Because not all rulesets need come from the Datadog API (e.g. they can also be passed in via a file with the CLI param `--rules rule-file.json`), the validation logic is not consistent everywhere.

## What is your solution?
Enforce validation when a `RuleSet` is created by forcing the use of `RuleSet::new`.

### DIFF
--- a/crates/static-analysis-kernel/src/model/ruleset.rs
+++ b/crates/static-analysis-kernel/src/model/ruleset.rs
@@ -1,9 +1,132 @@
 use crate::model::rule::Rule;
-use serde::{Deserialize, Serialize};
+use serde::{Deserialize, Deserializer, Serialize};
 
-#[derive(Clone, Deserialize, Debug, Serialize)]
+#[derive(Clone, Debug, Serialize)]
 pub struct RuleSet {
     pub name: String,
     pub description: Option<String>,
-    pub rules: Vec<Rule>,
+    rules: Vec<Rule>,
+    /// A private field to prevent this struct from being created manually. Use [`RuleSet::new`] instead.
+    _private: std::marker::PhantomData<()>,
+}
+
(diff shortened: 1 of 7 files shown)


crates/static-analysis-kernel/src/model/ruleset.rs
```diff
@@ -1,9 +1,132 @@
...
+    rules: Vec<Rule>,
+    /// A private field to prevent this struct from being created manually. Use [`RuleSet::new`] instead.
+    _private: std::marker::PhantomData<()>,
+}
+
+impl RuleSet {
+    /// Creates a new `RuleSet` from the provided rules, ensuring that the name of each rule
+    /// has the ruleset id prepended in the format `ruleset_id/rule_name`.
+    pub fn new(
+        ruleset_id: impl Into<String>,
```
amaanq, line 17: nit(non-blocking): References to this method pass in either a variable called `ruleset_name` or a name field of a ruleset called `ruleset.name`, and it's also used to populate the name field of this struct. For consistency, we could call this `ruleset_name` instead, and update relevant comments to use "ruleset name" instead of "ruleset id".

amaanq (approved):
approved but there's 1 non-blocking comment

PR #677, opened 2025-04-04
Handle invalid UTF-8 characters

juli1:

We want to be able to scan files with encoding different than UTF-8

## What is your solution?

Change the functions to read files so that
 - we try to read UTF-8 first
 - if we fail, we fallback and convert the invalid UTF-8 characters into artifacts

## Testing

Added integration tests with the `sast-files-encoding` [repository](https://github.com/muh-nee/sast-files-encoding) that exposes data with different encodings.


juli1:
Closed to a merge issue, using https://github.com/DataDog/datadog-static-analyzer/pull/678 instead

PR #678, opened 2025-04-04
Handle invalid UTF-8 characters

juli1:

We want to be able to scan files with encoding different than UTF-8

## What is your solution?

Change the functions to read files so that
 - we try to read UTF-8 first
 - if we fail, we fallback and convert the invalid UTF-8 characters into artifacts

## Testing

Added integration tests with the `sast-files-encoding` [repository](https://github.com/muh-nee/sast-files-encoding) that exposes data with different encodings.

### DIFF
--- a/misc/integration-test-secrets.sh
+++ b/misc/integration-test-secrets.sh
@@ -11,7 +11,7 @@ git clone --depth=1 https://github.com/muh-nee/secrets-tests.git "${REPO_DIR}"
 
 # Test without the static-analysis.datadog.yml file
 rm -f "${REPO_DIR}/static-analysis.datadog.yml"
-./target/release-dev/datadog-static-analyzer --directory "${REPO_DIR}" -o "${REPO_DIR}/results1.json" -f sarif -x --enable-secrets true --enable-static-analysis false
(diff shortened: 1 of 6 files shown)


amaanq (changes requested):
This looks like it's stacked off of #676, if so, for reviewing purposes can you please set the target branch to your other feature branch, until that PR is merged? I was very confused at first as to why a lot of the code was the same here.

juli1:
> This looks like it's stacked off of #676, if so, for reviewing purposes can you please set the target branch to your other feature branch, until that PR is merged? I was very confused at first as to why a lot of the code was the same here.

Interesting since I use a different and clean branch. Will revert the other changes for sure.

misc/integration-test-secrets.sh
```diff
@@ -50,4 +50,4 @@ fi
 
 echo "All tests passed"
 
-exit 0
+exit 0
```
amaanq, line 53: Undo please, this is not relevant.

amaanq (approved):
approved but undo the seemingly accidental newline change to `integration-test-secrets.sh`

PR #690, opened 2025-04-16
[SDS-1024] Bump dd-sds in datadog-static-analyzer

artslidd:

The next time the SCA team tries to bump the dd-sds lib, they will have a hard time understanding breaking changes.

## What is your solution?

I prefer bumping the lib myself, as I know how to solve the breaking changes.

### DIFF
--- a/crates/cli/src/sarif/sarif_utils.rs
+++ b/crates/cli/src/sarif/sarif_utils.rs
@@ -51,7 +51,7 @@ pub struct SarifReportMetadata {
 
 #[derive(Debug, Clone)]
 pub enum SarifRule {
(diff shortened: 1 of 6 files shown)


crates/cli/src/sarif/sarif_utils.rs
```diff
@@ -51,7 +51,7 @@ pub struct SarifReportMetadata {
 
 #[derive(Debug, Clone)]
 pub enum SarifRule {
-    StaticAnalysis(Rule),
+    StaticAnalysis(Box<Rule>),
```
artslidd, line 54: I did this because of 

image

fuchsnj (approved):

PR #692, opened 2025-04-21
Update to Rust 1.86

jasonforal:
If we aren't using at least [Rust 1.85](https://blog.rust-lang.org/2025/02/20/Rust-1.85.0/), we may be unable to compile some 3rd party dependencies that have adopted Rust Edition 2024.

## What is your solution?
* Update to Rust 1.86
* Run Clippy and rustfmt and apply required changes

* https://github.com/DataDog/datadog-static-analyzer/commit/456a4d000d5e43bc800afa1b46def5461c1db010 is included because Clippy's "elided_named_lifetimes" rule indicated that the lifetime is resolving to `TSQueryCursor<'a, 'a>` and needs to be named. However, that lifetime is overly-restrictive, so the actual lifetime was specified.


robertohuertasm (approved):
LGTM 🚀

juli1 (approved):

PR #776, opened 2025-09-29
[K9VULN-8370] Fetch Secret rule priority from API, and convert it to rule severity

jamesphlewis:

Severity for rules, as currently implemented, is just a base severity modified by the rule validation status. This is step one of an effort to revamp how the severity is calculated.

This fetches the priority of the rule from the secrets API, and maps the priority into a severity field, which is then propagated to the rule results.

## What is your solution?

Implementing severity based off the rule priority makes it more clear what the input to the severity of the finding actually is.


crates/cli/src/model/datadog_api.rs
```diff
@@ -352,6 +352,7 @@ pub struct SecretRuleApiAttributes {
     pub description: String,
     pub pattern: String,
     pub sds_id: String,
+    pub priority: String,
```
jamesphlewis, line 355: This is returned from the /secrets API

PR #875, opened 2026-04-14
Update workflows to get DD keys using dd-sts

gillarramendi:

- Update additional workflows to get DD keys using dd-sts
- Related [PR](https://github.com/DataDog/dd-source/pull/410158) to update sts policy and grant access to all workflows on the repository


xaldama (approved):
LGTM

PR #877, opened 2026-04-15
Detect violation enclosing method

xaldama:
SARIF violations only report file, line and column. There is no information about which method or function contains the violation, which makes it harder for consumers to attribute findings to specific code constructs.

## What is your solution?
  - Add a method_name and fully_qualified_name fields to Violation
  - Implement per-language enclosing function detection using tree-sitter AST traversal for Python, Java, Go, JavaScript, TypeScript and C#
  - Populate method_name and fully_qualified_name in execute_rule after violations are converted from JS to Rust, reusing the already-parsed tree
  - Export the method name and the fully qualified name in the SARIF output under locations[].logicalLocations with kind: "function"     

```json
{
  "locations": [
    {
      "logicalLocations": [
        {
          "fullyQualifiedName": "full name",
          "kind": "function",
          "name": "name"
        }
      ],
      "physicalLocation": {
        "artifactLocation": {
          "uri": "file_path"
        },
        "region": {
          "endColumn": x,
          "endLine": x,
          "startColumn": y,
          "startLine": y
        }
      }
    }
  ]
}
```

| Language   | Format                                                           | Example                                                                              |
|------------|------------------------------------------------------------------|--------------------------------------------------------------------------------------|
| Java       | package.ClassName#methodName(ParamType1, ParamType2):ReturnType  | org.example.DashboardController#processSimple(MultipartFile, Model):java.lang.String |
| Python     | module.ClassName.method_name                                     | controllers.dashboard.MyClass.compute                                                |
| Go         | package.TypeName.MethodName                                      | main.Server.Handle                                                                   |
| C#         | Namespace.ClassName.MethodName(ParamType1, ParamType2)           | MyApp.Controllers.Foo.Handle(string, int)                                            |
| JavaScript | ClassName.methodName                                             | MyClass.compute                                                                      |
| TypeScript | ClassName.methodName                                             | MyService.compute                                                                    |

## Alternatives considered
A generic approach using a shared list of function node types across all languages. Rejected because different languages have edge cases that require specific handling (e.g. async def in Python, constructors in Java and C#, arrow functions assigned to variables in JavaScript).

- Each language module (python/methods.rs, java/methods.rs, etc.) follows the same structure as the existing import parsers
- Two variants are exposed: find_enclosing_function (parses from scratch) and find_enclosing_function_with_tree (reuses an existing tree), consistent with the parse_imports / parse_imports_with_tree pattern
- runtime.rs uses the _with_tree variant since the tree is always available at that point
- Languages not yet covered will return None by now. They can be added incrementally

### DIFF
--- a/crates/cli/src/csv.rs
+++ b/crates/cli/src/csv.rs
@@ -85,6 +85,7 @@ mod tests {
                     fixes: vec![],
                     taint_flow: None,
                     is_suppressed: false,
+                    method_name: None,
                 }],
                 errors: vec![],
                 execution_error: None,
--- a/crates/cli/src/csv.rs
+++ b/crates/cli/src/csv.rs
@@ -85,7 +85,7 @@ mod tests {
                     fixes: vec![],
                     taint_flow: None,
                     is_suppressed: false,
-                    method_name: None,
+                    enclosing_function: None,
                 }],
                 errors: vec![],
                 execution_error: None,
--- a/crates/cli/src/file_utils.rs
+++ b/crates/cli/src/file_utils.rs
@@ -452,6 +452,7 @@ mod tests {
             fixes: vec![],
             taint_flow: None,
             is_suppressed: false,
+            method_name: None,
         };
         let directory_string = d.into_os_string().into_string().unwrap();
         let fingerprint = get_fingerprint_for_violation(
@@ -497,6 +498,7 @@ mod tests {
             fixes: vec![],
             taint_flow: Some(vec![region0, region1]),
             is_suppressed: false,
+            method_name: None,
         };
         let fingerprint = get_fingerprint_for_violation(
             "taint_flow_rule".to_string(),
@@ -528,6 +530,7 @@ mod tests {
             fixes: vec![],
             taint_flow: None,
             is_suppressed: false,
+            method_name: None,
         };
         let directory_string = d.into_os_string().into_string().unwrap();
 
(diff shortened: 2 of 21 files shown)


gillarramendi:
comments as a non rust expect 
- how can we build an integration test to ensure method name is resolved correctly? Maybe by building something like [this](https://github.com/DataDog/datadog-static-analyzer/blob/b17033b5042f8f774fc453057d480444fc4ef0a1/misc/integration-test-git.sh) to download a repo, run the  analyzer and check that sarif output contains expected method names
- [implementation is per language](https://github.com/DataDog/datadog-static-analyzer/pull/877/changes#diff-8958b1d8299c6aa0be22e35941afcae9898600572027613f5dd553cad4cd8e2dR25), so how can we ensure new languages will implement it? maybe integration test can fail if no methods found on a repo (except from a knows list of languages)

xaldama:
> comments as a non rust expect
> 
> * how can we build an integration test to ensure method name is resolved correctly? Maybe by building something like [this](https://github.com/DataDog/datadog-static-analyzer/blob/b17033b5042f8f774fc453057d480444fc4ef0a1/misc/integration-test-git.sh) to download a repo, run the  analyzer and check that sarif output contains expected method names
> * [implementation is per language](https://github.com/DataDog/datadog-static-analyzer/pull/877/changes#diff-8958b1d8299c6aa0be22e35941afcae9898600572027613f5dd553cad4cd8e2dR25), so how can we ensure new languages will implement it? maybe integration test can fail if no methods found on a repo (except from a knows list of languages)

Added

PR #878, opened 2026-04-15
[K9VULN-13253] Panic on error node fix

jdelgo:
The analyzer panics when processing Python files with invalid nodes

## What is your solution?
`parse_field_child_node` now returns `Option<MaybeAliased>` instead of panicking on unexpected node types. All 3 callers handle `None` gracefully

## Alternatives considered
An alternative was to skip files at the tree level by checking `has_error()` and `has_missing()` on the root node in `analyze.rs` but this was too broad as it would skip an entire file for any parse error, even when most of the file is valid

Added two unit tests that first assert the input has an `ERROR` node and then verifies that no panic occurs and the correct imports are returned

### DIFF
--- a/crates/static-analysis-kernel/src/analysis/languages/python/imports.rs
+++ b/crates/static-analysis-kernel/src/analysis/languages/python/imports.rs
@@ -232,14 +232,17 @@ fn parse_import_statement<'tree, 'text: 'tree>(
         // We loop so that if creating the `Import` fails, we can try any subsequent imports within
         // this node (e.g. in a comma-separated list) to avoid prematurely yielding `None` on the iterator.
         while idx < node.named_child_count() {
-            debug_assert_eq!(node.field_name_for_named_child(idx as u32), Some("name"));
+            let field_name = node.field_name_for_named_child(idx as u32);
             let name_node = node.named_child(idx).expect("should be in-bounds");
-            let parsed = parse_field_child_node(source_code, name_node);
-
             idx += 1;
 
-            if let Ok(import) = Import::try_new(parsed.full_text, parsed.alias, None, false) {
-                return Some(import);
+            if field_name != Some("name") {
+                continue;
+            }
+            if let Some(parsed) = parse_field_child_node(source_code, name_node) {
+                if let Ok(import) = Import::try_new(parsed.full_text, parsed.alias, None, false) {
+                    return Some(import);
+                }
             }
         }
         None
@@ -254,7 +257,8 @@ fn parse_import_from_statement<'text>(
     debug_assert_eq!(node.kind(), "import_from_statement");
     let module_name_node = node.child_by_field_name("module_name").expect(FIELD_EXISTS);
     let is_relative = module_name_node.kind() == "relative_import";
-    let parsed_module = parse_field_child_node(source_code, module_name_node);
+    let parsed_module = parse_field_child_node(source_code, module_name_node)
+        .ok_or_else(|| format!("invalid node type `{}`", module_name_node.kind()))?;
     // Python syntax invariant: the module of a "from" import can't have an alias.
     debug_assert!(parsed_module.alias.is_none());
     // tree-sitter grammar invariant: a valid wildcard import must end in a `wildcard_import` node.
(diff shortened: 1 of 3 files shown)


crates/static-analysis-kernel/src/analysis/languages/python/imports.rs
```diff
@@ -432,6 +436,14 @@ mod tests {
         assert!(Import::try_new("numpy.\nrandom", None, None, false).is_err());
     }
 
+    #[test]
```
bahar-shah: Have you verified the tree-sitter AST for this input to confirm an ERROR node is actually inserted? The test name says "does_not_panic" but doesn't explicitly document that this input creates an ERROR node

crates/static-analysis-kernel/src/analysis/languages/python/imports.rs
```diff
@@ -232,12 +232,16 @@ fn parse_import_statement<'tree, 'text: 'tree>(
...
         while idx < node.named_child_count() {
-            debug_assert_eq!(node.field_name_for_named_child(idx as u32), Some("name"));
+            let field_name = node.field_name_for_named_child(idx as u32);
             let name_node = node.named_child(idx).expect("should be in-bounds");
-            let parsed = parse_field_child_node(source_code, name_node);
-
             idx += 1;
 
+            // Skip nodes that aren't in the "name" field (for example, ERROR nodes).
+            if field_name != Some("name") {
```
bahar-shah: Is there any logging/telemetry in place to track when ERROR nodes are encountered? If not, consider adding a debug log or metric

crates/static-analysis-kernel/src/analysis/languages/python/imports.rs
```diff
@@ -432,6 +436,14 @@ mod tests {
         assert!(Import::try_new("numpy.\nrandom", None, None, false).is_err());
     }
 
+    #[test]
+    fn parse_error_node_does_not_panic() {
```
bahar-shah: What happens with import + , +? The test should probably include this case to ensure the function returns an empty vec gracefully

bahar-shah (dismissed):
few small comments

crates/static-analysis-kernel/src/analysis/languages/python/imports.rs
jdelgo: After speaking with Jason I'm implementing a completely new change so all of these changes have been reverted

crates/static-analysis-kernel/src/analysis/languages/python/imports.rs
jdelgo: These changes have been reverted

crates/static-analysis-kernel/src/analysis/languages/python/imports.rs
jdelgo: Change has been reverted

crates/static-analysis-kernel/src/analysis/languages/python/imports.rs
jdelgo: We decided to bring these changes back but there is no easy way to adding debugging or logging without passing around some `use_debug` variable

crates/static-analysis-kernel/src/analysis/languages/python/imports.rs
jdelgo: Updated my tests to also assert that there is an error

crates/static-analysis-kernel/src/analysis/languages/python/imports.rs
jasonforal: Please rewrite to indicate that it returns `None` in this scenario.

crates/static-analysis-kernel/src/analysis/languages/python/imports.rs
```diff
@@ -477,6 +479,30 @@ mod tests {
...
+            imports,
+            vec![
+                Import::try_new("foo", None, None, false).unwrap(),
+                Import::try_new("bar", None, None, false).unwrap(),
+            ]
+        );
+    }
+
+    #[test]
+    fn all_error_nodes_does_not_panic() {
```
jasonforal: Please rename
`parse_imports_all_error_nodes`

The fact it doesn't panic is implied by the test's success

crates/static-analysis-kernel/src/analysis/languages/python/imports.rs
```diff
@@ -477,6 +479,30 @@ mod tests {
             }
         }
     }
+
+    #[test]
+    fn error_node_does_not_panic() {
```
jasonforal: Please rename
`parse_imports_error_node`

jasonforal (dismissed):
Nice! Small nits, because the language feels a bit Claude-y

jasonforal (approved):

PR #880, opened 2026-04-20
Detect enclosing function for Java and export in SARIF

xaldama:
SARIF violations only report file, line and column. There is no information about which method or function contains the violation, which makes it harder for consumers to attribute findings to specific code constructs.

## What is your solution?
  - Add a method_name field to Violation
  - Implement Java language enclosing function detection using tree-sitter AST traversal 
  - Populate method_name in execute_rule after violations are converted from JS to Rust, reusing the already-parsed tree
  - Export the method name in the SARIF output under locations[].logicalLocations with kind: "function"     

```json
{
  "locations": [
    {
      "logicalLocations": [
        {
          "kind": "function",
          "name": "name"
        }
      ],
      "physicalLocation": {
        "artifactLocation": {
          "uri": "file_path"
        },
        "region": {
          "endColumn": x,
          "endLine": x,
          "startColumn": y,
          "startLine": y
        }
      }
    }
  ]
}
```

## Alternatives considered
A generic approach using a shared list of function node types across all languages. Rejected because different languages have edge cases that require specific handling (e.g. async def in Python, constructors in Java and C#, arrow functions assigned to variables in JavaScript).

- The implementation is per language module. In this PR Java is the implemented one. It follows the same structure as the existing import parsers
- Two variants are exposed: find_enclosing_function (parses from scratch) and find_enclosing_function_with_tree (reuses an existing tree), consistent with the parse_imports / parse_imports_with_tree pattern
- runtime.rs uses the _with_tree variant since the tree is always available at that point
- Languages not yet covered will return None by now. They can be added incrementally.
- If a new language is added and it is not included in the find_enclosing_function_with_tree list the project will not compile.


crates/static-analysis-kernel/src/analysis/languages.rs
```diff
@@ -9,6 +9,100 @@ pub mod javascript;
...
+    tree: &tree_sitter::Tree,
+    line: u32,
+    col: u32,
+    language: &Language,
+) -> Option<EnclosingFunction> {
+    match language {
+        Language::Java => {
+            java::methods::find_enclosing_function_with_tree(source_code, tree, line, col)
+        }
+        _ => None,
```
jasonforal: List each language here.

This also guards against Gotzon's concern about not adding a detection for new languages. If you didn't use `_` wildcard, it would be a compile error to not indicate the "EnclosingFunction"

crates/static-analysis-kernel/src/model/violation.rs
```diff
@@ -4,6 +4,20 @@ use common::model::position::{Position, Region};
...
+pub struct EnclosingFunction {
+    /// Simple identifier (e.g. `handle`, `doSomething`).
+    pub name: String,
+    /// Language-specific fully-qualified function or method signature. The format varies by language:
+    ///
+    /// - **Java**: `package.ClassName#methodName(ParamType1, ParamType2):ReturnType`
+    /// - **Go**: `package.FunctionName` / `package.TypeName.MethodName`
+    /// - **Python / JS / TS**: `ClassName.methodName` for methods, `functionName` for top-level functions
+    /// - **C#**: `Namespace.ClassName.MethodName(ParamType1, ParamType2)`
+    pub fully_qualified_name: String,
```
jasonforal: This isn't the "fully qualified name". It's a method signature, and I don't think it should be mixed with FQN

crates/static-analysis-kernel/src/analysis/languages.rs
```diff
@@ -9,6 +9,100 @@ pub mod javascript;
...
 pub mod typescript;
 
+use crate::model::common::Language;
+use crate::model::violation::EnclosingFunction;
+
+/// Per-file context precomputed once and reused across all violations in the same source file.
+///
+/// Construct with [`LanguageFileContext::new`] before iterating over violations, then call
+/// [`LanguageFileContext::find_enclosing_function`] for each one.
+pub enum LanguageFileContext {
```
jasonforal: This makes things a lot more complicated.

Did you benchmark the overhead of scanning once? How long does it take?

crates/static-analysis-kernel/src/analysis/languages.rs
```diff
@@ -9,6 +9,100 @@ pub mod javascript;
...
+            Language::Java => Self::Java(java::methods::JavaFileContext::new(source_code, tree)),
+            lang => Self::Other(*lang),
+        }
+    }
+
+    pub fn find_enclosing_function(
+        &self,
+        source_code: &str,
+        tree: &tree_sitter::Tree,
+        line: u32,
```
jasonforal: Please use ranges, not a single point

In theory, the two could return a different value, and range is always correct

crates/static-analysis-kernel/src/analysis/languages.rs
```diff
@@ -21,3 +115,49 @@ pub(crate) fn ts_node_text<'text>(parsed_text: &'text str, node: tree_sitter::No
...
+mod tests {
+    use crate::model::common::{Language, ALL_LANGUAGES};
+
+    // Languages with an enclosing-function implementation in this module.
+    const SUPPORTED: &[Language] = &[Language::Java];
+
+    // Languages that intentionally have no implementation yet.
+    // When adding a new language to the analyzer, add it here (no detection) or to
+    // SUPPORTED (detection implemented) — leaving it out causes this test to fail.
+    const NOT_IMPLEMENTED: &[Language] = &[
```
jasonforal: Rust's type system guards against this. You can remove this and follow what I said above about the match.

crates/static-analysis-kernel/src/model/violation.rs
```diff
@@ -42,4 +56,8 @@ pub struct Violation {
     #[serde(default)]
     #[builder(default)]
     pub is_suppressed: bool,
+    /// The function or method enclosing this violation, if any.
+    #[serde(default)]
+    #[builder(default)]
```
jasonforal, line 54: No builder please

crates/static-analysis-kernel/src/model/violation.rs
xaldama: Yes, i changed the implementation but not the name.
serde_sarif builder does not provide a method for code signature...
I am going to change it to get the FQMN instead of the method signature.

crates/static-analysis-kernel/src/analysis/languages.rs
xaldama: I am going to revert to the simplest implementation.
The difference between the 2 implementations with a big repository with 131k violations is 4 secs.
IMO not enough to justify the complexity

crates/static-analysis-kernel/src/model/violation.rs
xaldama: The field is `Option<EnclosingFunction>`. 
`#[builder(default)]` simply mirrors that behavior: it allows existing call sites to omit the field without any changes. 
Can you explain more your reasoning?
`is_suppressed` follows the same pattern above for the same reason.

jasonforal: Sorry, I glossed over this -- I initially thought you were introducing it to the struct in the first place.

misc/integration-test-method-name.sh
```diff
@@ -0,0 +1,61 @@
+#!/bin/bash
+
+# Integration test: verify that method names are populated in SARIF logicalLocations
+# for Java, which is the first language with enclosing-function detection.
+#
+# Each language block added here as support is implemented.
```
jasonforal: Would you please make this a bit more deterministic?

I understand you're modeling this implementation off the existing integration tests, but they aren't structured in the best way.

But the comment
```sh
# Each language block added here as support is implemented.
```

Suggests to me that we're blurring the lines between the integration test's purpose and the unit test.

The integration test is really just to test your [code here](https://github.com/DataDog/datadog-static-analyzer/pull/880/changes#diff-b81f2591a4aabf3dcabda3d1635b61c17a68e2583851c41d34307432629a6bb2R643-R664). The language isn't actually relevant at all, because those are covered (as you've done) with unit tests.

---

So please:
1. Create a dummy rule that flags an arbitrary but stable piece of code. For example, if you were using JS:

e.g. `file.js`
```js
function foo() {
   const a = 2;
}
```

(and then your rule will do anything that will flag something on the line `var a = 2;`)

2. Pass that rule in via `--rules`

crates/static-analysis-kernel/src/analysis/languages/java/methods.rs
```diff
@@ -0,0 +1,157 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::find_enclosing_function_with_tree;
+    use crate::analysis::tree_sitter::get_tree;
+    use crate::model::common::Language;
+    use crate::model::violation::EnclosingFunction;
+
+    fn find(source: &str, line: u32, col: u32) -> Option<EnclosingFunction> {
```
jasonforal, line 75: Small nit -- would you please change this to, instead of accept `line` and `col`, you accept a substring, and then find the substring's line in the `source` (panic if you don't find it). Then just pass `(<line>, 0)` to `find_enclosing_function_with_tree`.

It makes the test much easier to read, especially if we get to the point where we need to nest functions within functions to test corner cases or use single-line lambdas

crates/static-analysis-kernel/src/analysis/languages.rs
```diff
@@ -9,6 +9,98 @@ pub mod javascript;
...
+/// Returns the enclosing function for the given source position, reusing an already-parsed tree.
+/// See [`find_enclosing_function`] for documentation.
+pub fn find_enclosing_function_with_tree(
+    source_code: &str,
+    tree: &tree_sitter::Tree,
+    start_line: u32,
+    start_col: u32,
+    end_line: u32,
+    end_col: u32,
+    language: &Language,
```
jasonforal: Nit, please change to a value instead of a pointer. I know we use pointers elsewhere, but that never should have been the case (the value is `Copy`)

crates/static-analysis-kernel/src/analysis/languages/java/methods.rs
```diff
@@ -0,0 +1,157 @@
...
+                return Some(EnclosingFunction { name });
+            }
+            _ => {}
+        }
+        node = node.parent()?;
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
jasonforal, line 69: Should we be handling lambdas? Or if not, at the very least, add a test case that shows we didn't forget about it, we just haven't implemented it yet

misc/integration-test-method-name.sh
xaldama: done

crates/static-analysis-kernel/src/analysis/languages/java/methods.rs
xaldama: The walk-up ignores lambda expressions and continues until the method_declaration.
I added a new test to reflect this.

crates/static-analysis-kernel/src/analysis/languages/java/methods.rs
xaldama: Thanks for the suggestion! 
I would prefer to keep line numbers. 
If the same text appears twice in the source, the helper would silently pick the wrong one. 
Happy to change it if you disagree though.

misc/integration-test-method-name.sh
```diff
@@ -0,0 +1,96 @@
...
+    public void doWork() {
+        int x = 1;
+    }
+}
+EOF
+
+# ---------------------------------------------------------------------------
+# Rule: flag every local_variable_declaration.
+#
+# code (base64 of):
```
jasonforal, line 33: Just a bit nicer to do it like

```sh
code_b64=$(cat <<'EOF' | base64 | tr -d '\n'

function visit(node, filename, code) {
  const n = node.captures["decl"];
  addError(buildError(n.start.line, n.start.col, n.end.line, n.end.col, "test violation"));
}

EOF
)
```

misc/integration-test-method-name.sh
```diff
@@ -0,0 +1,96 @@
...
+#   function visit(node, filename, code) {
+#     const n = node.captures["decl"];
+#     addError(buildError(n.start.line, n.start.col, n.end.line, n.end.col,
+#                         "test violation", "WARNING", "BEST_PRACTICES"));
+#   }
+#
+# tree_sitter_query (base64 of):
+#   (local_variable_declaration) @decl
```
jasonforal, line 41: Nicer to do like
```sh
query_b64=$(echo '(local_variable_declaration) @decl' | base64)
```

jasonforal (approved):
Nice work! Left a small nit on making the integration test a bit easier to read.

Also, we're mixing the words "method" and "function" a lot: a file called `methods.rs` that contains a function called `find_enclosing_function`...an integration test named `method-name.sh`...a SARIF field named "function".

Could we just standardize on "function"? I understand "function" is inaccurate for Java (which this PR implements), but it's the de facto default for cross-language terminology, and we'll add many more languages.

crates/static-analysis-kernel/src/analysis/languages/java/methods.rs
jasonforal: Sure

PR #887, opened 2026-04-20
Create a new Config parse endpoint for the IDEs

abrooksv:
The IDEs current call `get_rulesets` but this doesn't scale to the new unified format file where we may need to extract out multiple fields from the config file. Instead we will add a new config/parse API that returns a JSON object with per-product top level keys.

## What is your solution?
* Create a new route to parse the config file that returns a structured output so we can add more products to output
* Make APIs aware of Base64 encoded String vs raw YAML String since new API does not encode the payload

Starting to implement https://docs.google.com/document/d/1PpSy12I1xMGmxOOA3qAdw05DFveUUqmETK7pnTjpnS0/

### DIFF
--- a/crates/bins/src/bin/datadog_static_analyzer_server/ide/configuration_file/endpoints.rs
+++ b/crates/bins/src/bin/datadog_static_analyzer_server/ide/configuration_file/endpoints.rs
@@ -1,12 +1,13 @@
 use std::path::PathBuf;
 
 use crate::datadog_static_analyzer_server::ide::configuration_file::models::{
-    CanOnboardRequest, GetRulesetsRequest,
+    CanOnboardRequest, GetRulesetsRequest, ParseConfigRequest, ParseConfigResponse,
+    SastParsedConfig,
 };
 
 use super::error::ConfigFileError;
 use super::models::{AddRuleSetsRequest, IgnoreRuleRequest};
-use super::static_analysis_config_file::StaticAnalysisConfigFile;
+use super::static_analysis_config_file::{Base64String, StaticAnalysisConfigFile};
 use kernel::utils::encode_base64_string;
 use rocket::http::Status;
 use rocket::response::status::Custom;
@@ -37,19 +38,20 @@ pub fn post_ignore_rule(
         ..
     } = request.into_inner();
     tracing::debug!(rule, content = &configuration_base64);
-    let result = StaticAnalysisConfigFile::with_ignored_rule(rule.into(), configuration_base64);
+    let result = StaticAnalysisConfigFile::with_ignored_rule(
+        rule.into(),
+        Base64String(configuration_base64),
+    );
     to_response_result(result, encoded)
 }
 
 /// Checks if onboarding is allowed for the configuration file (deprecated).
 ///
-/// # Deprecation
-/// Deprecated: Use [`post_can_onboard_v2`] instead.
-///
 /// # Arguments
 /// * `content` - The path to the configuration file.
 #[instrument()]
 #[rocket::get("/v1/config/can-onboard/<content..>")]
+#[deprecated(note = "IDEs stopped supporting onboarding")]
 pub fn get_can_onboard(content: PathBuf) -> Result<Json<bool>, Custom<ConfigFileError>> {
     let content_str = content.to_string_lossy().into_owned();
     can_onboard(content_str)
@@ -65,6 +67,7 @@ pub fn get_can_onboard(content: PathBuf) -> Result<Json<bool>, Custom<ConfigFile
     format = "application/json",
     data = "<request>"
 )]
+#[deprecated(note = "IDEs stopped supporting onboarding")]
 pub fn post_can_onboard_v2(
     request: Json<CanOnboardRequest>,
 ) -> Result<Json<bool>, Custom<ConfigFileError>> {
@@ -77,19 +80,23 @@ pub fn post_can_onboard_v2(
 
 /// Gets the rulesets from the static analysis configuration file (deprecated).
 ///
-/// # Deprecation
-/// Deprecated: Use [`post_get_rulesets_v2`] instead.
+/// Best-effort: returns `[]` on any parse or decode failure to preserve the wire
+/// contract for older IDE clients. Strict parsing lives in [`post_parse_config`].
 ///
 /// # Arguments
 /// * `content` - The path to the configuration file.
 #[instrument()]
 #[rocket::get("/v1/config/rulesets/<content..>")]
+#[deprecated(note = "Use post_parse_config")]
 pub fn get_get_rulesets(content: PathBuf) -> Json<Vec<String>> {
     let content_str = content.to_string_lossy().into_owned();
     get_rulesets(content_str)
 }
 
 /// Gets the rulesets from the static analysis configuration file (v2).
+///
+/// Best-effort: returns `[]` on any parse or decode failure to preserve the wire
+/// contract for older IDE clients. Strict parsing lives in [`post_parse_config`].
 ///
 /// # Arguments
 /// * `request` - The request containing the configuration file (base64).
@@ -99,6 +106,7 @@ pub fn get_get_rulesets(content: PathBuf) -> Json<Vec<String>> {
     format = "application/json",
     data = "<request>"
 )]
+#[deprecated(note = "Use post_parse_config")]
 pub fn post_get_rulesets_v2(request: Json<GetRulesetsRequest>) -> Json<Vec<String>> {
     let GetRulesetsRequest {
         configuration_base64,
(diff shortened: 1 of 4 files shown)


crates/bins/src/bin/datadog_static_analyzer_server/ide/configuration_file/static_analysis_config_file.rs
```diff
@@ -38,12 +38,32 @@ impl From<file_legacy::ConfigFile> for StaticAnalysisConfigFile {
...
+        Self::from_yaml_content(content)
+    }
+}
+
+impl StaticAnalysisConfigFile {
+    /// Parses a raw YAML configuration string (not base64-encoded) into a [`StaticAnalysisConfigFile`].
+    fn from_yaml_content(content: String) -> Result<Self, ConfigFileError> {
         use serde::de::Error;
-        let content = decode_base64_string(base64_str)?;
         if content.trim().is_empty() {
```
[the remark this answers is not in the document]
abrooksv: This is expected, we are still building support for unified format

crates/bins/src/bin/datadog_static_analyzer_server/ide/configuration_file/endpoints.rs
```diff
@@ -107,15 +115,33 @@ pub fn post_get_rulesets_v2(request: Json<GetRulesetsRequest>) -> Json<Vec<Strin
...
+/// * `request` - The request containing the raw YAML string.
+#[instrument()]
+#[rocket::post("/v2/config/parse", format = "application/json", data = "<request>")]
+pub fn post_parse_config(
+    request: Json<ParseConfigRequest>,
+) -> Result<Json<ParseConfigResponse>, Custom<ConfigFileError>> {
+    let ParseConfigRequest { configuration } = request.into_inner();
+    tracing::debug!(%configuration);
+    let config = StaticAnalysisConfigFile::try_from(configuration)
+        .map_err(|e| Custom(Status::InternalServerError, e))?;
```
albertvaka: Should probably be a 400 Bad Request, since it means the input is not valid YAML.

crates/bins/src/bin/datadog_static_analyzer_server/ide/configuration_file/endpoints.rs
```diff
@@ -107,15 +115,33 @@ pub fn post_get_rulesets_v2(request: Json<GetRulesetsRequest>) -> Json<Vec<Strin
...
+    Ok(Json(ParseConfigResponse {
+        sast: SastParsedConfig {
+            rulesets: config.sast_rulesets(),
```
albertvaka, line 133: In `get_rulesets` there's some handling to return empty array if `config.sast_rulesets()` fails. What happens in this endpoint if it fails? Since this new endpoint will unify both validation and parsing, I think it should provide some specific error in that case. If it already implicitly does somehow (I don't know rust enough to tell), it's at least lacking documentation in the comment above the method.

albertvaka (commented):
Left some comments but I think this new API is much nicer :)

crates/bins/src/bin/datadog_static_analyzer_server/ide/configuration_file/static_analysis_config_file.rs
```diff
@@ -38,12 +38,32 @@ impl From<file_legacy::ConfigFile> for StaticAnalysisConfigFile {
...
 impl TryFrom<String> for StaticAnalysisConfigFile {
     type Error = ConfigFileError;
 
-    fn try_from(base64_str: String) -> Result<Self, Self::Error> {
+    fn try_from(content: String) -> Result<Self, Self::Error> {
+        Self::from_yaml_content(content)
+    }
+}
+
+impl StaticAnalysisConfigFile {
```
robertohuertasm-datadog, line 88: NIT: There's already an impl block for StaticAnalysisConfigFile in ln 121. We could probably move this code there.

crates/bins/src/bin/datadog_static_analyzer_server/ide/configuration_file/endpoints.rs
robertohuertasm-datadog: It will return a `ConfigFileError`, in particular the `Parser` variant (see [here](https://github.com/DataDog/datadog-static-analyzer/blob/496925be240e79b77fd4ac1e4625ad83d6af3103/crates/bins/src/bin/datadog_static_analyzer_server/ide/configuration_file/static_analysis_config_file.rs#L74-L77)).

The [error type](https://github.com/DataDog/datadog-static-analyzer/blob/496925be240e79b77fd4ac1e4625ad83d6af3103/crates/bins/src/bin/datadog_static_analyzer_server/ide/configuration_file/error.rs#L8) in the signature gives you a hint, but we could definitely add that to the documentation of the method.

robertohuertasm-datadog (dismissed):
LGTM! 🚀

crates/bins/src/bin/datadog_static_analyzer_server/ide/configuration_file/static_analysis_config_file.rs
abrooksv: Opps moved

crates/bins/src/bin/datadog_static_analyzer_server/ide/configuration_file/endpoints.rs
abrooksv: Changed

crates/bins/src/bin/datadog_static_analyzer_server/ide/configuration_file/endpoints.rs
albertvaka: 👍 It still could explain the possible error responses in the comment above IMO.

albertvaka (approved):

robertohuertasm-datadog (approved):

crates/bins/src/bin/datadog_static_analyzer_server/ide/configuration_file/static_analysis_config_file.rs
```diff
@@ -99,6 +86,37 @@ fn create_ignored_rule() -> RuleConfig {
 }
 
 impl StaticAnalysisConfigFile {
+    /// Parses a raw YAML configuration string (not base64-encoded) into a [`StaticAnalysisConfigFile`].
+    fn from_yaml_content(content: String) -> Result<Self, ConfigFileError> {
```
jasonforal, line 90: I don't know if `TryFrom<String>` is the correct level of abstraction because here, we need to indicate whether the file is a legacy config or a v1 config. Otherwise, we could parse v1 syntax from a legacy file (`static-analysis.datadog.yml`)

crates/bins/src/bin/datadog_static_analyzer_server/ide/configuration_file/static_analysis_config_file.rs
```diff
@@ -99,6 +86,37 @@ fn create_ignored_rule() -> RuleConfig {
...
+        use serde::de::Error;
+        if content.trim().is_empty() {
+            return Ok(Self::default());
+        }
+        let parsed = if cfg!(test) {
```
jasonforal, line 95: Similarly, we'd likely want to remove this -- this was a guard I put in place so I could ship the v1 config in the analyzer without introducing it to the server (specifically because the IDE extension didn't have support)

jasonforal (commented):
(Note: I didn't review correctness of the new routes/request/response)

Just added a couple comments about how the parsing is done

crates/bins/src/bin/datadog_static_analyzer_server/ide/configuration_file/static_analysis_config_file.rs
abrooksv: The parser seems to internally have that data though, right? After talking with Roberto, we thought we could rely on the parser determining it based on the content shape:  https://github.com/DataDog/datadog-static-analyzer/blob/95b201c0d069affc1ece6052b7b7bfe0ccc2c2a0/crates/static-analysis-kernel/src/config/common.rs#L312

The only caveat is when we pass an empty string, we need to default that to an empty config file with unified format (which we would do in the 1.0.0 release), pre-1.0.0 that would still fall back to legacy.

Are you seeing something I am overlooking?

crates/bins/src/bin/datadog_static_analyzer_server/ide/configuration_file/static_analysis_config_file.rs
abrooksv: Yep, going to make a dedicated PR for that when we get closer to having the IDEs ready

PR #901, opened 2026-04-28
[K9VULN-14008] Fix stale v8 script cache when using `datadog-static-analyzer-server`

jasonforal:
When running `datadog-static-analyzer-server` with the `--use-rules-cache` CLI argument, there are two layers of cache, both keyed by the rule's name:
1. (Enabled by the flag): A cache for the `RuleInternal` on the `datadog-static-analyzer-server`, which exists primary to cache the [compiled tree-sitter query](https://github.com/DataDog/datadog-static-analyzer/blob/b118e6e8f82faf7265b4f8fc293fd58d56c94ed8/crates/static-analysis-kernel/src/model/rule.rs#L182).
2. (Already existing): A cache for the precompiled v8 script (Javascript code) on the `JsRuntime`.

The cache is designed such that the caller is expected to to manage the `JsRuntime`'s v8 script cache. Currently, the `datadog-static-analyzer-server` explicitly evicts the v8 script cache for a specific rule when the incoming JSON payload represents a separate rule:

https://github.com/DataDog/datadog-static-analyzer/blob/b118e6e8f82faf7265b4f8fc293fd58d56c94ed8/crates/bins/src/bin/datadog_static_analyzer_server/rule_cache.rs#L99-L102

This scenario essentially only occurs when working on a custom rule (where each edit to the JavaScript or tree-sitter query sends a brand new rule).

The issue is that there are multiple `JsRuntime` -- one for each thread:
https://github.com/DataDog/datadog-static-analyzer/blob/b118e6e8f82faf7265b4f8fc293fd58d56c94ed8/crates/bins/src/bin/datadog_static_analyzer_server/endpoints.rs#L128-L131

But there is only a single `RuleInternal` cache, shared across all threads:

https://github.com/DataDog/datadog-static-analyzer/blob/b118e6e8f82faf7265b4f8fc293fd58d56c94ed8/crates/bins/src/bin/datadog-static-analyzer-server.rs#L14

The issue is that while the `is_update` boolean works correctly, it is only ever triggered once. Because the `cached_analysis_request` accepts the `JsRuntime` [as an argument](https://github.com/DataDog/datadog-static-analyzer/blob/b118e6e8f82faf7265b4f8fc293fd58d56c94ed8/crates/bins/src/bin/datadog_static_analyzer_server/rule_cache.rs#L83-L84), this means that only a single runtime's v8 script cache is cleared. Thus, if the server is run with multiple threads, all other threads will continue to have a stale cache.

## What is your solution?
* Track the hash of the rule code associated with the v8 script within the `JsRuntime` itself. Within `datadog-static-analyzer-server`, always hash and check equivalence of rule code.
    * Note: this does introduce overhead for the overwhelming majority of cases (i.e. normal IDE usage where rules don't change), though sha256 hashing is generally hardware-accelerated and should have minimal impact.

## Alternatives considered
* Introduce a query param to the server's [`/analyze` endpoint](https://github.com/DataDog/datadog-static-analyzer/blob/b118e6e8f82faf7265b4f8fc293fd58d56c94ed8/crates/bins/src/bin/datadog_static_analyzer_server/endpoints.rs#L115-L120) and thread this through to `cached_analysis_request`, such that the server would only hash and check the v8 script cache if that query param is present. Then, the VS Code extension would modify only the [custom rule editor](https://docs.datadoghq.com/security/code_security/static_analysis/custom_rules/#write-rules-with-the-vs-code-extension) to pass this query param. This would be a "correct" and surgical implementation that avoids unnecessary hashing.
   * This would require coordination across multiple codebases. I'd rather only resort to this if benchmarks show this PR introduces a noticeable regression. I don't expect it will.

* This bug only really surfaces in the VS Code extension's custom rule editor and thus doesn't affect correctness of standard IDE usage.
* Beyond multiple runtimes being a blind spot, this issue also slipped by because there was no actual test coverage on the v8 script cache (despite a test comment misleading saying there was). This has been addressed with [this update](https://github.com/DataDog/datadog-static-analyzer/commit/d2fbb08059f52c2af04d6c269fc671a3bd5407cd#diff-89b524c584a90856b90882d46e11761569ebf4b573efd8ca84b2d33e66be931dR239-R285).

### DIFF
--- a/crates/bins/src/bin/datadog_static_analyzer_server/rule_cache.rs
+++ b/crates/bins/src/bin/datadog_static_analyzer_server/rule_cache.rs
@@ -154,7 +154,7 @@ def abc():
         };
         AnalysisRequest {
             filename: "file.py".to_string(),
-            language: Language::Python,
+            language,
             file_encoding: "utf-8".to_string(),
             code_base64: encode_base64_string(file_contents.to_string()),
             rules: vec![server_rule],
--- a/crates/bins/src/bin/datadog_static_analyzer_server/rule_cache.rs
+++ b/crates/bins/src/bin/datadog_static_analyzer_server/rule_cache.rs
@@ -16,11 +16,6 @@ pub struct CompiledServerRule {
 }
 
 impl CompiledServerRule {
-    /// Returns a reference to the [`ServerRule`] that this cache represents.
-    pub fn inner(&self) -> &ServerRule {
-        &self.rule
-    }
-
     /// Returns a reference to the [`RuleInternal`] that this cache represents.
     pub fn as_internal(&self) -> &RuleInternal {
         &self.internal
@@ -32,23 +27,17 @@ impl CompiledServerRule {
 #[derive(Debug)]
 pub struct RuleCache(DashMap<String, Arc<CompiledServerRule>>);
 
-/// A return value from looking up or inserting a value into a [`RuleCache`]. `is_update` will be `true`
-/// if this value previously existed in the map and was updated, or `false` if either a
-/// previous value was returned or the rule name has been cached for the first time.
-#[derive(Debug)]
-pub(crate) struct CachedValue {
-    pub(crate) rule: Arc<CompiledServerRule>,
-    pub(crate) is_update: bool,
-}
-
 impl RuleCache {
     /// Returns a new `RuleCache` with a capacity of 0.
     pub fn new() -> Self {
         Self(DashMap::new())
     }
 
-    /// Returns a [`CachedValue`] for the provided [`ServerRule`], utilizing a cache under the hood.
-    pub fn get_or_insert_from(&self, server_rule: ServerRule) -> Result<CachedValue, &'static str> {
+    /// Returns the [`CompiledServerRule`] for the provided [`ServerRule`], utilizing a cache under the hood.
+    pub fn get_or_insert_from(
+        &self,
+        server_rule: ServerRule,
+    ) -> Result<Arc<CompiledServerRule>, &'static str> {
         if let Some(cached) = self.0.get(&server_rule.name) {
             // (Note that even if the cache has this rule re-written after this function returns, the caller
             // will still hold an owned reference to the original compiled rule, so there is no race).
@@ -57,10 +46,7 @@ impl RuleCache {
             // This check is necessary because the HashMap is keyed by rule name, so there
             // could be a collision if the rule is updated.
             if cached.rule == server_rule {
-                return Ok(CachedValue {
-                    rule: Arc::clone(cached.value()),
-                    is_update: false,
-                });
+                return Ok(Arc::clone(cached.value()));
             }
         }
         // If there was no cached rule (or the rule wasn't an exact match), compile and cache a new rule.
@@ -71,11 +57,8 @@ impl RuleCache {
             internal: rule_internal,
         });
         let cloned_rule = Arc::clone(&compiled);
-        let existing = self.0.insert(rule_name, compiled);
-        Ok(CachedValue {
-            rule: cloned_rule,
-            is_update: existing.is_some(),
-        })
+        self.0.insert(rule_name, compiled);
+        Ok(cloned_rule)
     }
 }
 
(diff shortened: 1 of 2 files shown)


crates/static-analysis-kernel/src/analysis/ddsa_lib/runtime.rs
```diff
@@ -214,13 +214,21 @@ impl JsRuntime {
...
-    /// Clears the cache for the given rule name, returning `true` if a script
-    /// existed and was removed from the cache, or `false` if it didn't exist.
+    /// Drops the cached [v8::UnboundScript] for the given `rule_name` if its `rule_code` doesn't match the provided.
+    /// Returns `true` if an entry was evicted, or `false` if not.
     ///
     /// # Panics
     /// Panics if the `rule_cache` has an existing borrow.
-    pub fn clear_rule_cache(&self, rule_name: &str) -> bool {
-        self.rule_cache.borrow_mut().remove(rule_name).is_some()
+    pub fn evict_script_if_stale(&self, rule_name: &str, rule_code: &str) -> bool {
```
bahar-shah, line 222: This could avoid holding a mutable borrow when no eviction is needed:

  ```rust
  pub fn evict_script_if_stale(&self, rule_name: &str, rule_code: &str) -> bool {
      // First check with read-only access
      if let Some(entry) = self.rule_cache.borrow().get(rule_name) {
          if entry.code_hash == CompiledRule::get_code_hash(rule_code) {
              return false; // No eviction needed, never took mut borrow
          }
      } else {
          return false;
      }
      // Only take mut borrow if eviction is actually needed
      self.rule_cache.borrow_mut().remove(rule_name).is_some()
  }
```
  This reduces lock contention in the common case where rules haven't changed.

File: crates/static-analysis-kernel/src/analysis/ddsa_lib/runtime.rs [hunk 2]
```diff
@@ -214,13 +214,21 @@ impl JsRuntime {
...
+        let mut cache = self.rule_cache.borrow_mut();
+        let Some(entry) = cache.get(rule_name) else {
+            return false;
+        };
+        if entry.code_hash != CompiledRule::get_code_hash(rule_code) {
```
bahar-shah, line 227: get_code_hash(rule_code) is called on every request, even for cache hits. not sure if this is a concern from a performance perspective or if the overhead here is negligible. You could otherwise precompute it in RuleInternal during construction and pass it in but that might be overkill tbh.

crates/bins/src/bin/datadog_static_analyzer_server/rule_cache.rs
```diff
@@ -96,11 +79,9 @@ pub(crate) fn cached_analysis_request(
             // we know the runtime's v8::Script cache cannot change out from under us.
             // Thus, there is no potential race in the time after clearing the cache
             // but before actually executing the JavaScript rule).
-            if cached.is_update {
-                // The runtime's `v8::Script` cache is keyed only by name -- we need to manually clear it.
-                runtime.clear_rule_cache(&cached.rule.inner().name);
-            }
-            compiled_rules.push(cached.rule);
+            let internal = cached.as_internal();
+            let _ = runtime.evict_script_if_stale(&internal.name, &internal.code);
```
bahar-shah, line 83: nit- the return value isn't used here. can we remove it?

bahar-shah (approved):

crates/static-analysis-kernel/src/analysis/ddsa_lib/runtime.rs
jasonforal: There will never be lock contention, otherwise this would actually panic. We're using an `Rc<RefCell<_>>` here, not a mutex. It's governed by runtime borrow checking.

crates/static-analysis-kernel/src/analysis/ddsa_lib/runtime.rs
jasonforal: This is mentioned in the description

crates/bins/src/bin/datadog_static_analyzer_server/rule_cache.rs
jasonforal: It's essentially a style question in this case. I just prefer to be explicit

PR #914, opened 2026-05-17
fix(kernel): emit 1-based UTF-16 code-unit columns from tree-sitter nodes

robertohuertasm-datadog:
## Summary

This PR fixes a long-standing semantic bug in how the static analyzer reports column numbers. Tree-sitter's `Point.column` is a **0-based UTF-8 byte offset**, but our internal `Position.col` field was simply doing `byte_col + 1` and calling it a column number. This produces correct results for ASCII-only source files, but silently emits wrong columns for any source containing multibyte characters (emoji, CJK ideographs, accented letters, combining marks, etc.).

LSP, VS Code, and SARIF v2.1 all define columns as **1-based UTF-16 code units**. This PR aligns `Position.col` with that standard.

### Errors

Playground:

image

VS Code: 

image

---

### Core conversion helper (`crates/common`)

Introduced `LineColumnIndex` in `crates/common/src/utils/position_utils.rs`, a thin newtype wrapper around the [`line-index`](https://crates.io/crates/line-index) crate (the line/column indexing engine extracted from rust-analyzer):

- Built once per source string via `LineIndex::new`.
- Exposes `byte_col_to_utf16_col(row, byte_col) -> u32` which delegates to `LineIndex::to_wide(WideEncoding::Utf16, ..)` and applies the `+ 1` adjustment for 1-based columns.
- `line-index` splits on `\n` only, which mirrors tree-sitter's line model exactly. Tree-sitter does not treat bare `\r` as a line terminator, and for `\r\n` files the `\r` is part of the column count on the same line.
- `RootContext` owns an `Option<LineColumnIndex>` directly, populated once in `set_text`; ops borrow `&LineColumnIndex` per call (no re-scanning, no per-call `Vec<usize>` clones).

> [!NOTE]
> Earlier revisions of this PR shipped a hand-rolled implementation with an ASCII `is_ascii()` fast path and a per-scalar `char::len_utf16()` slow path. Following [reviewer feedback](https://github.com/DataDog/datadog-vscode/pull/2418#discussion_r3225316557), commit [8947d3bb](https://github.com/DataDog/datadog-static-analyzer/commit/8947d3bbda2decb494140adb628ce37085abcf24) replaced it with the `line-index` wrapper to lean on a battle-tested upstream implementation. All existing unit tests (ASCII, BMP non-ASCII, supplementary-plane emoji, CJK, combining marks, CRLF, EOL/empty lines) pass unchanged against the new backing.

### Contract documentation (`crates/common`)

`Position.col` is now documented as a **1-based UTF-16 code-unit column** in `crates/common/src/model/position.rs`, locking in the semantics for all future contributors.

### Kernel — tree-sitter layer (`crates/static-analysis-kernel`)

- `map_node` and `get_query_nodes` in `tree_sitter.rs` now accept a `&LineColumnIndex` parameter and use it instead of the raw `+ 1` offset.
- `LineColumnIndex` is built once per `get_query_nodes` call (one scan of the source), then reused for every node in that query result.

### Kernel — ddsa JavaScript bridge

The `LineColumnIndex` is threaded through the full ddsa call chain that feeds tree-sitter nodes into V8:

```
JsRuntime::execute_rule_internal
  └─ QueryMatchBridge::set_data
       └─ TsNodeBridge::insert_capture / insert / build_v8_node
            └─ TreeSitterNode::from_ts_node_with_index   ← single source of truth
```

- `TreeSitterNode::from_ts_node` (the old ASCII-only constructor) has been **removed**. There is now a single constructor, `from_ts_node_with_index`, which always takes a `&LineColumnIndex`. This eliminates the footgun of accidentally using the byte-based constructor on a multibyte source.
- The deno ops `op_ts_node_named_children` and `op_ts_node_parent` build a `LineColumnIndex` from the source text stored in `RootContext` before inserting newly discovered nodes into the bridge. The `ctx_bridge` borrow in `op_ts_node_named_children` was moved into the `else` branch (when there are actually children to process) to keep the borrow scope tight and the intent clear.

### Kernel — taint-flow region builders

`LocatedNode::new_cst` in `flow/graph.rs` (used by the taint-flow graph engine) previously did `node.start_position().column + 1`. It now accepts the full source string, builds a `LineColumnIndex`, and emits the same UTF-16 column as `from_ts_node_with_index` — ensuring the taint-flow graph deduplication key stays consistent with the rest of the system.

The `position_eq` test helper in `flow/java.rs` was updated to compare against `LineColumnIndex`-derived columns rather than raw byte offsets.

### Server (`crates/static-analysis-server`)

`process_tree_sitter_tree_request` now builds a `LineColumnIndex` from the decoded source before calling `map_node`, so IDE AST responses carry UTF-16 columns.

### SARIF and CSV output (`crates/cli`)

No changes were required to the serialization logic — SARIF and CSV both read `Position.col` directly, so fixing the kernel's output automatically fixes their output. Focused tests were added to assert that the values pass through unchanged.

---

## Why UTF-16?

- **SARIF v2.1** specifies column numbers as 1-based UTF-16 code units (the format's default encoding is UTF-16).
- **LSP / VS Code** use 0-based UTF-16 code units; our 1-based variant is the natural complement.
- **CJK characters** (e.g. 日, 本) are 3 UTF-8 bytes but 1 UTF-16 unit — they are common in production code and the byte-based columns were visibly wrong.
- **Emoji** (e.g. 🚀) are 4 UTF-8 bytes but 2 UTF-16 units (surrogate pair) — even more extreme divergence.
- For **pure ASCII** (the vast majority of lines), byte col == UTF-16 col, so there is zero behavioral change on ASCII-only files.

---

## Blast radius

| Area | Impact |
|------|--------|
| ASCII-only source files | **No change** — byte col == UTF-16 col for ASCII |
| Source files with CJK, accented characters | Column numbers decrease (fewer UTF-16 units than UTF-8 bytes) |
| Source files with emoji (surrogate pairs) | Column numbers decrease further (4 bytes → 2 UTF-16 units) |
| SARIF output | Now conforms to SARIF v2.1 UTF-16 column semantics |
| CSV output | Faithfully serializes whatever `Position.col` contains (now UTF-16) |
| IDE / server AST responses | Column numbers updated to UTF-16 |
| Taint-flow graph deduplication | Consistent — all paths use the same `LineColumnIndex` helper |
| Secrets scanner (`get_position_in_string`) | **Not affected** — uses a separate grapheme-based column contract that does not flow through the kernel output |

The change is intentionally a **semantic breaking change** for columns on non-ASCII input. Any downstream consumer (e.g. suppression rules keyed on exact column numbers for non-ASCII code) will need to update their column values. For ASCII code — which is the overwhelming majority of suppression use cases — nothing changes.

---

## What we considered and rejected

**Storing `LineColumnIndex` in the bridge instead of passing it as a parameter**
Rejected because the source text lives in `RootContext` which is reset per file, and the bridge is long-lived across files. Storing the index in the bridge would require either cloning the source string (expensive) or carefully managing a lifetime that would become dangling. Passing `&LineColumnIndex` as a parameter keeps lifetimes simple and makes the dependency explicit.

**Keeping `TreeSitterNode::from_ts_node` (the old byte-based constructor) for ASCII tests**
Rejected because having two constructors with different semantics is a footgun — a future test author could reach for the shorter name on a multibyte source and get silently wrong columns. The ASCII tests were trivially updated to use `from_ts_node_with_index` with a `LineColumnIndex::new` call; for pure ASCII sources the results are identical.

**Using `str::lines()` for line splitting in `LineColumnIndex`**
Rejected because `str::lines()` also splits on bare `\r` (old Mac OS 9 line endings), which tree-sitter does not recognise as a line terminator. Diverging here would produce wrong row/column pairs on `\r\n` files. The `\n`-only scan is intentional and documented at the call site.

**Using `usize` for column numbers everywhere**
`byte_col_to_utf16_col` returns `u32` to match `TreeSitterNode`'s field types and `Position.col`. The `LocatedNode::Cst.col` field uses `usize` (a pre-existing inconsistency); the cast at the boundary is harmless but noted as a mild code smell if a future cleanup is desired.

---

## Tests added

- **`LineColumnIndex` unit matrix** — ASCII fast path, BMP non-ASCII (`é`), supplementary-plane emoji (🚀), CJK (日本), combining marks, CRLF line endings, EOL boundaries, empty lines.
- **`map_node` multibyte** — emoji prefix shifts `num` identifier to UTF-16 col 11 instead of raw byte col 13.
- **`get_query_nodes` CJK** — 日 (3 UTF-8 bytes, 1 UTF-16 unit) shifts `end` identifier to UTF-16 col 8.
- **`TsNodeBridge` multibyte** — v8 `_startCol` field is 7 (UTF-16) for `abc` after `"🚀"; `.
- **`TreeSitterNode::from_ts_node_with_index` multibyte** — emoji and CJK prefix tests.
- **`LocatedNode::new_cst` regression** — asserts UTF-16 col differs from raw byte col + 1 on multibyte input and agrees on ASCII input.
- **`stella_compat::getCode` multibyte** — end-to-end JS execution test; `violation.base_region.start_col` is 7 (UTF-16), not 9 (byte).
- **Server integration test** — base64-encoded `x = "🚀"; num = 5` parsed as Python; asserts a node with `start.col == 11` exists and no node reports `start.col == 13`.
- **SARIF serialization** — `startColumn`/`endColumn` pass through unchanged.
- **CSV serialization** — UTF-16 col 3 appears in the CSV row, not raw byte col.

## Notes for the reviewers

See https://github.com/DataDog/datadog-static-analyzer/pull/914#issuecomment-4471186747


crates/static-analysis-kernel/src/analysis/ddsa_lib/ops.rs
```diff
@@ -104,6 +105,12 @@ pub fn op_ts_node_named_children<'s>(
         None
     } else {
         let array = v8::Array::new(scope, count as i32);
+        // ctx_bridge is only needed in this branch. Build the index while ctx_borrow is alive;
+        // ts_node_bridge and ctx_bridge are separate RefCells so the concurrent borrows are valid.
+        let ctx_bridge = state.borrow::<Rc<RefCell<bridge::ContextBridge>>>();
+        let ctx_borrow = ctx_bridge.borrow();
+        let source = ctx_borrow.ddsa_root_context().get_text().unwrap_or("");
+        let idx = LineColumnIndex::new(source);
```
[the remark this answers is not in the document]
robertohuertasm-datadog: Good catch. Quick analysis: `LineColumnIndex::new` is a single linear \ scan (memchr-level perf). On a 10 KB file that's ~10 µs; on 100 KB, ~100 µs. Worst realistic case (rule doing a 500-node walk on 100 KB) burns ~50 ms; typical case is sub-millisecond. The constants are small enough that it hadn't appeared in any flamegraph yet, but caching is cheap and eliminates the worst-case asymptotic — the same pattern the existing `parent_map` cache uses. Fixed in [76571835](https://github.com/DataDog/datadog-static-analyzer/commit/76571835): `RootContext` now stores pre-computed `line_starts: Vec<usize>` (built once in `set_text`), and `line_column_index()` builds a `LineColumnIndex` from the cached offsets in O(1).

crates/static-analysis-kernel/src/analysis/ddsa_lib/ops.rs
```diff
@@ -152,14 +159,16 @@ pub fn op_ts_node_parent(
...
-    let ctx_bridge = ctx_bridge.borrow_mut();
-    let root_ctx = ctx_bridge.ddsa_root_context();
+    let ctx_borrow = ctx_bridge.borrow();
+    let root_ctx = ctx_borrow.ddsa_root_context();
     let safe_raw_parent =
         OpSafeRawTSNode::from_root_context(root_ctx, |ctx| ctx.get_ts_node_parent(ts_node))?;
     let parent_ts_node = safe_raw_parent.to_node();
 
+    let source = root_ctx.get_text().unwrap_or("");
+    let idx = LineColumnIndex::new(source);
```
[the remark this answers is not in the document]
robertohuertasm-datadog: Same fix as above — fixed in [76571835](https://github.com/DataDog/datadog-static-analyzer/commit/76571835f5efd68d91f31f334c1238836bb75364). `op_ts_node_parent` now calls `root_ctx.line_column_index()` which uses the pre-computed offsets cached in `RootContext::set_text`, so repeated parent walks stay O(depth) without the O(source_len) scan on each hop.

crates/static-analysis-kernel/src/analysis/ddsa_lib/js/flow/graph.rs
```diff
@@ -621,11 +622,17 @@ impl From<LocatedEdge<'_>> for dot_structures::Edge {
 
 impl<'a> LocatedNode<'a> {
     /// Constructs a new `LocatedNode` from a tree-sitter node.
-    pub fn new_cst(node: tree_sitter::Node, text: &'a str) -> LocatedNode<'a> {
+    ///
+    /// `source` is the full source text that was parsed to produce `node`; it is used via
+    /// [`LineColumnIndex`] to emit a 1-based UTF-16 code-unit column consistent with
+    /// [`TreeSitterNode::from_ts_node_with_index`].
+    pub fn new_cst(node: tree_sitter::Node, text: &'a str, source: &str) -> LocatedNode<'a> {
+        let idx = LineColumnIndex::new(source);
```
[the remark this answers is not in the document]
robertohuertasm-datadog: Fixed in [a719e858](https://github.com/DataDog/datadog-static-analyzer/commit/a719e85877a376f6a3d78ef7dfcaa5269f9c699f). You're right — building a LineColumnIndex per vertex was wasteful regardless of constants, and the refactored API (passing `&LineColumnIndex`) makes the dependency explicit and lets the caller decide where to build/cache it. `op_digraph_adjacency_list_to_dot` now builds the index once before the closure (reusing the `RootContext`-cached offsets when available); `cst_dot_digraph` in `graph_test_utils.rs` does the same before the preorder iterator.

crates/static-analysis-kernel/src/analysis/tree_sitter.rs
```diff
@@ -745,6 +758,98 @@ SELECT * FROM table WHERE column = 'value';
...
+    ///   x(0) ' '(1) =(2) ' '(3) "(4) 🚀(5..8) "(9) ;(10) ' '(11) n(12) u(13) m(14) ' '(15)
+    /// UTF-16 layout:
+    ///   x(1) ' '(2) =(3) ' '(4) "(5) 🚀(6,7) "(8) ;(9) ' '(10) n(11) ...
+    ///   → `num` starts at UTF-16 col 11 (0-based byte 11, but 🚀 takes 2 UTF-16 units → 11-2+1 = 10?)
+    ///
+    /// Let's compute carefully:
+    ///   bytes 0-10: `x = "🚀";` (11 bytes: x, ' ', =, ' ', ", 🚀×4, ", ;)
+    ///   byte 11 (space), byte 12..14 = 'n','u','m'
+    ///   UTF-16 prefix to byte 12: x(1),' '(1),=(1),' '(1),"(1),🚀(2),"(1),;(1),' '(1) = 10 units
+    ///   → num starts at UTF-16 col 11 (10 units + 1-based)
```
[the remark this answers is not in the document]
robertohuertasm-datadog: Fixed in [ef4527b8](https://github.com/DataDog/datadog-static-analyzer/commit/ef4527b880192b61fffe05638bedcf8c050c82b3). Removed the speculative '10?' line — the comment now has a single step-by-step UTF-16 prefix sum that reads straight to the `assert_eq!(... 11)` below it.

crates/static-analysis-kernel/src/analysis/tree_sitter.rs
```diff
@@ -745,6 +758,98 @@ SELECT * FROM table WHERE column = 'value';
...
+    /// (3 UTF-8 bytes, 1 UTF-16 code unit each).
+    ///
+    /// Source (Python): `日 = 1; end = 2`
+    /// 日 = 3 UTF-8 bytes, 1 UTF-16 unit.
+    /// `end` starts at byte offset 7 (`日`(3) + ' '(1) + '='(1) + ' '(1) + '1'(1) = 7),
+    /// but UTF-16 units before it: 日(1),' '(1),=(1),' '(1),1(1),;(1),' '(1) = 7 → col 8.
+    /// Since 日 is BMP (1 UTF-16 unit = 1 byte col unit at the codepoint level, but 3 UTF-8
+    /// bytes), we have: 3 bytes - 1 UTF-16 = 2 byte savings on that char, so byte 7 → UTF-16 col
+    /// = 7 - 2 + 1 = 6.  Let's recompute step by step:
+    ///   日(3 bytes → 1 UTF-16), ' '(1→1), =(1→1), ' '(1→1), 1(1→1), ;(1→1), ' '(1→1) = 7 UTF-16 units → col 8
```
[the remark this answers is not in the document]
robertohuertasm-datadog: Fixed in [0243860e](https://github.com/DataDog/datadog-static-analyzer/commit/0243860ef8a2d1af53900c81bb39bce767a62f9a). Removed the first (wrong) byte-7 derivation and replaced the 'let's recompute' tangent with a single clean calculation: 日(3)+' '(1)+=(1)+' '(1)+1(1)+;(1)+' '(1) = 9 bytes → 7 UTF-16 units → 1-based col 8.

crates/static-analysis-kernel/src/analysis/ddsa_lib/runtime.rs
```diff
@@ -1446,6 +1448,44 @@ function visit(query, filename, code) {
         assert_eq!(*violation, expected);
     }
 
+    /// `stella_compat.js::getCode` correctly extracts an identifier that follows a multibyte
+    /// character on the same line.
+    ///
+    /// The source `"🚀protectedName = 1;"` has the emoji (4 UTF-8 bytes / 2 UTF-16 code units)
+    /// before the identifier.  With UTF-16 columns, `col - 1` correctly indexes into the JS
+    /// string via `substring`, so `getCodeForNode(node, code)` returns `"protectedName"`.
```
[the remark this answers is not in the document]
robertohuertasm-datadog: Fixed in [60e4f955](https://github.com/DataDog/datadog-static-analyzer/commit/60e4f955142b354a2106af83628a3427e24a7fb3). The header now accurately describes the actual source layout: the emoji is inside a string literal (`"\u{1F680}";`), not adjacent to `protectedName`. Also updated the byte-offset and UTF-16 derivation to match the real token positions.

crates/common/src/model/position.rs
```diff
@@ -6,6 +6,18 @@ use derive_builder::Builder;
...
+/// * `line` — 1-based line number.
+/// * `col` — 1-based **UTF-16 code-unit** column number for every position produced by the
+///   static-analysis kernel (tree-sitter path).  On ASCII-only lines, one UTF-16 code unit equals
+///   one byte, so the value is identical to what a byte-column would be.  For non-ASCII characters
+///   (CJK ideographs, emoji surrogate pairs, combining marks, etc.) the value reflects UTF-16
+///   semantics, which matches LSP / VS Code and the SARIF v2.1 default encoding.
+///
+///   The `get_position_in_string` helper (used by the secrets scanner) emits grapheme-based
+///   columns under a different contract; that path does **not** flow through `Position.col` in the
+///   kernel output.
```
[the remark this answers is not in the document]
robertohuertasm-datadog: Fixed in [04e6d1fe](https://github.com/DataDog/datadog-static-analyzer/commit/04e6d1fe3c918e152e2fe8346e6ddcfff6777291). The docstring now leads with a producer-agnostic 'the unit depends on the producer' statement and lists both contracts (kernel = UTF-16 code units, secrets scanner = Unicode grapheme clusters) as equal-weight bullets. No more risk of a consumer reading just the headline and assuming UTF-16 everywhere.

crates/common/src/utils/position_utils.rs
```diff
@@ -2,6 +2,70 @@ use crate::model::position::Position;
...
+        let prefix = &source_from_line_start[..prefix_len];
+
+        // ASCII fast path: one byte == one UTF-16 code unit.
+        if prefix.is_ascii() {
+            return (byte_col + 1) as u32;
+        }
+
+        // Slow path: count UTF-16 code units emitted by each Unicode scalar.
+        let prefix_str = std::str::from_utf8(prefix).unwrap_or("");
+        let utf16_col: usize = prefix_str.chars().map(|c| c.len_utf16()).sum();
```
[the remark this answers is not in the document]
robertohuertasm-datadog: Good observation — the slow path does scan the non-ASCII prefix twice (once for `is_ascii()`, once for `chars().map(len_utf16).sum()`). A few reasons we're deferring this for now:

1. **Scope is bounded**: the double scan covers at most one line up to `byte_col` bytes, not the whole file. The asymptotic is O(line_len), not O(file_len).
2. **ASCII fast path absorbs almost all calls**: real-world source files are overwhelmingly ASCII on any given line. The `is_ascii()` check is a single SIMD-friendly slice scan that short-circuits before the slow path is ever reached.
3. **Index is now cached (PR items 1 & 2)**: with `RootContext::line_column_index()` building the index in O(1) per op call, the per-conversion cost is what's left — and we have no profiling evidence that the slow path is a hotspot.

The natural next step if this shows up in a flamegraph would be pre-computing per-line UTF-16 checkpoints (or non-ASCII byte offsets) during index construction so each conversion is O(1) regardless of the line content. Happy to revisit once we have profiling data.

crates/static-analysis-kernel/src/analysis/ddsa_lib/ops.rs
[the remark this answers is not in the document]
robertohuertasm-datadog: Follow-up in d39db47e: the `unwrap_or_else(|| LineColumnIndex::new(""))` fallback was a silent failure — if `set_text` were ever skipped, every node would get col=1 with no visible error. Replaced with `.expect("tree text must be set before ops run")` in both `op_ts_node_named_children` and `op_ts_node_parent`, consistent with the panic already present in `op_digraph_adjacency_list_to_dot` for the same invariant.

crates/static-analysis-kernel/src/analysis/ddsa_lib/ops.rs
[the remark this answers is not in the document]
robertohuertasm-datadog: Same follow-up as above (d39db47e): replaced `unwrap_or_else` with `.expect()` for consistency. Silent wrong columns are harder to debug than an explicit panic with a clear message.

robertohuertasm-datadog:
### Heads-up for reviewers — temporary loosening of `check-regressions.js` in 329df0cc

> **Update**: the follow-up that restores the column-aware comparison is now open as [#915](https://github.com/DataDog/datadog-static-analyzer/pull/915), stacked on top of this branch. It will close the loop the moment this PR lands on `main`.

**TL;DR**: this PR also includes a one-shot, intentional weakening of the regression workflow (`startColumn` / `endColumn` removed from the comparison key). #915 reverts that part immediately after this lands on `main`.

**Why was the regression CI failing?**

The workflow runs the analyzer twice — once on `main`, once on the PR branch — and compares the two SARIF outputs as JSON strings (`set1` vs `set2` in `check-regressions.js`). This PR changes the kernel from emitting 1-based UTF-8 byte columns to 1-based UTF-16 code-unit columns. While the PR is in review:

- `main` build  → byte columns
- branch build → UTF-16 columns

Every violation that lives on a line containing a non-ASCII character drifts by N columns (N = number of multibyte chars before the violation position). For example, on `numpy/_core/tests/test_strings.py` line 1284 (`"λ" * 5 + "μ" * 2`):

| | startColumn | endColumn |
|---|---|---|
| main (bytes)   | 14 | 32 |
| branch (UTF-16)| 14 | 31 |

The string-equality check sees this as "violation X removed, new violation Y added" — even though it's the **same violation**, only the column number changed. No rule is firing in a new place, no rule has stopped firing.

**Reproduced locally on `_core/tests/test_strings.py`:**

- Without the fix: 3 "removed" + 3 "added" false positives (lines with `λ`, `μ`, `·`)
- With the fix: 0 + 0 (clean)

**What 329df0cc does**

1. `parseFile` strips `startColumn` / `endColumn` from the in-memory region — the comparison key becomes `(file, ruleId, message, startLine, endLine)`.
2. The summary table location string drops columns too, since they no longer participate in the comparison.
3. A long inline comment explains why this is temporary and what the follow-up will do.

**Why not just fix the regression script properly here?**

We do — in #915, stacked on this branch. Once this PR merges to `main`, both pre and post runs are on UTF-16, so column comparison becomes meaningful again and #915 restores the original behaviour. Doing it in-place here would either require leaving `main` as the "loosened" baseline forever, or shipping two competing changes that touch the same lines in the same PR.

**What we are NOT giving up**

Column-level regression detection — for one PR cycle. The check still catches:
- Rules that stop firing
- Rules that fire in new files/lines
- Different error messages on the same line

That's the bulk of what this workflow protects against. Column shifts are rare in practice and the gap is fully closed by #915.

Sorry for the extra commit on this PR — happy to discuss alternatives if you'd prefer to land the kernel change with column-comparison fully disabled long-term, or with a different escape-hatch mechanism.

robertohuertasm-datadog:
* **#915** <a href="https://app.graphite.com/github/pr/DataDog/datadog-static-analyzer/915?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#914** <a href="https://app.graphite.com/github/pr/DataDog/datadog-static-analyzer/914?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/datadog-static-analyzer/914?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

crates/common/src/utils/position_utils.rs
```diff
@@ -1,87 +1,45 @@
...
     /// Converts a tree-sitter 0-based `(row, byte_col)` point to a 1-based UTF-16 code-unit
     /// column.
     ///
-    /// **ASCII fast path**: when the line prefix up to `byte_col` is pure ASCII (all bytes
-    /// < 128), each byte is one UTF-16 code unit, so the result is simply `byte_col + 1`.
-    /// This path costs one slice-level `is_ascii()` check and is taken for the vast majority of
-    /// real-world source lines.
+    /// Falls back to `byte_col + 1` (correct for ASCII-only content) when the coordinates fall
+    /// outside the indexed source, preserving pre-existing behaviour for out-of-range inputs.
     pub fn byte_col_to_utf16_col(&self, row: usize, byte_col: usize) -> u32 {
```
jasonforal: In general, I worry about "core" functions that decide how to handle edge cases for you.

Conceptually this seems to me like [to_str](https://doc.rust-lang.org/std/path/struct.Path.html#method.to_str) from `std::Path` -- something that works pretty much every time (at least for our use case), but could have an edge case.

So I would expect a function like this to have a signature like
```rs
byte_col_to_utf16_col(&self, row: usize, byte_col: usize) -> Option<u32>
```

I think this is particularly important given the baked in assumption:
> correct for ASCII-only content

Thoughts?

robertohuertasm-datadog: I think you have a strong point. I defaulted to `byte_col` because I didn't know if you would be ok with a panic here, or if you'd like to panic at call time (~25 callers) and repeat the same `.expect`  all the time, or maybe go with another default.

Strictly thinking about APIs I think returning an `Option` is probably the cleanest solution. But a panic with a clear message gets us the same runtime behavior (assuming that if the contract breaks, we crash either way) while removing a subtle footgun: with `Option<u32>` a future careless caller can write `.unwrap_or(byte_col as u32 + 1)` and silently reintroduce exactly the ASCII fallback you flagged. With an internal panic that mistake isn't reachable at the call site.

WDYT?

crates/cli/src/csv.rs
```diff
@@ -97,4 +97,39 @@ mod tests {
         );
         assert_eq!(res_with_result, "filename,rule,category,severity,message,start_line,start_col,end_line,end_col\nfilename,myrule,performance,error,message,10,12,12,10\n");
     }
+
+    /// The CSV output faithfully serializes whatever col value is stored in the violation.
+    /// When the kernel produces UTF-16 columns (e.g. col 3 for a node after "🚀"), the CSV row
+    /// contains 3, not the raw byte col (5).
+    #[test]
+    fn test_export_csv_multibyte_col() {
```
jasonforal: Do you think this test is necessary? It seems to cover the same code path as      `test_export_csv`

crates/cli/src/sarif/sarif_utils.rs
```diff
@@ -2231,4 +2231,88 @@ mod tests {
...
     }
+
+    /// The SARIF `region.startColumn` / `region.endColumn` fields carry whatever value is in
+    /// `Position.col`.  When the kernel emits UTF-16 columns (as it now does), the SARIF output
+    /// automatically conforms to the SARIF v2.1 default encoding (UTF-16).
+    ///
+    /// This test constructs a violation whose col was computed as UTF-16 (e.g. after an emoji on
+    /// the same line, col 3) and asserts the SARIF JSON contains that value.
+    #[test]
+    fn sarif_region_carries_utf16_col() {
```
jasonforal: Same conceptual question -- not sure if this is testing anything valuable

crates/common/src/utils/position_utils.rs
```diff
@@ -137,4 +178,116 @@ mod tests {
             Position::new(3, 13)
         );
     }
+
+    // ── LineColumnIndex tests ─────────────────────────────────────────────────
```
jasonforal: Seems like now that we've switched to the crate, we can remove all the vanilla tests covered by line-index and just have a single test confirming we're calling `to_wide(WideEncoding::Utf16)`

crates/static-analysis-kernel/src/analysis/ddsa_lib/bridge/ts_node.rs
```diff
@@ -108,24 +119,28 @@ impl TsNodeBridge {
         self.mirrored_im.as_local(scope)
     }
 
-    /// Inserts the nodes from a [`TSQueryCapture<tree_sitter::Node>`] into a v8 scope, consuming the `QueryCapture`.
-    /// Returns a transformed `QueryCapture` containing the ids of the inserted nodes.
+    /// Inserts the nodes from a [`TSQueryCapture<tree_sitter::Node>`] into a v8 scope, consuming
+    /// the `QueryCapture`. Returns a transformed `QueryCapture` containing the ids of the inserted
+    /// nodes.
+    ///
+    /// `idx` is forwarded to [`insert`](Self::insert) for UTF-16 column conversion.
```
jasonforal: I don't think this comment is needed, feels self-explanatory

crates/static-analysis-kernel/src/analysis/ddsa_lib/bridge/ts_node.rs
```diff
@@ -252,16 +272,50 @@ mod tests {
...
+    /// Byte layout: `"` `🚀`(4 bytes) `"` `;` ` ` = 8 bytes → `abc` at byte 8.
+    /// UTF-16 prefix: `"` + 🚀(2 units) + `"` + `;` + ` ` = 6 units → col 7.
+    #[test]
+    fn ts_node_bridge_multibyte_utf16_col() {
+        let (mut runtime, bridge) = setup_bridge();
+        let scope = &mut runtime.handle_scope();
+        let mut bridge = bridge.borrow_mut();
+
+        // Parse `"\u{1F680}"; abc = 1` as JavaScript.
+        let source = "\"\u{1F680}\"; abc = 1";
```
jasonforal: Not: given that the point of the test is for these "weird" scenarios using literal `\u`, needing to escape the `"` makes the source string slightly harder to read.

I'd suggest using `r#`

crates/static-analysis-kernel/src/analysis/ddsa_lib/context/root.rs
```diff
@@ -45,10 +51,24 @@ impl RootContext {
...
+    /// Also rebuilds the internal [line/column index](Self::line_column_index) cache so that
+    /// subsequent calls to [`line_column_index`](Self::line_column_index) are O(1).
     pub fn set_text(&mut self, text: Arc<str>) -> Option<Arc<str>> {
+        self.lci = Some(LineColumnIndex::new(&text));
         Option::replace(&mut self.tree_text, text)
     }
 
+    /// Returns a reference to the cached [`LineColumnIndex`].
+    ///
+    /// The index is built once in [`set_text`] and reused on every deno op call, avoiding
```
jasonforal: Not sure this comment is needed

crates/static-analysis-kernel/src/analysis/ddsa_lib/js/flow/graph.rs
```diff
@@ -621,11 +622,20 @@ impl From<LocatedEdge<'_>> for dot_structures::Edge {
 
 impl<'a> LocatedNode<'a> {
     /// Constructs a new `LocatedNode` from a tree-sitter node.
-    pub fn new_cst(node: tree_sitter::Node, text: &'a str) -> LocatedNode<'a> {
+    ///
+    /// `idx` must be a [`LineColumnIndex`] built from the same source text that was parsed to
```
jasonforal: Same thing -- I think this reasoning is self-explanatory and just reads like LLM direction instead of documenting

crates/static-analysis-kernel/src/analysis/ddsa_lib/js/ts_node.rs
```diff
@@ -24,26 +25,32 @@ pub struct TreeSitterNode<T> {
...
-    pub fn from_ts_node(id: NodeId, node: tree_sitter::Node) -> Self {
-        // NOTE: We normalize the 0-based `tree_sitter::Point` to be 1-based.
-        fn normalize_ts_point_num(num: usize) -> u32 {
-            num as u32 + 1
-        }
+    /// Converts the provided [`tree_sitter::Node`] into a `TreeSitterNode`, assigning the
+    /// provided id.
+    ///
+    /// Columns are produced as **1-based UTF-16 code-unit** values using `idx`, which must be
+    /// built from the same source string that was parsed to obtain `node`.  Use this constructor
```
jasonforal: > Use this constructor everywhere the source string is available

Comment not needed  -- it's the only constructor :)

crates/static-analysis-kernel/src/analysis/ddsa_lib/js/ts_node.rs
```diff
@@ -152,22 +160,84 @@ mod tests {
...
+        assert_eq!(
+            ts_node.start_col, 7,
+            "UTF-16 col of `foo` should be 7 (emoji as string literal)"
+        );
+
+        // Also verify with a CJK prefix: `"日"; foo(bar);`
+        // 日 = 3 UTF-8 bytes, 1 UTF-16 unit.
+        // Bytes before foo: " (1) + 日 (3) + " (1) + ; (1) + space (1) = 7 bytes.
+        // UTF-16: " (1) + 日 (1) + " (1) + ; (1) + space (1) = 5 units → col 6.
+        let source_cjk = "\"\u{65E5}\"; foo(bar);";
```
jasonforal: Seems like an unnecessary test

crates/static-analysis-kernel/src/analysis/ddsa_lib/runtime.rs
```diff
@@ -1446,6 +1448,47 @@ function visit(query, filename, code) {
...
+        // 🚀 is placed in a string literal so that `protectedName` is a distinct identifier.
+        // Bytes before `protectedName`: " (1) + 🚀 (4) + " (1) + ; (1) + space (1) = 8 bytes.
+        // UTF-16 col: " (1) + 🚀 surrogate pair (2) + " (1) + ; (1) + space (1) = 6 units → col 7.
+        let text = "\"\u{1F680}\"; protectedName = 1;";
+        let filename = "some_filename.js";
+        let ts_query = r#"((identifier) @cap_name (#eq? @cap_name "protectedName"))"#;
+        let rule = r#"
+function visit(query, filename, code) {
+    const node = query.captures["cap_name"];
+    const nodeText = getCodeForNode(node, code);
```
jasonforal: Under the hood, `getCodeForNode` calls the `node.text` getter (which uses the Deno op). This is already tested in the ops module, so this test can be removed

https://github.com/DataDog/datadog-static-analyzer/blob/bd396d7d58e739c5721baa1dfd6e75f93c99c65d/crates/static-analysis-kernel/src/analysis/ddsa_lib/js/stella_compat.js#L67-L70

jasonforal (commented):
Ok, I started to leave some comments but then noticed an overall pattern, so some of my comments will overlap with this summary, but I left them in just to demonstrate what I meant. But they aren't exhaustive so please take a look over the full diff to find other examples:

**Redundant Tests**
Now that we've switched to the `line-index` crate, the only unit test we need for each fixed call site that previous just did `+ 1` is that we're calling `byte_col_to_utf16_col`. So just one source string is needed to prove that. We trust the `byte_col_to_utf16_col` unit tests cover all these other cases.

In general, I prefer tests to be single-responsibility (no redundant code paths tested) and highly readable

**Unnecessary doc comments**
Seems like a lot of LLM comments leaked into the doc comments (like how `idx` is passed to provide O(1) lookup). I think this is just noise. If it was a code comment, I'd question it but let it go, but since these are doc comments, they are just clutter.

I think the reason to pass in the index is self-explanatory and doesn't need any documentation.

---

Otherwise, this is a great fix!

crates/common/src/utils/position_utils.rs
jasonforal: To be honest, I think we can be "future careless callers", today :)

All the tests can use `unwrap`, and so I'm counting just 7 where we do `unwrap_or` with the + 1.

If we run into a scenario where it's actually a problem, we can just update all of those.

crates/cli/src/csv.rs
robertohuertasm-datadog: Agreed, removed in 1f052b9f.

crates/cli/src/sarif/sarif_utils.rs
robertohuertasm-datadog: Agreed, removed in 1f052b9f.

crates/common/src/utils/position_utils.rs
robertohuertasm-datadog: Done in 1f052b9f — collapsed the eight `lci_*` tests into a single wiring test that proves we call `to_wide(WideEncoding::Utf16, ..)` and add 1.

crates/static-analysis-kernel/src/analysis/ddsa_lib/bridge/ts_node.rs
robertohuertasm-datadog: Removed in 1f052b9f.

crates/static-analysis-kernel/src/analysis/ddsa_lib/bridge/ts_node.rs
robertohuertasm-datadog: Good catch — the test itself was removed in 1f052b9f (now covered by the single wiring test in `position_utils.rs`), so the escaped string goes with it.

crates/static-analysis-kernel/src/analysis/ddsa_lib/context/root.rs
robertohuertasm-datadog: Removed in 1f052b9f.

crates/static-analysis-kernel/src/analysis/ddsa_lib/js/flow/graph.rs
robertohuertasm-datadog: Removed in 1f052b9f.

crates/static-analysis-kernel/src/analysis/ddsa_lib/js/ts_node.rs
robertohuertasm-datadog: Removed in 1f052b9f.

crates/static-analysis-kernel/src/analysis/ddsa_lib/js/ts_node.rs
robertohuertasm-datadog: Removed in 1f052b9f.

crates/static-analysis-kernel/src/analysis/ddsa_lib/runtime.rs
robertohuertasm-datadog: Removed in 1f052b9f.

crates/common/src/utils/position_utils.rs
robertohuertasm-datadog: Done in 1f052b9f — `byte_col_to_utf16_col` now returns `Option<u32>`, the seven production call sites use `unwrap_or(byte_col as u32 + 1)`, and tests just `unwrap()`.

crates/common/src/model/position.rs
```diff
@@ -6,6 +6,18 @@ use derive_builder::Builder;
 use serde::{Deserialize, Serialize};
 use std::fmt;
 
+/// A source-code position.
```
robertohuertasm-datadog, line 9: @jasonforal let me know if you want me to remove this comment. I thought it could be useful for the readers to understand what to expect, and why. Happy to remove it if you think it's not worth it.

jasonforal: Nah, this is fine, especially because of the differing behavior between the `static-analysis-kernel` and `secrets`

jasonforal (approved):

PR #921, opened 2026-06-16
[INC-56036] avoid stack overflow in secrets scanner by splitting scan and validate

MikaYuoadas:
## Summary

`Scanner::scan_with_options(validate_matches=true)` (introduced in [f51ffceb [SDS] Implement SDS supporting rules](https://github.com/DataDog/datadog-static-analyzer/commit/f51ffceb805564fd3650e4bffae79e6b5b557b94)) triggers a stack overflow inside `dd-sds` when scanning very large monorepos. In prod, `code-workload-runner-secrets` has been crashing ~40-75k tasks/day on Datadog dogfood orgs since the v0.8.5 cut on 2026-05-29.

Switch `find_secrets` back to the pre-v0.8.5 pattern: `scan()` to collect matches, then `validate_matches()` as a separate step (which uses the bounded rayon thread pool internally).

The two paths are equivalent for plain validators. They differ for the new `is_supporting_rule` HTTP match-pairing feature: cross-rule template-variable resolution (a supporting rule's match feeding `$PARAM` placeholders in a primary rule's validator URL) only fires inside `scan_with_options`. The corresponding test is marked `#[ignore]` until the underlying recursion is fixed upstream in `dd-sds`.

Incident: https://app.datadoghq.com/incidents/56036

## Test plan

- [x] `cargo test --workspace` — 487 passed, 2 ignored.
- [x] Reproduced the crash on the full `dd-source` monorepo with the staging configuration before the fix (`fatal runtime error: stack overflow, aborting`).
- [x] After the fix, full `dd-source` scan (894k files) completes in ~140s with identical results to the pre-regression v0.8.4 baseline (515 files with secrets, 1020 total).

### DIFF
--- a/crates/secrets/src/scanner.rs
+++ b/crates/secrets/src/scanner.rs
@@ -9,7 +9,7 @@ use anyhow::Error;
 use common::analysis_options::AnalysisOptions;
 use common::model::position::Position;
 use common::utils::position_utils::get_position_in_string;
-use dd_sds::{RootRuleConfig, RuleConfig, ScanOptionBuilder, Scanner};
+use dd_sds::{RootRuleConfig, RuleConfig, Scanner};
 use itertools::Itertools;
 use std::sync::Arc;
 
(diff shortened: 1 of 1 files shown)


isabella-garza-datadog (approved):
LGTM as a short term fix while we work on resolving the root cause in `dd-sds`

jasonforal (approved):

PR #923, opened 2026-06-17
feat(ide/config): add unified config support to the static-analyzer server

robertohuertasm-datadog:
## Context

As part of the work described in [Section 4 — Prerequisite: Unified Config File Support](https://docs.google.com/document/d/1L4rdpTwafiVxFn1HIlSw0f6cJ61GihDOzMFvZwuzlMw/edit?tab=t.0#heading=h.mr90lg23mmfd), the static-analyzer server needs to understand both the legacy `static-analysis.datadog.yml` format and the new unified `code-security.datadog.yaml` format (schema version `v1.0`), so the VS Code extension can drive mutations against whichever file the user has.

The dd-source backend already supports `schema_version=v1` (unified) on `POST /api/v2/static-analysis/config/client`. This PR covers only the **local static-analyzer server** side — the VS Code extension changes will follow in a [separate PR](https://github.com/DataDog/datadog-vscode/pull/2459).

## Problem

The server had two issues that would prevent it from being usable as a unified-config host:

1. **Production parsing was locked to legacy.** `from_yaml_content` called `parse_any_schema_yaml` only under `cfg!(test)` and `file_legacy::parse_yaml` in release builds. A `code-security.datadog.yaml` sent to the server would silently be parsed as legacy content — or fail without a useful error — making every mutation endpoint incorrect for unified files.

2. **No format signal on mutation/validation requests.** Endpoints like `ignore-rule`, `add-rulesets`, and `can-onboard` had no way for the caller to declare what format the file is in. Without that signal, the server could not validate that the content matches the declared filename format (matching CLI behaviour), could not pick the right empty-file default when creating a new config, and could not produce the correct YAML serialization on write-back.

3. **Blank-content edge case.** When an existing but empty `code-security.datadog.yaml` (e.g. just created, content = `base64("\n")`) was sent to `add-rulesets`, `from_yaml_content`'s early-return for blank content always yielded a legacy default, which then failed `validate_format(Unified)` with `SchemaMismatch` even though there was nothing wrong.

## Approach: Option A — `schema_version` field on existing routes, no new routes

Following the design in the RFC above, format is signalled via an optional `schema_version` field added to the JSON body of existing endpoints. This avoids proliferating product-scoped routes: the IDE determines format from the filename, sends it as `"v1"` (unified) or omits it (legacy), and the server validates accordingly.

**Backward compatibility is a hard requirement.** Old extensions that never send `schema_version` must continue to work unchanged. This is guaranteed by:
- `schema_version` is `#[serde(default)] Option<String>` on every request struct — missing field deserializes as `None`.
- `None` → `ConfigFormat::Legacy` everywhere.
- No `#[serde(deny_unknown_fields)]` on request structs, so a new extension talking to an old binary simply has its `schema_version` field ignored.

### `static_analysis_config_file.rs`

- **Un-gated dual-format parsing.** Removed the `cfg!(test)` guard; production now always calls `parse_any_schema_yaml`, which detects the actual schema variant and stores it in `WithVersion<Legacy, CodeSecurity>`.
- **`ConfigFormat` enum + `detected_format()` accessor.** `ConfigFormat::Legacy` / `ConfigFormat::Unified` derived from the `WithVersion` discriminant. Used everywhere format validation is needed.
- **`validate_format(declared)`** method returns `Err(SchemaMismatch)` when detected ≠ declared, mirroring CLI filename-strict behaviour.
- **`default_unified()` constructor.** Returns an empty `WithVersion::CodeSecurity` config — used when `add-rulesets` is called with no existing content and `schema_version=v1`.
- **`WithHint<T>` wrapper + new `TryFrom` impls.** Replaces the old format-unaware `TryFrom<String>` / `TryFrom<Base64String>`. `WithHint<T>(T, Option<ConfigFormat>)` carries the caller's declared format so `from_yaml_content` can pick the right empty default. Callers with `schema_version` pass `Some(format)`; deprecated/legacy routes pass `None`.
- **Bug fix — blank content edge case.** `from_yaml_content` now accepts `Option<ConfigFormat> hint`; when content is blank it returns `default_unified()` for `Some(Unified)` and the legacy default otherwise. This prevents `add-rulesets` and `ignore-rule` from failing with `SchemaMismatch` when a `code-security.datadog.yaml` exists but is empty.
- **`with_ignored_rule` and `with_added_rulesets`** updated to forward `declared_format` via `WithHint`.

### `models.rs`

Added `#[serde(default)] schema_version: Option<String>` to:
- `IgnoreRuleRequest`
- `AddRuleSetsRequest`
- `CanOnboardRequest`
- `ParseConfigRequest`

### `endpoints.rs`

- **`post_parse_config`** — now accepts `schema_version`; constructs via `WithHint(configuration, Some(format))`, then calls `validate_format` — returns `400` on mismatch.
- **`post_can_onboard_v2`** — now accepts `schema_version`; same parse + validate pattern; returns `500` on mismatch (existing error path the client already treats as blocking).
- **`post_ignore_rule` / `add_rulesets`** — delegate to the updated `with_ignored_rule` / `with_added_rulesets` which now carry the format hint.
- **Deprecated routes** (`get_can_onboard`, `get_rulesets`) updated to `WithHint(..., None)` — legacy-default, no behaviour change.

### `error.rs`

Added `SchemaMismatch` variant (`code: 4`) to `ConfigFileError`.

### `mod.rs` (IDE integration tests)

Extended the test suite with coverage for:
- Strict parse: unified content + `schema_version=v1` → 200; legacy content + `schema_version=v1` → 400 (`SchemaMismatch`); unified content + no `schema_version` → 400.
- Mutation (`add-rulesets`, `ignore-rule`) format validation.
- `can-onboard` with `schema_version`.
- Backward compat: legacy content + no `schema_version` → behaves as today.
- Empty content + `schema_version=v1` → creates a unified default file.

### `file_v1.rs`

Added `Default` impl for `YamlConfigFile`, needed by `default_unified()`.

## Validation

`cargo test` ->? 77 tests, all pass 
`cargo fmt -- --check` -> clean 
`cargo clippy -- -D warnings` -> clean

Retrocompatibility: Tested this binary with the current extension and it works. 

## What's not in this PR

The VS Code extension changes (unified-first discovery, sending `schema_version` in request bodies, binary version gate, etc.) are the next step and will be in a separate PR against the `datadog-vscode` repo.

K9CODESEC-2169
K9CODESEC-2303

### DIFF
--- a/crates/bins/src/bin/datadog_static_analyzer_server/ide/configuration_file/endpoints.rs
+++ b/crates/bins/src/bin/datadog_static_analyzer_server/ide/configuration_file/endpoints.rs
@@ -7,7 +7,9 @@ use crate::datadog_static_analyzer_server::ide::configuration_file::models::{
 
 use super::error::ConfigFileError;
 use super::models::{AddRuleSetsRequest, IgnoreRuleRequest};
-use super::static_analysis_config_file::{Base64String, StaticAnalysisConfigFile};
+use super::static_analysis_config_file::{
+    Base64String, ConfigFormat, StaticAnalysisConfigFile, WithHint,
+};
 use kernel::utils::encode_base64_string;
 use rocket::http::Status;
 use rocket::response::status::Custom;
@@ -35,12 +37,13 @@ pub fn post_ignore_rule(
         rule,
         configuration_base64,
         encoded,
-        ..
+        schema_version,
     } = request.into_inner();
     tracing::debug!(rule, content = &configuration_base64);
     let result = StaticAnalysisConfigFile::with_ignored_rule(
         rule.into(),
         Base64String(configuration_base64),
+        ConfigFormat::from(schema_version.as_deref()),
     );
     to_response_result(result, encoded)
 }
@@ -51,10 +54,18 @@ pub fn post_ignore_rule(
 /// * `content` - The path to the configuration file.
 #[instrument()]
 #[rocket::get("/v1/config/can-onboard/<content..>")]
-#[deprecated(note = "IDEs stopped supporting onboarding")]
 pub fn get_can_onboard(content: PathBuf) -> Result<Json<bool>, Custom<ConfigFileError>> {
-    let content_str = content.to_string_lossy().into_owned();
-    can_onboard(content_str)
+    // NOTE: this method is still used by Visual Studio.
+    let mut content_str = content.to_string_lossy().into_owned();
+    if cfg!(target_os = "windows") {
+        // NOTE: this is needed due to how Rocket works with multiple segment captures.
+        content_str = content_str.replace("\\", "/");
+    }
+    let config = StaticAnalysisConfigFile::try_from(WithHint(Base64String(content_str), None))
+        .map_err(|e| Custom(Status::InternalServerError, e))?;
+
+    let can_onboard = config.is_onboarding_allowed();
+    Ok(Json(can_onboard))
 }
 
 /// Checks if onboarding is allowed for the configuration file (v2).
@@ -67,15 +78,27 @@ pub fn get_can_onboard(content: PathBuf) -> Result<Json<bool>, Custom<ConfigFile
     format = "application/json",
     data = "<request>"
 )]
-#[deprecated(note = "IDEs stopped supporting onboarding")]
 pub fn post_can_onboard_v2(
     request: Json<CanOnboardRequest>,
 ) -> Result<Json<bool>, Custom<ConfigFileError>> {
     let CanOnboardRequest {
         configuration_base64,
-        ..
+        schema_version,
     } = request.into_inner();
-    can_onboard(configuration_base64)
+
+    let format = ConfigFormat::from(schema_version.as_deref());
+    let config = StaticAnalysisConfigFile::try_from(WithHint(
+        Base64String(configuration_base64),
+        Some(format),
+    ))
+    .map_err(|e| Custom(Status::InternalServerError, e))?;
+
+    config
+        .validate_format(format)
+        .map_err(|e| Custom(Status::InternalServerError, e))?;
+
+    let can_onboard = config.is_onboarding_allowed();
+    Ok(Json(can_onboard))
 }
 
 /// Gets the rulesets from the static analysis configuration file (deprecated).
@@ -117,6 +140,9 @@ pub fn post_get_rulesets_v2(request: Json<GetRulesetsRequest>) -> Json<Vec<Strin
 
 /// Parses the static analysis configuration YAML and returns per-product fields.
 ///
+/// Returns `400` when the content cannot be parsed or when the detected schema does not
+/// match the caller-declared `schema_version` — mirroring the CLI's filename-strict behavior.
+///
 /// # Arguments
 /// * `request` - The request containing the raw YAML string.
 #[instrument()]
@@ -124,9 +150,16 @@ pub fn post_get_rulesets_v2(request: Json<GetRulesetsRequest>) -> Json<Vec<Strin
 pub fn post_parse_config(
     request: Json<ParseConfigRequest>,
 ) -> Result<Json<ParseConfigResponse>, Custom<ConfigFileError>> {
-    let ParseConfigRequest { configuration } = request.into_inner();
+    let ParseConfigRequest {
+        configuration,
+        schema_version,
+    } = request.into_inner();
     tracing::debug!(%configuration);
-    let config = StaticAnalysisConfigFile::try_from(configuration)
+    let format = ConfigFormat::from(schema_version.as_deref());
+    let config = StaticAnalysisConfigFile::try_from(WithHint(configuration, Some(format)))
+        .map_err(|e| Custom(Status::BadRequest, e))?;
+    config
+        .validate_format(format)
         .map_err(|e| Custom(Status::BadRequest, e))?;
     Ok(Json(ParseConfigResponse {
         sast: SastParsedConfig {
@@ -169,22 +202,21 @@ pub fn post_add_rulesets_v2(
 
 /// Adds rulesets to the configuration file.
 ///
-/// # Arguments
-/// * `request` - The request containing rulesets and configuration file (base64).
+/// When no config content is supplied, creates a new file in the format implied by
+/// `schema_version` (unified when `"v1"`, legacy otherwise). When content is supplied,
+/// validates that the detected format matches the declared one before mutating.
 fn add_rulesets(request: Json<AddRuleSetsRequest>) -> Result<String, Custom<ConfigFileError>> {
     let AddRuleSetsRequest {
         rulesets,
         configuration_base64,
         encoded,
-        ..
+        schema_version,
     } = request.into_inner();
-    tracing::debug!(
-        rulesets=?&rulesets,
-        content=&configuration_base64
-    );
+    tracing::debug!(rulesets=?&rulesets, content=&configuration_base64);
     let result = StaticAnalysisConfigFile::with_added_rulesets(
         &rulesets,
         configuration_base64.map(Base64String),
+        ConfigFormat::from(schema_version.as_deref()),
     );
     to_response_result(result, encoded)
 }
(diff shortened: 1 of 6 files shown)


crates/bins/src/bin/datadog_static_analyzer_server/ide/configuration_file/static_analysis_config_file.rs
```diff
@@ -14,6 +14,26 @@ use tracing::instrument;
...
+impl<'a> From<Option<&'a str>> for ConfigFormat {
+    /// Maps the IDE-supplied `schema_version` string to a [`ConfigFormat`].
+    /// Absent or unrecognized values default to [`ConfigFormat::Legacy`] for backward
+    /// compatibility with older extensions that never sent this field.
+    fn from(schema_version: Option<&'a str>) -> Self {
+        match schema_version {
+            Some("v1") => Self::Unified,
+            _ => Self::Legacy,
+        }
+    }
```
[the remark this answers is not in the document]
robertohuertasm-datadog: Thanks for the review! These are two distinct version strings at different layers that should not be conflated:

- `schema-version: v1.0` is the version embedded **inside the YAML file content**, used by the parser (`file_v1`) to identify the unified schema format.
- `schema_version` in the **JSON request body** is a simple discriminator sent by the VS Code extension to signal which format the file uses, derived purely from the filename (`code-security.datadog.*` → `"v1"`, `static-analysis.datadog.*` → omitted). It is not derived from the file content.

No client would ever send `"v1.0"` in the request body — the mapping is defined by us (the server contract and the extension we're writing), not by end users. The `"v1"` → Unified mapping is intentional per the design. The two version strings serve different purposes and are completely independent.

crates/bins/src/bin/datadog_static_analyzer_server/ide/configuration_file/endpoints.rs
```diff
@@ -51,10 +54,18 @@ pub fn post_ignore_rule(
 /// * `content` - The path to the configuration file.
```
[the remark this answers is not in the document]
robertohuertasm-datadog: Fixed in 245ba4f5 — updated the doc comment to accurately describe the parameter as base64-encoded YAML content passed as a URL path segment.

crates/bins/src/bin/datadog_static_analyzer_server/ide/configuration_file/endpoints.rs
```diff
@@ -35,12 +37,13 @@ pub fn post_ignore_rule(
...
-        ..
+        schema_version,
     } = request.into_inner();
     tracing::debug!(rule, content = &configuration_base64);
     let result = StaticAnalysisConfigFile::with_ignored_rule(
         rule.into(),
         Base64String(configuration_base64),
+        ConfigFormat::from(schema_version.as_deref()),
     );
     to_response_result(result, encoded)
```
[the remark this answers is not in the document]
robertohuertasm-datadog: Good point — fixed in 245ba4f5. `to_response_result` now returns `400 Bad Request` for `SchemaMismatch` (client input error) and keeps `500 Internal Server Error` for genuine server failures. Updated the integration test accordingly.

crates/bins/src/bin/datadog_static_analyzer_server/ide/configuration_file/endpoints.rs
```diff
@@ -35,26 +37,35 @@ pub fn post_ignore_rule(
...
-    let content_str = content.to_string_lossy().into_owned();
-    can_onboard(content_str)
+    // NOTE: this method is still used by Visual Studio.
+    let mut content_str = content.to_string_lossy().into_owned();
+    if cfg!(target_os = "windows") {
+        // NOTE: this is needed due to how Rocket works with multiple segment captures.
+        content_str = content_str.replace("\\", "/");
+    }
+    let config = StaticAnalysisConfigFile::try_from(WithHint(Base64String(content_str), None))
+        .map_err(|e| Custom(Status::InternalServerError, e))?;
```
jasonforal: Sounds like from the description that this 500 is intentional. Would you please add a comment explaining why? I would otherwise wonder why it's not 400 like elsewhere

crates/bins/src/bin/datadog_static_analyzer_server/ide/configuration_file/endpoints.rs
```diff
@@ -67,15 +78,27 @@ pub fn get_can_onboard(content: PathBuf) -> Result<Json<bool>, Custom<ConfigFile
...
+    let format = ConfigFormat::from(schema_version.as_deref());
+    let config = StaticAnalysisConfigFile::try_from(WithHint(
+        Base64String(configuration_base64),
+        Some(format),
+    ))
+    .map_err(|e| Custom(Status::InternalServerError, e))?;
+
+    config
+        .validate_format(format)
+        .map_err(|e| Custom(Status::InternalServerError, e))?;
```
jasonforal: And maybe a comment why this isn't 422?

crates/bins/src/bin/datadog_static_analyzer_server/ide/configuration_file/static_analysis_config_file.rs
```diff
@@ -243,17 +287,26 @@ impl StaticAnalysisConfigFile {
...
+                let config =
+                    Self::try_from(WithHint(content, Some(declared_format))).map_err(|e| {
+                        tracing::error!(error = ?e, "Error trying to parse config file");
+                        e
+                    })?;
+                config.validate_format(declared_format)?;
+                config
+            }
+            None => match declared_format {
+                ConfigFormat::Unified => Self::default_unified(),
```
jasonforal: Seeing this here made me wonder: should we change the `impl Default for StaticAnalysisConfigFile` to return a unified file by default and have "default_legacy" be the exception?

Only thinking if we end up having other places to create these structs, it might not be intuitive why the Default impl for a legacy format. Up to you

jasonforal (dismissed):

crates/bins/src/bin/datadog_static_analyzer_server/ide/configuration_file/endpoints.rs
robertohuertasm-datadog: Good catch! You're right,  it should be 400 and I've corrected it. Any error on this route is a parse failure on client-supplied content (bad base64 or invalid YAML), so 500 was semantically wrong. I initially kept the original 500 to avoid inadvertently breaking the Visual Studio extension (the only known caller of this deprecated route), but after checking its source I confirmed it uses a plain try/catch with no status-code inspection, so the change is safe. A comment has been added to the code explaining this. Thanks for flagging it!

crates/bins/src/bin/datadog_static_analyzer_server/ide/configuration_file/endpoints.rs
robertohuertasm-datadog: Fixed, changed to 422 Unprocessable Entity, consistent with every other SchemaMismatch mapping in the file. I initially kept the original 500 behavior here as well out of caution, to avoid breaking older clients. After verifying that the Visual Studio extension (the only client calling the deprecated v1 route) does not inspect specific HTTP status codes, and that VS Code's canOnboard uses a generic try/catch with no status-code check either, we confirmed the change is safe to make.

crates/bins/src/bin/datadog_static_analyzer_server/ide/configuration_file/static_analysis_config_file.rs
robertohuertasm-datadog: Great point, done. `Default` now returns a unified file, and `default_legacy()` is the explicitly named fallback used only for the backward-compat paths (blank content with no declared format, and `ConfigFormat::Legacy` in `with_added_rulesets`).

jasonforal (approved):

albertvaka (approved):
Stamping so it can be merged, I have not reviewed it though.

MarshalX (approved):
Stamping so it can be merged, I have not reviewed it though.

PR #933, opened 2026-06-25
Add ignore-paths/only-paths support to the secrets config section

MikaYuoadas:
## What

Introduces a typed `secrets` section in `code-security.datadog.yaml` (schema v1.4) with a `global-config` block supporting `ignore-paths` and `only-paths`, allowing users to control which files are included in secrets scanning.

```yaml
schema-version: v1.4
secrets:
  global-config:
    ignore-paths:
      - "tests/**"
      - "vendor/**"
```

This mirrors the existing `sast` and `iac` schema.

## Why

Previously there was no way to exclude files from secrets scanning via configuration.

## Behavior

Per-section validation is now conditional on whether the product is enabled. `parse_yaml` accepts `parse_sast` and `parse_secrets` flags (forwarded from the binary's `--enable-static-analysis` / `--enable-secrets` flags):

- An invalid `secrets` section causes a failure (exit 50, product-specific message) only when secrets scanning is enabled.
- Symmetrically an invalid `sast` section only cause a failure when static analysis scanning is enabled.
- When a product is disabled, its config section is silently skipped so, as per the RFC, one product's malformed config cannot break another.

Two new `ParseError` variants (`SastParse`, `SecretsParse`) carry product-specific messages while reusing the existing `EXIT_CODE_INVALID_CONFIGURATION`.

### DIFF
--- a/crates/bins/src/bin/datadog-static-analyzer-git-hook.rs
+++ b/crates/bins/src/bin/datadog-static-analyzer-git-hook.rs
@@ -27,6 +27,7 @@ use kernel::analysis::ddsa_lib::v8_platform::{initialize_v8, Initialized, V8Plat
 use kernel::classifiers::ArtifactClassification;
 use kernel::config::common::{ConfigMethod, PathConfig};
 use kernel::config::file_v1;
+use kernel::config::file_v1::ParseError;
 use kernel::constants::{CARGO_VERSION, VERSION};
 use kernel::model::common::OutputFormat::Json;
 use kernel::model::rule::{Rule, RuleResult};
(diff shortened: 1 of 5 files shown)


jasonforal:
Nice! This is a good starting point.

At a high level, I think it's wise to avoid enmeshing the SAST/Secrets products any further than they already are.

For example, instead of modifying SAST's parse function signature to include [`parse_sast`, `parse_secrets`](https://github.com/DataDog/datadog-static-analyzer/pull/933/changes#diff-8d120e3237acf1ab131a25f1e4efb9f6f89532bad0fdb049e895f07a4da46983R34-R35), we should write an entirely separate Secrets config parser -- conceptually something like `secrets::file_v1::parse_yaml(&contents)`.

Ideally, the implementation would be designed such that we could eventually split the Secrets scanner off into a separate binary without touching any of SAST code.

PR #935, opened 2026-07-08
Scan git history for removed secrets

MikaYuoadas:
## What

Adds a `--scan-git-history-only` flag to the secrets scanner. When enabled, the scan reports secrets that existed in past commits but are no longer present at the branch HEAD.

The scan runs in two passes:
- **Pass 1** scans every unique blob in the object database once (content-addressed, so each blob is scanned a single time regardless of how many paths or commits reference it).
- **Pass 2** attributes only the blobs that actually contained a secret back to their path, introducing commit, and removal commit by streaming a single `git log --raw` over all branches. A blob with no reachable addition on any branch is dropped.

Historic findings are published in a separate SARIF run under the `datadog-static-analyzer-secrets-history` tool driver, so they stay distinct from the regular HEAD scan results.

## Why

Secrets that were committed and later deleted still live in the repository history and can be recovered by anyone with a clone. A HEAD-only scan never surfaces them. This flag lets a scan cover that gap without slowing down or altering the normal HEAD scan, which continues to run as before.

## Notes

- The two-pass design keeps the hot path off any per-commit tree walk: Pass 1 is linear in the number of unique blobs, and the expensive reverse-mapping in Pass 2 only runs for the handful of blobs that contain a secret.
- Blob scanning in Pass 1 runs in parallel.

### DIFF
--- a/crates/bins/src/bin/datadog-static-analyzer.rs
+++ b/crates/bins/src/bin/datadog-static-analyzer.rs
@@ -5,10 +5,10 @@ use std::path::PathBuf;
 
 use cli::config_file::get_config;
 use cli::constants::{
-    DEFAULT_MAX_CPUS, DEFAULT_MAX_FILE_SIZE_KB, EXIT_CODE_FAIL_ON_VIOLATION,
+    DEFAULT_MAX_CPUS, DEFAULT_MAX_FILE_SIZE_KB, DEFAULT_TOOL_NAME, EXIT_CODE_FAIL_ON_VIOLATION,
     EXIT_CODE_INVALID_CONFIGURATION, EXIT_CODE_INVALID_DIRECTORY, EXIT_CODE_NO_DIRECTORY,
     EXIT_CODE_NO_OUTPUT, EXIT_CODE_RULESET_NOT_FOUND, EXIT_CODE_RULE_FILE_WITH_CONFIGURATION,
-    EXIT_CODE_UNSAFE_SUBDIRECTORIES,
+    EXIT_CODE_UNSAFE_SUBDIRECTORIES, SECRETS_HISTORY_TOOL_NAME,
 };
 use cli::csv;
 use cli::datadog_utils::{
@@ -25,12 +25,14 @@ use cli::rule_utils::{
     convert_secret_result_to_rule_result, count_violations_by_severities, get_languages_for_rules,
     get_rulesets_from_file,
 };
-use cli::sarif::sarif_utils::{generate_sarif_file, SarifReportMetadata};
+use cli::sarif::sarif_utils::{generate_sarif_file, HistoricalSecretResult, SarifReportMetadata};
 use cli::utils::{choose_cpu_count, print_configuration};
 use cli::violations_table;
 use common::analysis_options::AnalysisOptions;
 use common::model::diff_aware::DiffAware;
-use datadog_static_analyzer::{secret_analysis, static_analysis, CliResults};
+use datadog_static_analyzer::{
+    git_history_secret_analysis, secret_analysis, static_analysis, CliResults,
+};
 use kernel::analysis::ddsa_lib::v8_platform::{initialize_v8, Initialized, V8Platform};
 use kernel::analysis::generated_content::DEFAULT_IGNORED_GLOBS;
 use kernel::classifiers::ArtifactClassification;
@@ -40,7 +42,7 @@ use kernel::constants::{CARGO_VERSION, VERSION};
 use kernel::model::common::OutputFormat;
 use kernel::model::rule::{Rule, RuleResult, RuleSeverity};
 use kernel::rule_config::RuleConfigProvider;
-use secrets::model::secret_result::SecretValidationStatus;
+use secrets::model::secret_result::{SecretResult, SecretValidationStatus};
 use secrets::secret_files::should_ignore_file_for_secret;
 use std::collections::HashMap;
 use std::io::prelude::*;
@@ -154,6 +156,14 @@ fn main() -> Result<()> {
         "how long a rule can run before being killed, in milliseconds",
         "1000",
     );
+    opts.optflag(
+        "",
+        "scan-git-history-only",
+        "scan the entire git history for secrets that are not present at the current HEAD. Every \
+         commit is scanned (not just what current refs point to); a finding is 'history-only' \
+         relative to HEAD, so a secret still live at the tip of another branch, tag, or remote may \
+         also be reported as history-only.",
+    );
 
     let matches = match opts.parse(&args[1..]) {
         Ok(m) => m,
@@ -227,6 +237,15 @@ fn main() -> Result<()> {
         .map(|value| value == "true" || value == "yes")
         .unwrap_or(false);
     let secrets_enabled = secrets_enabled_old_option || secrets_enabled_new_option;
+
+    let scan_git_history_only = matches.opt_present("scan-git-history-only");
+
+    // A git-history scan produces a historic-only secrets report, so it is meaningless
+    // without secrets enabled. Fail fast rather than silently running a no-op.
+    if scan_git_history_only && !secrets_enabled {
+        eprintln!("--scan-git-history-only requires secrets scanning; pass --enable-secrets true");
+        exit(EXIT_CODE_INVALID_CONFIGURATION);
+    }
 
     let output_file = matches
         .opt_str("o")
@@ -611,8 +630,9 @@ fn main() -> Result<()> {
     }
 
     // Secrets detection
-
-    if secrets_enabled {
+    //
+    // Skipped entirely for a git-history scan.
+    if secrets_enabled && !scan_git_history_only {
         let secrets_start = Instant::now();
 
         let secrets_files: Vec<PathBuf> = files_in_repository
@@ -671,6 +691,23 @@ fn main() -> Result<()> {
         result.secrets = Some(execution_results);
     }
 
+    // Git history scanning
+    //
+    // When `--scan-git-history-only` is set, the secrets results are replaced with the
+    // historical findings (the HEAD secret scan is skipped above). Static analysis is an
+    // independent product and is left untouched.
+    let mut historic_secrets: Vec<HistoricalSecretResult> = Vec::new();
+    if scan_git_history_only {
+        let history_start = std::time::Instant::now();
+        historic_secrets = git_history_secret_analysis(&configuration, &analysis_options)
+            .context("git history secret analysis failed")?;
+
+        let history_duration = history_start.elapsed().as_secs_f64();
+        println!("Git History Secrets Summary");
+        println!("  Historical secrets found: {}", historic_secrets.len());
+        println!("  Duration: {:.3}s", history_duration);
+    }
+
     let global_execution_time_secs = global_start_time.elapsed().as_secs();
 
     // if we have more than one static analysis violation and printing is enabled, show all
@@ -685,6 +722,15 @@ fn main() -> Result<()> {
     let secrets_violations = secrets_violations
         .map(|r| r.rule_results)
         .unwrap_or_default();
+
+    // Historic findings carry their commit provenance in `historic_secrets` (consumed by the SARIF
+    // output). For JSON/CSV, surface their base results alongside any HEAD secrets. HEAD and
+    // historic scans are mutually exclusive, so at most one of the two is non-empty.
+    let secrets_for_text_output: Vec<SecretResult> = secrets_violations
+        .iter()
+        .cloned()
+        .chain(historic_secrets.iter().map(|h| h.inner.clone()))
+        .collect();
 
     let nb_total_static_analysis_violations: usize = static_analysis_rule_results
         .iter()
@@ -705,7 +751,7 @@ fn main() -> Result<()> {
 
     let value = match configuration.output_format {
         OutputFormat::Csv => {
-            csv::generate_csv_results(&static_analysis_rule_results, &secrets_violations)
+            csv::generate_csv_results(&static_analysis_rule_results, &secrets_for_text_output)
         }
         OutputFormat::Json => {
             // make sure suppressed results are not included
@@ -717,8 +763,7 @@ fn main() -> Result<()> {
                     r
                 })
                 .collect();
-            let filtered_secrets: Vec<RuleResult> = secrets_violations
-                .clone()
+            let filtered_secrets: Vec<RuleResult> = secrets_for_text_output
                 .iter()
                 .map(convert_secret_result_to_rule_result)
                 .map(|mut r| {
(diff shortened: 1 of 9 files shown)


crates/bins/src/bin/datadog-static-analyzer.rs
```diff
@@ -228,6 +235,16 @@ fn main() -> Result<()> {
...
 
+    let scan_git_history_only = matches.opt_present("scan-git-history-only");
+
+    // A git-history scan produces a historic-only secrets report, so it is meaningless
+    // without secrets enabled. Fail fast rather than silently running a no-op.
+    if scan_git_history_only && !secrets_enabled {
+        eprintln!("--scan-git-history-only requires secrets scanning; pass --enable-secrets true");
+        exit(EXIT_CODE_INVALID_CONFIGURATION);
+    }
+    let git_history_only = scan_git_history_only;
```
robertohuertasm-datadog: NIT: Consider removing this alias. It adds indirection without meaning. The name `git_history_only` is not more descriptive than `scan_git_history_only`. Now there are two variables holding the same boolean, which forces the reader to mentally track both names and confirm they're identical. WDYT?

crates/bins/src/lib.rs
```diff
@@ -359,6 +376,363 @@ pub fn secret_analysis(
     Ok(analyzer_result)
 }
 
+/// Get the `BlobOccurrence` for a (blob, path) pair, creating an empty one if it
+/// does not exist yet.
+fn find_or_create_occurrence<'a>(
```
robertohuertasm-datadog: NIT: We could use the entry API to simplify and remove a possible-panic path. With #[derive(Default)] added to `BlobOccurrence` (trivially valid, since both fields are `Option` and default to `None`), this collapses to:

```rs
fn find_or_create_occurrence<'a>(
    registry: &'a mut HashMap<Oid, HashMap<String, BlobOccurrence>>,
    blob_oid: Oid,
    path: &str,
) -> &'a mut BlobOccurrence {
    registry
        .entry(blob_oid)
        .or_default()
        .entry(path.to_string())
        .or_default()
}
```

This is functionally identical (returns the existing occurrence if present, or a fresh `{ None, None }` one otherwise), but it's fewer lines, removes the `.expect("inserted above if absent")` panic path entirely (there's no longer a separate insert-then-fetch step whose invariant could be violated by a future edit), and is the more idiomatic way to express "get or create" on a HashMap in Rust. Not meaningfully a performance change either way, but worth doing for clarity.

crates/bins/src/lib.rs
```diff
@@ -359,6 +376,363 @@ pub fn secret_analysis(
...
+        "[git-history] pass2: attributing {} secret-bearing blobs via 'git log --raw'",
+        targets.len()
+    );
+    let t_walk = std::time::Instant::now();
+
+    let args = [
+        "-c",
+        "core.quotePath=false",
+        "log",
+        "--branches",
```
robertohuertasm-datadog: To consider: A blob that contains a secret but is reachable only from a tag, a remote-tracking branch, a stash, or the reflog is found in pass 1, gets no `introduced_at` in pass 2, and is then silently dropped at emit time (the `let Some(introduced_at) = occurrence.introduced_at else { continue }` guard). For a security tool scanning history, missing secrets that live in tagged releases or remote branches is a meaningful false negative.

Why it matters: users will reasonably assume "scan git history" covers all history, including tags. Consider `--all` (covers refs/tags, refs/remotes, refs/heads) instead of `--branches`, or at minimum document the limitation prominently. 

Since pass 1 already pays to scan every ODB object (including unreachable ones), widening attribution is essentially free and removes the asymmetry.

crates/bins/src/lib.rs
```diff
@@ -359,6 +376,363 @@ pub fn secret_analysis(
     Ok(analyzer_result)
 }
 
+/// Get the `BlobOccurrence` for a (blob, path) pair, creating an empty one if it
```
robertohuertasm-datadog: NIT: This file went from 657 lines to 1247. The entire git history feature could be possibly moved to its own file, as it has a very particular concern. 

We could either move it to: 

- `crates/bins/src/git_history.rs`  or
- a new sub-crate under `crates/secrets` (it's conceptually a secrets product concern)

The former is the lower-friction path.

crates/secrets/src/model/secret_result.rs
```diff
@@ -74,6 +74,21 @@ pub struct SecretResult {
     pub message: String,
     pub priority: RulePriority,
     pub matches: Vec<SecretResultMatch>,
+    #[serde(default)]
```
robertohuertasm-datadog: For your consideration:

`SecretResult` now carries three fields that are always false/None in every non-history call path:

```rs
pub is_git_history_only: bool,
pub introducing_commit_sha: Option<String>,
pub removed_at_sha: Option<String>,
```

The cost is visible: 8 existing test sites had to be updated with `is_git_history_only: false`, `introducing_commit_sha: None`, `removed_at_sha: None` just to satisfy struct construction. Every future test in this crate will carry the same boilerplate forever. The secrets crate model is now permanently coupled to a `git-layer` concept it should know nothing about.

`SecretHistoryInfo` in `sarif_utils.rs` is a direct consequence of this: it exists only to copy those three fields back out of `SecretResult` and ferry them through the `SarifViolation::Secret` enum tuple. It doesn't add any behavior or change any invariant. It's pure re-packaging.

The cleaner model could be a thin wrapper owned by the git history layer (e.g. in the new `git_history.rs` module):

```rs
pub struct HistoricalSecretResult {
    pub inner: SecretResult,
    pub introducing_commit_sha: Oid,   // non-optional here — already filtered
    pub removed_at_sha: Option<Oid>,
}
```

`git_history_secret_analysis` returns `Vec<HistoricalSecretResult>`. `SecretResult` stays clean. `SecretHistoryInfo` disappears entirely. The SARIF layer reads directly from `HistoricalSecretResult`. The `SarifViolation::Secret` 3-tuple goes back to 2 elements. No test churn in the secrets crate.

Note: `introducing_commit_sha` can be typed as `Oid` rather than `Option<String>` in the wrapper, because by the time you construct a `HistoricalSecretResult` you've already confirmed the introducing commit exists (the `let Some(introduced_at) = occurrence.introduced_at else { continue }` guard in `git_history_secret_analysis`). The string conversion happens at SARIF serialization time, not in the model.

crates/cli/src/sarif/sarif_utils.rs
```diff
@@ -940,18 +985,66 @@ pub fn generate_sarif_file(
...
-        sarif_report_metadata,
-        path_metadata,
-    ) {
-        Ok(report) => {
-            Ok(serde_json::to_string(&report).expect("error when getting the SARIF report"))
+    // In normal mode the secrets tool name is the default, and both kinds of findings share a
+    // single concatenated run. In git-history mode the secrets tool name is distinct, and we emit
+    // two separate runs so static-analysis and historic-secret findings are attributed to their own
+    // tool.
+    let report = if sarif_report_metadata.tool_name == DEFAULT_TOOL_NAME {
```
robertohuertasm-datadog: NIT: `SarifReportMetadata.tool_name` was introduced to let the history scan emit a differently-named tool driver ("datadog-static-analyzer-secrets-history"). That's a legitimate label. But line 992 means the value of the label is now also controlling how many SARIF runs are emitted. Those are independent concerns: a future caller could legitimately set a custom tool name without intending to switch to split-run mode, and it would silently fall through to merge_sarif_runs.

`SarifReportMetadata` already has `add_git_info: bool` and `debug: bool` as explicit boolean flags. We could add a new property called `split_runs_by_tool: bool`. 

Ideally, we would be doing here something like:

```rs
let report = if sarif_report_metadata.split_runs_by_tool {
    merge_sarif_runs(...)
} else {
    generate_sarif_report(...)
};
```

This would keep the mode explicit, and the `else` branch would no longer be a footgun for potential future tool names.

With the current approach, any value that isn't the exact default string triggers `merge_sarif_runs`. If someone adds a third tool name for a different product (say, IAST), they get two runs, not one, with no warning. The intent of the `else` is "this is history mode", but it reads as "this is anything other than the standard name."

The problem with the current code isn't that the equality check exists. It's that it's been pushed down into `generate_sarif_file`, which has no business knowing what mode the binary is running in. That function receives finished data and should only worry about how to format it.

So we would be adding something like this in `datadog-static-analyzer.rs:782`:

```rs
SarifReportMetadata {
    // ...
    tool_name: if git_history_only { SECRETS_HISTORY_TOOL_NAME } else { DEFAULT_TOOL_NAME }.to_string(),
    split_runs_by_tool: git_history_only, 
}
```

crates/bins/src/bin/datadog-static-analyzer.rs
MikaYuoadas: Indeed that’s a leftover that no longer serve any purpose. Removed it.

crates/bins/src/lib.rs
MikaYuoadas: Far cleaner, thanks 👍

crates/cli/src/sarif/sarif_utils.rs
MikaYuoadas: Agreed

crates/bins/src/lib.rs
MikaYuoadas: Good catch, I hadn’t thought of non-branch refs 🙈

crates/bins/src/lib.rs
MikaYuoadas: Indeed, move everything to lib.rs to crates/bins/src/git_history.rs 👍

crates/secrets/src/model/secret_result.rs
MikaYuoadas: Good idea, I changed it to match your suggestion.

crates/bins/src/bin/datadog-static-analyzer.rs
```diff
@@ -671,6 +688,23 @@
...
+    // independent product and is left untouched.
+    let mut historic_secrets: Vec<HistoricalSecretResult> = Vec::new();
+    if scan_git_history_only {
+        let history_start = std::time::Instant::now();
+        historic_secrets = git_history_secret_analysis(&configuration, &analysis_options)
+            .context("git history secret analysis failed")?;
+
+        let history_duration = history_start.elapsed().as_secs_f64();
+        println!("Git History Secrets Summary");
+        println!("  Historical secrets found: {}", historic_secrets.len());
```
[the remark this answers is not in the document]
MikaYuoadas: It’s not sensitive information, it’s the summary of the Secret security scan...

crates/cli/src/sarif/sarif_utils.rs
```diff
@@ -465,7 +498,7 @@ fn generate_tool_section(rules: &[SarifRule], options: &SarifGenerationOptions)
     }
 
     let driver: ToolComponent = ToolComponentBuilder::default()
-        .name("datadog-static-analyzer")
+        .name(options.tool_name.clone())
```
robertohuertasm-datadog: NIT: you could avoid the allocation by using `as_str()` instead of `clone` or just passing a ref `&options.tool_name`.

crates/bins/src/git_history.rs
```diff
@@ -0,0 +1,598 @@
...
+
+                // `read_header` reads only the object header (type + size) without
+                // decompressing, so non-blob and oversized objects are filtered cheaply.
+                let (size, kind) = odb.read_header(oid).ok()?;
+                if kind != ObjectType::Blob || size > max_blob_bytes {
+                    return None;
+                }
+                let obj = odb.read(oid).ok()?;
+                // Skip binary blobs; copy text content out so the borrow can be released.
+                std::str::from_utf8(obj.data()).ok().map(str::to_owned)
```
robertohuertasm-datadog: For your consideration:

Pass 1 reads blob content with std::str::from_utf8(obj.data()).ok().map(str::to_owned), so any blob containing even a single invalid UTF-8 byte is dropped entirely and never scanned.

The HEAD scanner falls back to from_utf8_lossy, so it does scan files that contain a secret alongside some non-UTF-8 bytes (e.g. latin1, mixed binary). The history scan will silently miss those same secrets, a false negative that is invisible to the user. The code comment even calls this "Skip binary blobs", but it actually skips all non-UTF-8 content, not just binary.

Why it matters: parity between the HEAD scan and history scan is an implicit contract of this feature ("find secrets we removed"); a secret that was scrubbed from history but lived in a non-UTF-8-clean file won't be reported.

Suggested fix: mirror read_file's behavior, use `String::from_utf8_lossy(obj.data())`. We could still skip obviously-binary blobs via a NUL-byte heuristic, rather than dropping the blob on any invalid byte.

git2 provides `Blob::is_binary()`. You could replace `odb.read(oid)` with:

```rs
let blob = repo.find_blob(oid).ok()?;
if blob.is_binary() {
    return None;
}
let content = String::from_utf8_lossy(blob.content()).into_owned();
```

Performance impact should be modest:

- `find_blob` replaces the existing object read; don’t perform both.
- `is_binary()` examines at most the first 8 KB.
- Lossy conversion already allocates about the same `String` as the current `to_owned()`.
- Skipping blobs classified as binary may save substantial scanner work.

The main caveat is consistency: `Blob::is_binary()` is only usable for Git objects, while the HEAD scanner reads filesystem bytes. If exact parity is required, a shared byte-based binary-detection helper used by both paths would be better. Otherwise, using `git2` only in history still leaves a behavioral difference for blobs classified as binary.

MikaYuoadas: Nice catch. I’m updating it with the lossy path without using `is_binary()` to keep parity with the HEAD scanner.

crates/cli/src/sarif/sarif_utils.rs
MikaYuoadas: That’s fair, I’ll change it.

crates/bins/src/git_history.rs
```diff
@@ -0,0 +1,600 @@
...
+    let repo = Repository::open(&config.source_directory)?;
+
+    // Pass 1: scan all unique blobs.
+    let secret_blobs = scan_all_blobs_for_secrets(&repo, config, options)?;
+    if secret_blobs.is_empty() {
+        return Ok(Vec::new());
+    }
+
+    // Pass 2: attribute the secret-bearing blobs to their (path, commit) pairs.
+    let targets: HashSet<Oid> = secret_blobs.keys().copied().collect();
```
robertohuertasm-datadog, line 353: Not a blocker: more a `document the semantics or file a follow-up` kind of question. 

Pass 2 attributes across `--all` (branches, tags, remotes, stash), but the exclusion `set collect_head_blob_paths` is built only from the current `repo.head()`. So a secret that is live at the tip of another local branch, a remote-tracking branch, a tag, or a stash entry is emitted with `isGitHistoryOnly: true`. That property is inaccurate for content that currently exists on a ref, and consumers keying off `isGitHistoryOnly` may treat a live exposure as a merely-historical one.

Why it matters: the finding's core classification (history-only vs live) can be wrong depending on repo ref layout; it's also the kind of thing that differs between a fresh clone (has remotes) and a working checkout.

Suggested fix: either (a) document explicitly that "history-only" is relative to the current HEAD only, and reflect that in the property/name, or (b) build the exclusion set from all ref tips (`git for-each-ref` / walking all refs) so "history-only" means "not present at any current ref tip". Option (b) matches the `--all` traversal used for attribution.

robertohuertasm-datadog (dismissed):
LGTM! 🚀

crates/bins/src/git_history.rs
MikaYuoadas: Well it is the desired behavior, otherwise any tag would hide secrets at that point in history.
The goal of "git history only" is to surface everything that is not surface by a full scan on `HEAD` so that customer enabling the product on an already existing repo don’t falsely feel safe because nothing is reported on latest `HEAD` scan.
Even branch should be reported as stale branch won’t be scanned (we only scan if there’s a new push event).
Only `HEAD` is guaranteed to be scan at least once on repo onboarding.
I’ll update the flag description to make it clearer what is or isn’t included.

robertohuertasm-datadog (approved):
LGTM! 🚀

PR #938, opened 2026-07-13
[SDSP-485] Support suppressions for secret rules

vhourdel:

Some secrets matched by secret rules aren't actual secrets: they may be test tokens, test card numbers, dummy values, etc.

## What is your solution?

Add support for suppressions, which are natively supported by the secret scanning engine already. This allows users to specify "suppressions" which are some kind of filter on top of the matches, to determine whether the match should be kept or not.

### DIFF
--- a/crates/secrets/src/model/secret_rule.rs
+++ b/crates/secrets/src/model/secret_rule.rs
@@ -456,6 +456,26 @@ impl SecretRuleValidator {
     }
 }
 
+#[derive(Serialize, Deserialize, Debug, Clone, Eq, PartialEq, Default)]
+pub struct SecretRuleSuppressions {
+    #[serde(default)]
+    pub starts_with: Vec<String>,
+    #[serde(default)]
+    pub ends_with: Vec<String>,
+    #[serde(default)]
+    pub exact_match: Vec<String>,
+}
+
+impl From<&SecretRuleSuppressions> for dd_sds::Suppressions {
+    fn from(v: &SecretRuleSuppressions) -> Self {
+        dd_sds::Suppressions {
+            starts_with: v.starts_with.clone(),
+            ends_with: v.ends_with.clone(),
+            exact_match: v.exact_match.clone(),
+        }
+    }
+}
+
 // This is the secret rule exposed by SDS
 #[derive(Clone, Deserialize, Debug, Serialize, Eq, PartialEq)]
 pub struct SecretRule {
@@ -474,6 +494,8 @@ pub struct SecretRule {
     pub pattern_capture_groups: Vec<String>,
     #[serde(default)]
     pub is_supporting_rule: bool,
+    #[serde(default)]
+    pub suppressions: Option<SecretRuleSuppressions>,
 }
 
 impl SecretRule {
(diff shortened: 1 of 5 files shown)


crates/secrets/src/model/secret_rule.rs
```diff
@@ -542,20 +564,105 @@ impl SecretRule {
...
+                }
+
+                // Suppression order does not affect matching behavior, so canonicalize the
+                // lists to avoid invalidating diff-aware results for ordering-only changes.
+                starts_with.sort();
+                ends_with.sort();
+                exact_match.sort();
+
+                digest.push_str(&format!(
+                    ":suppressions:{starts_with:?}:{ends_with:?}:{exact_match:?}",
```
jasonforal: I wouldn't rely on the formatting of the Debug impl (which is what `:?` is) to format the contents.

I would use `starts_with.join` and then use the normal Display impl (`{starts_with}`)

crates/secrets/src/model/secret_rule.rs
```diff
@@ -542,20 +564,105 @@ impl SecretRule {
...
+        }))
+        .generate_diff_aware_digest();
+        let changed_suppressions = rule_with_suppressions(Some(SecretRuleSuppressions {
+            starts_with: vec!["prefix-a".to_string()],
+            ends_with: vec![],
+            exact_match: vec!["exact".to_string()],
+        }))
+        .generate_diff_aware_digest();
+
+        assert_ne!(legacy_digest, with_suppressions);
```
jasonforal: Not sure what value this adds. I would remove this test

crates/secrets/src/model/secret_rule.rs
```diff
@@ -639,12 +746,40 @@ mod tests {
...
             is_supporting_rule: false,
+            suppressions: None,
         };
 
         // Validates that convert_to_sds_ruleconfig doesn't panic and returns a valid config.
         let _config = rule.convert_to_sds_ruleconfig(false);
     }
 
+    #[test]
+    fn test_convert_to_sds_ruleconfig_with_suppressions() {
```
jasonforal: This test isn't testing anything. You should either have an assertion somewhere or remove the test.

crates/secrets/src/scanner.rs
```diff
@@ -608,4 +618,45 @@ mod tests {
         assert_eq!(first.start, Position { line: 2, col: 1 });
         assert!(first.is_suppressed);
     }
+
+    #[test]
+    fn test_find_secrets_with_rule_suppressions() {
```
jasonforal, line 623: Not sure I understand the point of this test -- could you explain?

crates/secrets/src/model/secret_rule.rs
```diff
@@ -456,6 +456,26 @@ impl SecretRuleValidator {
...
+pub struct SecretRuleSuppressions {
+    #[serde(default)]
+    pub starts_with: Vec<String>,
+    #[serde(default)]
+    pub ends_with: Vec<String>,
+    #[serde(default)]
+    pub exact_match: Vec<String>,
+}
+
+impl From<&SecretRuleSuppressions> for dd_sds::Suppressions {
```
jasonforal: It's more idiomatic to impl `From<SecretRuleSuppressions>` and let the caller control cloning if they want to.

crates/secrets/src/scanner.rs
vhourdel: The test is checking that the built scanner from a `SecretRule` that has suppressions configured does in fact suppress matches.

Here the pattern would match both lines in the dummy code provided, but the second line has an `exact_match` in the suppressions that would suppress it. The test makes sure that this is the case and that we indeed match the first line, but not the second.

crates/secrets/src/model/secret_rule.rs
vhourdel: I was more or less using the same pattern as the test above to ensure that converting the rule doesn't panic but I agree this is probably not testing much of anything. I'm removing this

crates/secrets/src/model/secret_rule.rs
vhourdel: Agreed, changed in a628e918

crates/secrets/src/model/secret_rule.rs
vhourdel: Removed in a628e918

crates/secrets/src/model/secret_rule.rs
vhourdel: Removed in a628e918

crates/secrets/src/model/secret_rule.rs
vhourdel: Addressed in a628e918

crates/secrets/src/scanner.rs
jasonforal: My concern is that the behavior it asserts:
```rs
assert_eq!(
    matches.first().unwrap().matches.first().unwrap().start,
    Position { line: 1, col: 1 }
);
```
Is merely testing that SDS's suppression implementation is correct. That belongs in SDS library tests, not here (and in fact, this is already covered by SDS's [test_match_suppression_suppress_half_of_the_matches](https://github.com/DataDog/dd-sensitive-data-scanner/blob/4e53b6ac37ab4b6699f4331868a1c486f0d21147/sds/src/scanner/test/mod.rs#L493))

And additionally, we have no test coverage for what new behavior this PR actually introduces -- that `build_sds_scanner` correctly threads the `SecretRule` suppressions through to SDS. Hypothetically if we (`datadog-static-analyzer`) forgot to pass the suppression to SDS, but SDS were to independently omit the second finding for whatever reason, this test would incorrectly pass.

So what we need is a control case. You'd want to test a `SecretRule { suppressions: None }` and assert `assert_eq!(matches[0].matches.len(), 2);` and then pass in `SecretRule { suppressions: Some(...) }` and `assert_eq!(matches[0].matches.len(), 1);`

And from there we know the delta can only be explained by the change to `suppressions`.

vhourdel: Ah yes that's very fair, thank you for the thorough explanation. I've reworked the test in 0af2092c to match the behavior you are proposing.

jasonforal (approved):

PR #942, opened 2026-07-31
Add secrets config: ignore/only paths, gitignore, generated files, max file size

MikaYuoadas:
## What

Splits `CliConfiguration` into `SastConfiguration`, and `SecretsConfiguration`, breaks `main()` into per-product functions, and gives secrets its own independent configuration-file parser. Adds four `secrets.global-config` options, mirroring the existing `sast` schema:

```yaml
schema-version: v1.5
secrets:
  global-config:
    ignore-paths:
      - "vendor/**"
    only-paths:
      - "src/**"
    use-gitignore: false
    ignore-generated-files: false
    max-file-size-kb: 500
```

## Why

Previously SAST's file-selection settings (paths, gitignore, generated-files, size limits) could silently apply to secrets scanning too, since both products shared one configuration struct and one file-selection code path. There was also no way to configure secrets scanning at all.

## Behavior

- SAST and secrets each build their own file selection and read their own configuration-file section, an invalid section for a disabled product is never even parsed, so it can't fail the other product's scan.
- Git-history secret scanning now uses secrets' own `max-file-size-kb` instead of SAST's (default limit 10MB)
- This groundwork will make it easier to eventually fork the binary into separate SAST-only and secrets-only codebases.


crates/bins/src/bin/datadog-static-analyzer-git-hook.rs
```diff
@@ -425,27 +465,63 @@ fn main() -> Result<()> {
...
+        let platform = initialize_v8(run_config.get_num_threads() as u32);
         _ = v8.insert(platform)
     }
 
     // This must be called _after_ `initialize_v8` (otherwise, PKU-related segfaults on Linux will occur).
     // Stack size is set explicitly to 64MB for the dd-sds scanner's regex matching depth.
     rayon::ThreadPoolBuilder::new()
-        .num_threads(configuration.get_num_threads())
+        .num_threads(run_config.get_num_threads())
         .build_global()?;
```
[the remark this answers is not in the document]
MikaYuoadas: Not changed or introduced by this PRs.

crates/bins/src/bin/datadog-static-analyzer.rs
```diff
@@ -386,343 +463,382 @@
...
+        .iter()
+        .unique_by(|v| v.rule_name.as_str())
+        .count();
+
+    let secrets_duration = secrets_start.elapsed().as_secs_f64();
+
+    println!("Secrets Summary");
+    println!("  Files scanned: {}", secrets_files.len());
+    println!("  Files with secrets: {}", files_with_secrets);
+    println!("  Total secrets: {}", nb_secrets_found);
```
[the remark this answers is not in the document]
MikaYuoadas: `nb_secrets_found` is the secrets scanner output results, not an actual secrets that should be redacted. Not an issue

crates/bins/src/git_history.rs
```diff
@@ -75,13 +78,18 @@ fn collect_head_blob_paths(repo: &Repository) -> HashSet<(Oid, String)> {
...
+    let secrets_rules = &secrets_config.rules;
     let sds_scanner =
-        build_sds_scanner(secrets_rules, config.use_debug).map_err(|e| anyhow::anyhow!(e))?;
-    let max_blob_bytes = (config.max_file_size_kb * 1024) as usize;
+        build_sds_scanner(secrets_rules, run_config.use_debug).map_err(|e| anyhow::anyhow!(e))?;
+    // Unbounded unless the user opts into a limit via secrets' own `max-file-size-kb`.
+    let max_blob_bytes = secrets_config
+        .max_file_size_kb
+        .map(|kb| kb as usize * 1024)
+        .unwrap_or(usize::MAX);
```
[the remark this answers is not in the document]
MikaYuoadas: It’s "silently" puting a limit in the petabytes to a single file. I think we should be fine...

crates/bins/src/bin/datadog-static-analyzer-git-hook.rs
```diff
@@ -9,24 +9,27 @@ use cli::constants::{
...
 use common::analysis_options::AnalysisOptions;
-use common::model::diff_aware::DiffAware;
 use datadog_static_analyzer::{secret_analysis, static_analysis};
 use getopts::Options;
 use git2::Repository;
 use itertools::Itertools;
 use kernel::analysis::ddsa_lib::v8_platform::{initialize_v8, Initialized, V8Platform};
 use kernel::classifiers::ArtifactClassification;
 use kernel::config::common::{ConfigMethod, PathConfig};
-use kernel::config::file_v1;
```
jasonforal, line 29: Please don't change the import to remove this qualifier. It's done to disambiguate from a "legacy" config.

crates/static-analysis-kernel/src/rule_config.rs
```diff
@@ -16,15 +16,16 @@ pub struct RuleConfigProvider {
 }
 
 impl RuleConfigProvider {
-    pub fn from_config(cfg: &file_v1::ConfigFile) -> RuleConfigProvider {
+    /// Builds a provider from SAST's own configuration section. Secrets doesn't have (or need)
+    /// an equivalent, since it doesn't support per-rule path/argument/severity overrides.
+    pub fn from_config(sast_config: Option<&file_v1::SastConfig>) -> RuleConfigProvider {
```
jasonforal: Please don't change the function signature to accept an `Option<T>`. It's a code smell

By doing this, you are encoding business logic ("doesn't support per-rule path/argument/severity overrides") into a `RuleConfigProvider`, which shouldn't know anything about that.

If you need to split these two, change it to `from_sast_config`, and add `from_secrets_config`.

crates/cli/src/config_file.rs
```diff
@@ -56,12 +96,18 @@ fn get_local_config(base_path: &Path) -> anyhow::Result<Option<(file_v1::ConfigF
 /// - If the user is a Datadog user (e.g. with API keys), we fetch the remote configuration
 ///   and merge it
 /// - If not, we just return the configuration
+///
+/// `parse_sast`/`parse_secrets` should match whether each product is enabled: a product's
+/// section is only ever read and validated when that product is enabled. An invalid section for
+/// a disabled product never causes a failure, because it's never parsed at all.
 pub fn get_config(
```
jasonforal, line 122: Hmm, we're separating some functions but not others...it doesn't feel consistent. The code smell here is needing to thread `parse_sast` and `parse_secrets` in as parameters.

Why not `get_config(...) -> Result<(String, ConfigMethod)>`? Small, contained

And then the consumer can do whatever they want with the (merged) config yaml

crates/cli/src/config_file.rs
```diff
@@ -209,8 +286,41 @@ sast:
...
                 fs::write(&file_path, content).unwrap();
-                assert!(get_local_config(test_dir.path()).unwrap().is_none());
+                assert!(get_local_config(test_dir.path(), true, false)
+                    .unwrap()
+                    .is_none());
             }
         }
     }
+
+    /// Secrets' own `global-config.ignore-paths`/`only-paths` are parsed independently of SAST.
```
jasonforal: Not sure I fully understand this test, but if the intent is to ensure

> Secrets' own `global-config.ignore-paths`/`only-paths` are parsed independently of SAST.

This doesn't do that. It's only checking `secrets`. I would expect such a test to include both `secrets` and `sast`.

But I am not sure I understand the point of it -- the type system already guarantees this statically, no?

crates/cli/src/model/cli_configuration.rs
```diff
@@ -74,31 +51,40 @@ impl DiffAware for CliConfiguration {
             self.ignore_gitignore,
             rules_string.join(","),
             self.max_file_size_kb,
-            self.source_subdirectories.join(","),
```
jasonforal, line 77: Why remove this?

crates/bins/src/git_history.rs
jasonforal: Seems like a footgun. Why not have a sensible default?

crates/cli/src/config_file.rs
```diff
@@ -6,10 +6,31 @@ use crate::datadog_utils::{
...
 use kernel::config::{file_legacy, file_v1};
 use kernel::utils::decode_base64_string;
+use secrets::config::file_v1::SecretsConfig;
 use std::path::Path;
 
+/// SAST's and secrets' configuration, read independently of one another from the same
+/// underlying configuration file: each product owns its own parser, so an invalid section in
+/// one never affects the other.
+#[derive(Debug, Clone, Default, PartialEq)]
+pub struct ConfigFile {
```
jasonforal: Why have this? If the intent is that the products exist independently of each other, then I'm not sure why we'd want to keep them together like this

crates/cli/src/file_utils.rs
```diff
@@ -776,6 +793,140 @@ mod tests {
         assert_not_contains_files!(&base_path, files, ["src/b/main.rs", "test/a/main.rs"]);
     }
 
+    // Two products with different `ignore_gitignore` settings must get independent results:
```
jasonforal: I don't know off the top of my head, but where did we used to test all this behavior? This just seems like a surprising amount of tests for what is a refactor. (Possible I missed where these used to be)

Also

> // Two products with different `ignore_gitignore` settings must get independent results:

That's not what's being tested. Is this just an LLM-prompt-leaking-into-code-thing or what was the intent here?

crates/bins/src/bin/datadog-static-analyzer-git-hook.rs
MikaYuoadas: I’m not sure what you mean.
IIUC you meant keeping `use kernel::config::file_v1` instead of things like `use kernel::config::file_v1::SastConfig;` so usage in the code are prefixed with `file_v1` (e.g. here `file_v1::SastConfig`?

If so, I’ve made sure all sast config import use the first one to keep the prefix.

crates/static-analysis-kernel/src/rule_config.rs
MikaYuoadas: Agreed 👍 
I added `from_sast_config(&file_v1::SastConfig)` without the `Option` to explicitely show it’s SAST-related, but I did not add a `from_secrets_config()`, as Secrets has no per-rule path, argument or severity overrides, it would be a dead constructor.

crates/cli/src/config_file.rs
MikaYuoadas: Done!

Now `get_config()` just returns the content and parses nothing, so each binary parses only the products it has enabled.

One consequence of this refactor is that schema dispatch now follows the document's schema-version key through `parse_any_schema_yaml()` instead of the file's name (like what the kernel, the server crate and the IDE path were already doing).
This means a `code-security.datadog.yaml` file holding a legacy-shaped config would now parses as legacy, and a `static-analysis.datadog.yaml` that sets `schema-version: v1.0` now parses as v1.

crates/cli/src/config_file.rs
MikaYuoadas: Yes, indeed the comment was no longer matching the code after a few rebase and refactor I did. 😞

I have deleted it, the parsing itself is already covered in the secrets crate.

crates/bins/src/git_history.rs
MikaYuoadas: Agreed! I left it at that because previously Secrets file size was just unbounded.
I made the default 10MB instead of 200kB like SAST because we may have larger file containing secrets (maybe notebook, plan file, etc... that can be quite large).

crates/cli/src/config_file.rs
MikaYuoadas: Agreed, I removed it and each product has its own independant struct.

crates/cli/src/file_utils.rs
MikaYuoadas: I added a few test (I counted 15): some to make sure I didn’t break anything during the refactor, and a few to test the new parse path for secrets.

As for the comments, they indeed fell out of sync with the actual cost so I updated them (it’s only testing both product parse path, not that they don’t "leak" into one another).

crates/cli/src/model/cli_configuration.rs
MikaYuoadas: It’s not really removed, `source_subdirectories` is still in the final digest (see `generate_full_diff_aware_digest()`) but we couldn’t access it from `SastConfiguration` as it now as been moved to `RunConfiguration` struct instead.

crates/cli/src/config_file.rs
jasonforal: Ah, we can't introduce a regression here. What about having `get_local_config` return the matched file path and then repurpose the existing logic instead of using `parse_any_schema_yaml`?

crates/cli/src/model/sast_configuration.rs
```diff
@@ -74,31 +51,40 @@ impl DiffAware for CliConfiguration {
...
     }
 }
 
-impl CliConfiguration {
-    pub fn get_num_threads(&self) -> usize {
-        get_num_threads_to_use(self.num_cpus)
+impl SastConfiguration {
+    /// The digest actually used to identify a diff-aware run: `SastConfiguration`'s own digest plus
+    /// `RunConfiguration.source_subdirectories`, so that changing the analyzed subdirectory scope
+    /// invalidates a stale cached diff-aware result. Use this (not `SastConfiguration::generate_diff_aware_digest`
```
jasonforal: > Use this (not `SastConfiguration::generate_diff_aware_digest` directly) anywhere the digest is used to identify or cache a run

The point of the trait is encapsulation, so needing to do this is an implementation smell. Seems pretty easy to miss/forget and introduce a bug.

Seems like implementing `CliConfigurationSast` and `CliConfigurationSecrets` with each storing the `RunConfiguration` would be cleaner.

crates/cli/src/model/run_configuration.rs
```diff
@@ -0,0 +1,25 @@
...
+    pub output_format: OutputFormat, // SARIF or JSON
+    pub output_file: String,
+    pub num_cpus: usize, // of cpus to use for parallelism
+    pub use_staging: bool,
+    pub static_analysis_enabled: bool,
+    pub secrets_enabled: bool,
+}
+
+impl RunConfiguration {
+    pub fn get_num_threads(&self) -> usize {
```
jasonforal: From a design perspective, this is a bit odd. This is a struct intended to be a model, and yet this is encoding opinionated business logic.

crates/bins/src/git_history.rs
```diff
@@ -2,10 +2,13 @@
 // (pass 1), then attribute only the secret-bearing blobs back to their
 // introducing/removal commits via a single `git log` pass (pass 2).
 
-use cli::model::cli_configuration::CliConfiguration;
+use cli::model::run_configuration::RunConfiguration;
+use cli::model::secrets_configuration::SecretsConfiguration;
 use cli::sarif::sarif_utils::HistoricalSecretResult;
 use common::analysis_options::AnalysisOptions;
 use git2::{ObjectType, Oid, Repository, TreeWalkMode, TreeWalkResult};
+#[cfg(test)]
```
jasonforal: Please import this within the test mod directly rather than relying on `super::*`

crates/common/src/model/config_method.rs
```diff
@@ -0,0 +1,104 @@
+// Unless explicitly stated otherwise all files in this repository are licensed under the Apache License, Version 2.0.
```
jasonforal, line 1: I don't think we want to try to pull out this into shared code. I think in general we discussed moving towards decoupling the two implementations.

The design smell here is having a `Legacy` enum variant in `YamlSchemaVersion`, but that is purely a SAST concern.

crates/common/src/model/path_config.rs
```diff
@@ -0,0 +1,185 @@
...
+                Some(only) => only.iter().any(|pattern| pattern.matches(file_name)),
+            }
+    }
+}
+
+/// A combination of `only-paths` and `ignore-paths` fields, intended to be inlined via [`serde(flatten)`].
+#[derive(Debug, Clone, Deserialize, Serialize, Default, PartialEq, Eq)]
+#[serde(rename_all = "kebab-case")]
+#[serde(deny_unknown_fields)]
+pub struct YamlPathConfig {
```
jasonforal: `YamlPathConfig` should not be here. This is a model for a PathConfig whereas yaml is a serialization/implementation detail. They shouldn't be coupled.

I think we generally agreed the configuration file would be separate between SAST/Secrets

crates/static-analysis-kernel/src/config/file_v1.rs
```diff
@@ -347,51 +348,6 @@ impl From<YamlRuleConfig> for RuleConfig {
     }
 }
```
jasonforal, line 347: See previous

crates/cli/src/model/run_configuration.rs
MikaYuoadas: I’m not sure I follow what you mean here. What’s the opinionated business logic here? The `get_num_threads_to_use()`?
Anyway I didn’t introduce this function, it was already present on the `CliConfiguration` impl (I just split the SAST specific part to `SastConfiguration` and renamed it `RunConfiguration`).

crates/common/src/model/config_method.rs
MikaYuoadas: I agree that the `Legacy` is SAST specific, but `ConfigMethod` and `YamlSchemaVersion` are a "unified code security config" matters so it makes sense for them to be shared, doesn’t it?
I’m not sure what the right alternative would be.

crates/bins/src/git_history.rs
MikaYuoadas: Agreed, moving it 👍

crates/cli/src/config_file.rs
MikaYuoadas: Well I simply made `get_local_config` return a `ConfigFileSchema` enum with either `CodeSecurity` or `Legacy` value.
It’s then used instead of using `parse_any_schema_yaml()` by each product parsing function.

crates/common/src/model/config_method.rs
MikaYuoadas: I’ve moved `YamlSchemaVersion` back to SAST’s `crates/static-analysis-kernel/src/config/common.rs` and secret will have its own (very basic string matching) version parsing that doesn’t knows about the legacy one.
I’m not sure about `ConfigMethod`, imho it’s shared as it relates to the unified code security config mechanism (merging local vs remote config).
For now I’ve left it alone in `crates/common/src/model/config_method.rs`, let me know if you have a better idea.

crates/common/src/model/path_config.rs
MikaYuoadas: Well the name `YamlPathConfig` isn’t by me, I’ve just moved it from its initial `crates/static-analysis-kernel/src/config/file_v1.rs` location.

But I agree on the decoupling point made, so I’ve moved it back to its original location (`crates/static-analysis-kernel/src/config/file_v1.rs`) and added secret’s own copy in `crates/secrets/src/config/file_v1.rs`.

crates/bins/src/git_history.rs
MikaYuoadas: 👍 Moved it.

crates/static-analysis-kernel/src/config/file_v1.rs
MikaYuoadas: yes, I’ve moved it back here.

crates/cli/src/model/sast_configuration.rs
MikaYuoadas: I’ve created both `CliConfigurationSast` and `CliConfigurationSecrets` in their respective crate so that removes the need for this `generate_full_diff_aware_digest()` hackish wrapper.
Let me know if that’s what you had in mind.

crates/secrets/src/config/file_v1.rs
```diff
@@ -88,11 +87,21 @@ impl From<YamlSecretsConfig> for SecretsConfig {
...
+    schema_version: Option<String>,
     #[serde(default)]
     secrets: Option<serde_yaml::Value>,
 }
 
+/// Whether a schema version names a Code Security `v1.x` document.
+///
+/// Matched as a string on purpose: secrets has no legacy schema of its own, so it does not depend
+/// on SAST's schema-version type to answer this.
+fn is_code_security_v1(schema_version: &str) -> bool {
```
jasonforal: This isn't correct, though, because Secrets config wasn't introduced until `v1.5`. Up until that point, your parser should error.

Why not just copy paste `YamlSchemaVersion` from the SAST implementation and remove the code for the legacy parser?

crates/secrets/src/config/file_v1.rs
```diff
@@ -0,0 +1,310 @@
...
+
+/// The minimal top-level shape secrets needs to read out of a Code Security v1.x configuration
+/// file: the schema version (to know whether the file is a schema secrets understands) and the
+/// `secrets` section itself (left unparsed until we know whether we should validate it).
+///
+/// This deliberately does *not* `deny_unknown_fields`: this same YAML may also contain `sast`,
+/// `iac`, `sca`, `iast` keys that secrets don't care about.
+#[derive(Debug, Default, Deserialize)]
+#[serde(rename_all = "kebab-case")]
+struct TopLevel {
```
jasonforal: This breaks the schema contract. A parser _must_ reject any top level keys that aren't in the major v1 specification.

Why don't you copy/paste the SAST implementation and remove anything SAST specific? We're in a weird place now where you are mixing your own implementations with existing concepts, and I think we need to step back and re-evaluate the strategy

MikaYuoadas: I’ve made a first cleanup by just duplicating sast original file_v1.rs and common.rs to the secrets::config crate and just adapting to secret use case.

We’ll discuss later today what we want to achieve exactly in this PR.

crates/secrets/src/config/file_v1.rs
MikaYuoadas: Fixed (see next comment’s response)

crates/bins/src/bin/datadog-static-analyzer.rs
```diff
@@ -386,343 +482,374 @@ fn main() -> Result<()> {
...
+            max_file_size_kb: Some(secrets_config.max_file_size_kb),
+        },
+        args.use_debug,
+    )
+    .context("unable to get the list of files to analyze for secrets")
+}
 
+/// Initialize V8 (only needed for SAST) and the shared rayon thread pool. Must run in this
+/// order: V8 first, then the thread pool (otherwise, PKU-related segfaults on Linux will occur).
+fn build_runtime(run_config: &RunConfiguration) -> Result<Option<V8Platform<Initialized>>> {
```
jasonforal: Would you please leave the v8 initialization code completely untouched (i.e. don't move it to a separate function)?

We don't have automated test coverage for Xeon 1st/2nd gen CPUs, so I would rather not make any changes (as simple as they may seem) since the consequence is segfaults

crates/cli/src/model/run_configuration.rs
jasonforal: This is the business logic:

https://github.com/DataDog/datadog-static-analyzer/blob/501172b2b14f1c1b892e0b735bf63bb08d3547c6/crates/cli/src/utils.rs#L19-L22

At higher core counts (32+) this silently removes ~4 cores unnecessarily.

So all I was saying is that if the _caller_ wants to enforce that logic, they can do so. That is, they can call it manually themselves:
```rs
RunConfiguration { num_cpus: get_num_threads_to_use(32) }
```

 But `RunConfiguration` itself should just store whatever `num_cpus` the caller requested, NOT apply its own logic on top of the caller's request.

Just a better software design in general to keep stuff like this as "plain ol' data"

crates/secrets/src/config/file_v1.rs
```diff
@@ -416,21 +264,24 @@ pub fn parse_yaml(config_contents: &str) -> Result<YamlConfigFile, ParseError> {
...
         YamlSchemaVersion::MajorMinor((1, minor)) => {
-            let mut sast: Option<YamlSastConfig> = None;
+            let mut secrets: Option<YamlSecretsConfig> = None;
             match minor {
-                0.. => {
-                    if let Some(value) = base.sast {
-                        let config: YamlSastConfigMinor0 =
+                // The `secrets` section was introduced in v1.5. Earlier v1.x schemas have no
+                // secrets configuration to read.
+                ..5 => {}
```
jasonforal: I think this introduces a silent acceptance (but no effect) if a secrets config is passed into a v1.0 - v1.4, right?

Any fields under `secrets` from minor 0 - 4 should be rejected loudly as an error.

I think the elegant way to model this is to define a `YamlSecretsConfigMinor0` like
```rs
#[derive(Debug, Deserialize)]
#[serde(deny_unknown_fields)]
pub struct YamlSecretsConfigMinor0;
```
And then on the match arm `..5`, try to deserialize into it (which we know will always fail).

(Note that with this strategy, `YamlSecretsConfigMinor0` _doesn't_ need to go into the `YamlSecretsConfig` enum -- we're merely just using this to get all the rich error messages from serde when it rejects an unexpected field)

crates/bins/src/bin/datadog-static-analyzer.rs
MikaYuoadas: Yes, on it 👍

crates/secrets/src/config/file_v1.rs
MikaYuoadas: I added the `..5` to satisfy the compiler but indeed I didn’t realize that it meant it was now silently allowing anything to be put in the secrets section. 🤦 
That looks like an elegant way to solve that, I’ll implement it as suggested, thanks. 🙏

crates/cli/src/model/run_configuration.rs
MikaYuoadas: Ok got it.
I left it in `RunConfiguration` because it was already done that way in the old `CliConfiguration` struct.
But that’s the right moment to fix that. 👍

jasonforal (approved):

PR #943, opened 2026-08-09
Compute git-history secret fingerprints from historical blob content

MikaYuoadas:
## What

Compute the SARIF `DATADOG_FINGERPRINT` for history-only secrets (present in git history but removed at HEAD) from the historical blob content, instead of re-reading the HEAD working tree.

## Why

For a removed secret, the file or line no longer exists at HEAD, so the fingerprint came out empty and downstream deduplication dropped the finding. Sourcing it from the historical blob with the same hashing algorithm yields a stable, non-empty fingerprint so these findings open correctly.

## Testing

- New regression test verifying a history-only secret gets a non-empty fingerprint matching the HEAD algorithm.
- Existing git-history pipeline test extended to assert the attached blob content lines up with the match position.
- Build, tests, fmt, and clippy pass.

### DIFF
--- a/crates/bins/src/git_history.rs
+++ b/crates/bins/src/git_history.rs
@@ -72,7 +72,7 @@ fn collect_head_blob_paths(repo: &Repository) -> HashSet<(Oid, String)> {
 /// content for free: there is no commit or tree walk. Every unique blob is scanned
 /// (including blobs whose content also exists at HEAD): a blob present at HEAD may
 /// also live at a historical path, and HEAD exclusion is applied per-(blob, path) in
-/// pass 2. Returns the blobs that contained at least one secret, keyed by blob OID.
+/// pass 2. Returns the secrets found, keyed by blob OID.
 fn scan_all_blobs_for_secrets(
     repo: &Repository,
     cli_config: CliConfigurationSecrets<'_>,
@@ -389,6 +389,7 @@ pub fn git_history_secret_analysis(
                     inner: secret.clone_with_path(path),
                     introducing_commit_sha: introduced_at,
                     removed_at_sha: occurrence.removed_at,
+                    blob_oid: *blob_oid,
                 });
             }
         }
@@ -609,6 +610,18 @@ mod git_history_tests {
         assert!(
             results.iter().all(|x| x.inner.filename != "keep.txt"),
             "a secret still present at HEAD must not be reported as historical"
+        );
+        assert!(
+            !r.blob_oid.is_zero(),
+            "historical result must carry its blob OID"
+        );
+        let blob = repo
+            .find_blob(r.blob_oid)
+            .expect("historical blob present in ODB");
+        assert_eq!(
+            std::str::from_utf8(blob.content()).unwrap(),
+            "SECRETVALUE123\n",
+            "blob_oid must point to the historical (pre-scrub) secret content"
         );
     }
 
--- a/crates/cli/src/sarif/sarif_utils.rs
+++ b/crates/cli/src/sarif/sarif_utils.rs
@@ -30,7 +30,7 @@ use serde_sarif::sarif::{
     SarifBuilder, SuppressionBuilder, Tool, ToolBuilder, ToolComponent, ToolComponentBuilder,
 };
 
-use crate::file_utils::get_fingerprint_for_violation;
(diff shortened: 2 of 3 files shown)


crates/bins/src/git_history.rs
```diff
@@ -12,12 +12,17 @@ use secrets::scanner::{build_sds_scanner, find_secrets};
...
+use std::sync::Arc;
 
 /// Placeholder path used when scanning a raw blob whose path is not yet known.
 /// The secrets scanner only uses the filename for log/error messages, not for
 /// matching, so any stable value works here.
 const GIT_HISTORY_BLOB_PLACEHOLDER: &str = "<git-history-blob>";
 
+/// A secret-bearing blob: its decoded content (shared, since one blob fans out
+/// into many (path, match) results) alongside every secret matched in it.
+type SecretBlob = (Arc<str>, Vec<SecretResult>);
```
jasonforal: This is keeping the entire blob in memory for the entirety of the scanning process. I would imagine if a rule is prone to overmatching, this could bloat process memory very quickly.

Have you done any capacity estimation to estimate the expected worst case memory bloat this PR would introduce?

MikaYuoadas: Well I didn’t.
But looking at current usage I found 99.9% of scan with < 1000 secrets findings (a few outliers up to 8k findings), so using an average file size of 10kb we would be adding an extra 10MB memory.
But I updated the PR to instead only keep the Oid and re-read the content from the repo at the end when generating the SARIF report like the regular SAST and Secrets scan were doing.

crates/bins/src/git_history.rs
```diff
@@ -188,6 +191,9 @@ fn scan_all_blobs_for_secrets(
                 None
             } else {
                 secret_blobs.fetch_add(1, Ordering::Relaxed);
+                // Drop the blob content here: the fingerprint is derived later by
```
jasonforal: Please remove this and other LLM-y comments that are just leaking iteration prompt (like `"The blob content itself is intentionally not retained:"` from the function's doc comment)

Just adds noise

jasonforal, line 194: Please remove this and other LLM-y comments that are just leaking the iteration prompt (like `"The blob content itself is intentionally not retained:"` from the function's doc comment)

Just adds noise

crates/cli/src/sarif/sarif_utils.rs
```diff
@@ -778,13 +781,26 @@ fn generate_results(
...
-                        violation,
-                        Path::new(options.repository_directory.as_str()),
-                        Path::new(rule_result.slash_path_str().as_ref()),
-                        options.debug,
-                    );
+                    let fingerprint_option = match historical {
+                        Some(h) => options.git_repo.as_ref().and_then(|repo| {
+                            let odb = repo.odb().ok()?;
+                            let blob = odb.read(h.blob_oid).ok()?;
+                            let content = String::from_utf8_lossy(blob.data());
```
jasonforal, line 788: This is introducing a behavior regression -- previously, `get_fingerprint_for_violation` would ignore files that aren't valid UTF-8:

https://github.com/DataDog/datadog-static-analyzer/blob/69ccd4839ca270364afe3e41c9e630e54ba3d039/crates/cli/src/file_utils.rs#L409-L417

However, here, you are doing a lossy conversion. If this isn't handled consistently elsewhere in the codebase, (which I imagine it isn't), this will introduce subtle bugs like silently shifting column numbers.

What is the Secrets product's stance on non UTF-8?

crates/bins/src/git_history.rs
MikaYuoadas: Agreed, I didn’t scrub it enough, I did a more thorough cleanup pass. 🧹

crates/cli/src/sarif/sarif_utils.rs
MikaYuoadas: Hmmm, Secrets product don’t really have a stance on that afaik, it was just following along with SAST.
Anyway we shouldn’t introduce any regression here, git history scan should behave like sast and secret HEAD-scans, so I’ll keep the previous behavior and I’ll add a note to myself to revisit what we want secrets’ stance to be when we do the cli split between SAST and Secrets. 🤔

jasonforal (approved):

PR #945, opened 2026-08-13
Bump SDS to include change to match CamelCase keywords

zchelseal:
`datadog-static-analyzer` pins `dd-sensitive-data-scanner` to `b2dca51` (2026-06-18), which predates https://github.com/DataDog/dd-sensitive-data-scanner/pull/369. 

Bumping this to include the change of improved matching with camelcase keywords (ie. a rule keyed on `access key` now also matches `accessKey`; not case-sensitive).

Also pulling in https://github.com/DataDog/dd-sensitive-data-scanner/pull/378, which is not yet wired to any rule used by `datadog-static-analyzer` and should have no impact today (needed only if a rule later opts into it).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -456,7 +456,16 @@ version = "0.5.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "0700ddab506f33b20a03b13996eccd309a48e5ff77d0d95926aa0210fb4e95f1"
 dependencies = [
- "bit-vec",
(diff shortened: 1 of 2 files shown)


fuchsnj (dismissed):

Cargo.lock
zchelseal, line 1: additional dependencies here are from https://github.com/DataDog/dd-sensitive-data-scanner/pull/378

fuchsnj (approved):

juli1 (approved):

PR #950, opened 2026-08-25
Ignore an invalid remote configuration (fix regression from 0.9.2)

### LINKED ISSUE #949: [Regression] v0.9.2: invalid SAST configuration
**Describe the bug**

After [0.9.2](https://github.com/DataDog/datadog-static-analyzer/releases/tag/0.9.2) (includes [#942](https://github.com/DataDog/datadog-static-analyzer/pull/942)), `datadog-static-analyzer` fails immediately with:

```text
Starting Static Analysis
Error: invalid SAST configuration: missing field 'schema-version'
Error: Process completed with exit code 50.
```

This started on CI with **no repo or workflow change**. The GitHub Action is pinned (`DataDog/datadog-static-analyzer-github-action@8340f18875fcefca86844b5f947ce2431387e552`, v3.0.0), but that Action still installs the analyzer from `releases/latest`, so 0.9.2 was picked up automatically.

The repo has **no** local `static-analysis.datadog.yml` or `code-security.datadog.yml`. The Action is given `DD_API_KEY` / `DD_APP_KEY`, so config comes from the Datadog remote config API.

Exit code 50 is `EXIT_CODE_INVALID_CONFIGURATION`. The error is printed here:

https://github.com/DataDog/datadog-static-analyzer/blob/0.9.2/crates/bins/src/bin/datadog-static-analyzer.rs

```rust
parse_sast_config(config_contents, schema).unwrap_or_else(|err| {
    eprintln!("Error: invalid SAST configuration: {err}");
    exit(EXIT_CODE_INVALID_CONFIGURATION)
})
```

**To Reproduce**

1. Use `DataDog/datadog-static-analyzer-github-action@v3.0.0` (or the SHA above).
2. Do **not** add a local `code-security.datadog.yml` / `static-analysis.datadog.yml`.
3. Pass Datadog API credentials (`dd_api_key`, `dd_app_key`, `dd_site`) so remote config is fetched.
4. Enable SAST, disable secrets, for example:

```yaml
- uses: DataDog/datadog-static-analyzer-github-action@8340f18875fcefca86844b5f947ce2431387e552 # v3.0.0
  with:
    dd_api_key: ${{ secrets.DD_API_KEY }}
    dd_app_key: ${{ secrets.DD_APP_KEY }}
    dd_site: datadoghq.eu
    secrets_enabled: false
    static_analysis_enabled: true
    cpu_count: 2
```

5. Run after 0.9.2 is `latest`. The **Install datadog-static-analyzer** step succeeds; **Run static analysis** fails in ~1s with the error above.

This was previously green on 0.9.1 with the same Action pin, same repo, and same remote-config setup.

**Expected behavior**

- Remote config without a Code Security `schema-version` should not fail the run.
- 0.9.1 parsed remote YAML with `file_v1::parse_yaml` and, on parse failure, **fell back** to local config or to no config + default rulesets.
- 0.9.2 returns remote YAML unparsed and always treats it as `ConfigFileSchema::CodeSecurity`, which requires `schema-version`. There is no fallback.
- The client still requests remote config with `schema_version=v1` (`QUERY_PARAM_SCHEMA_VERSION`). In this codebase `v1` is the **legacy** schema; Code Security is `v1.0` / `v1.5`. If the backend still returns legacy YAML (no `schema-version`, or `schema-version: v1`), 0.9.2 will fail exactly like this.
- 0.9.2 release does not report this breaking change
- Version pinning of https://github.com/chamelio-ai/datadog-static-analyzer-github-action is unstable and breaks in a situation like this one.

**Screenshots**

Image

**Additional context**

- Regression appears to come from the config-parser split in [#942](https://github.com/DataDog/datadog-static-analyzer/pull/942) (shipped in 0.9.2), not from the new `secrets.global-config` path options.
- Relevant 0.9.2 comment in `crates/cli/src/config_file.rs`:

  ```rust
  // The backend always serves the Code Security schema, whatever the local file used.
  Ok(Some((text, ConfigFileSchema::CodeSecurity, config_method)))
  ```

- In 0.9.1, a failed remote parse did **not** exit 50:

  ```rust
  let res = file_v1::parse_yaml(&text).inspect_err(|err| { /* fallback */ });
  let Ok(remote_yaml) = res else {
      return Ok(local_config.map(|c| (c, ConfigMethod::File)));
  };
  ```

---

MikaYuoadas:
## What

Restores the 0.9.1 fallback for configuration served by the backend. A payload that is not a usable Code Security document is ignored instead of failing the run.

0.9.2 parsed the served payload once, for both products, always as a Code Security document. Anything else failed the run, including the empty payload returned when there is nothing to merge.

Fixes #949

## How

Each product reads its own configuration, with its own parser, and only when that product is enabled. One product's section can no longer affect the other's run.

## Validation

0.9.1, 0.9.2 and this build were run side by side over 46 scenarios: both file names, valid and invalid `sast` and `secrets` sections including mixed ones, every combination of enabled products, and configuration coming from a local file, from the backend, or both. Served payloads covered the empty response, a legacy document, invalid YAML and an unknown major version.

Then end to end against a real backend across six states: no configuration at all, `sast` only, `secrets` only, both, a `secrets` section below the version that introduced it, and a forward minor version.

This build matches 0.9.1 everywhere 0.9.1 has a meaningful baseline, and a `secrets` configuration is applied as intended.


robertohuertasm-datadog (approved):
The parse isolation and remote fallback look right, and the file-level copies **make sense if the follow-up is a split into separate SAST and secrets codebases**. No ask to deduplicate this.

**Two leftovers** in the combined binary, **not blockers**:

- Double fetch + one config method line. When both products run (--enable-secrets true), each get_config hits the backend on its own, then sast_config_method.or(secrets_config_method) prints a single story. After the split, each process has one product, so no .or() and no two POSTs. Until then this is a wart on combined runs only.

- Four match/exit blocks. Same load-or-die shape is inlined twice in datadog-static-analyzer and twice in the git hook. After the split each binary keeps one. Deleting load_config_file is consistent with not leaving a shared helper to untangle later.

**Fine to merge as the pre-split shape.**

MikaYuoadas:
> The parse isolation and remote fallback look right, and the file-level copies **make sense if the follow-up is a split into separate SAST and secrets codebases**. No ask to deduplicate this.
> 
> **Two leftovers** in the combined binary, **not blockers**:
> 
> * Double fetch + one config method line. When both products run (--enable-secrets true), each get_config hits the backend on its own, then sast_config_method.or(secrets_config_method) prints a single story. After the split, each process has one product, so no .or() and no two POSTs. Until then this is a wart on combined runs only.
> * Four match/exit blocks. Same load-or-die shape is inlined twice in datadog-static-analyzer and twice in the git hook. After the split each binary keeps one. Deleting load_config_file is consistent with not leaving a shared helper to untangle later.
> 
> **Fine to merge as the pre-split shape.**

@robertohuertasm-datadog Agreed with all your point.
It’s inelegant but will be a bandaid while we wait for the split to clean this up.

PR #951, opened 2026-08-27
chore(toolchain): add rust-analyzer as a toolchain component

robertohuertasm-datadog:

When using Cursor or VS Code Rust Analyzer won't work and you'll see something like this

image

## What is your solution?

 Added rust-analyzer to rust-toolchain.toml so rustup installs a language server matching the pinned 1.91.1 compiler

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,3 +1,4 @@
 [toolchain]
 channel = "1.91.1"
 profile = "default"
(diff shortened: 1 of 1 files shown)


whitemerch (approved):

MikaYuoadas (approved):

ChouraquiBen (approved):
LGTM

PR #956, opened 2026-09-04
fix(ci): give reusable workflows unique concurrency groups

MikaYuoadas:
## What & why
`test-rules`, `docker-build`, and `integration-tests` are called as reusable workflows from `release.yml`. In a reusable call, `github.workflow` resolves to the caller's name, so all three computed the same concurrency group (`Release new version-<tag>`) and cancelled each other mid-release.

## Fix
Hardcode the concurrency group name per workflow file so it no longer depends on which workflow calls it.
Also use `queue: max` instead of `cancel-in-progress: true` so it worst case to just running them sequentially.

### DIFF
--- a/.github/workflows/coverage.yaml.yml
+++ b/.github/workflows/coverage.yaml.yml
@@ -7,8 +7,8 @@ on:
       - main
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.head_ref || github.ref_name }}
-  cancel-in-progress: true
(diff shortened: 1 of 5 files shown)


ChouraquiBen (dismissed):
LGTM

juli1 (dismissed):

ChouraquiBen (dismissed):
LGTM

ChouraquiBen (approved):

PR #958, opened 2026-09-04
Move AST secret filtering into find_secrets; dedupe language detection

juli1:
Follow-up from [this PR](https://github.com/DataDog/datadog-static-analyzer/pull/953), where we are removing `position_to_byte_offset` and keep the offset logic in `find_secrets`.

Will will make the `ast_filter` available to all flows.

Because we move the AST logic in the `find_secrets` function, some logic (language detection) is moving into the `common` crate.

We will do a follow-up PR later to add support for ast filtering for the IDE plugin.

### DIFF
--- a/crates/bins/src/git_history.rs
+++ b/crates/bins/src/git_history.rs
@@ -183,7 +183,7 @@ fn scan_all_blobs_for_secrets(
                 GIT_HISTORY_BLOB_PLACEHOLDER,
                 &content,
                 options,
-                false,
+                secrets_config.ast_filter,
             );
             if secrets.is_empty() {
                 None
--- a/crates/bins/src/git_history.rs
+++ b/crates/bins/src/git_history.rs
@@ -183,6 +183,7 @@ fn scan_all_blobs_for_secrets(
                 GIT_HISTORY_BLOB_PLACEHOLDER,
                 &content,
                 options,
+                false,
             );
             if secrets.is_empty() {
                 None
--- a/crates/common/src/model/language.rs
+++ b/crates/common/src/model/language.rs
@@ -151,7 +151,7 @@ pub fn get_extensions_for_language(language: &Language) -> Option<Vec<String>> {
     None
 }
 
-// if a langauge only match a file for an exact match, return it
+// if a language only match a file for an exact match, return it
 pub fn get_exact_filename_for_language(language: &Language) -> Option<Vec<String>> {
     for fe in FILE_EXACT_MATCH_PER_LANGUAGE_LIST {
         if fe.0 == *language {
--- a/crates/common/src/model/language.rs
+++ b/crates/common/src/model/language.rs
@@ -1,5 +1,6 @@
 use serde::{Deserialize, Serialize};
 use std::fmt;
+use std::path::Path;
 
 #[derive(Copy, Clone, Deserialize, Debug, Serialize, Eq, Hash, PartialEq)]
 pub enum Language {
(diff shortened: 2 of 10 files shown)


crates/secrets/src/scanner.rs
```diff
@@ -35,14 +38,8 @@ pub fn find_secrets(
     filename: &str,
     code: &str,
     options: &AnalysisOptions,
+    should_filter_using_ast: bool,
```
jasonforal: You are inevitably going to want to change this function to accept a pre-parsed tree, so I would imagine the better function signature would be to accept an `Option<(tree_sitter::Tree, Language)>`, (Some is true, None is false)

I think this is a better design in general, but it'll definitely be needed for IDE support. You mentioned a follow-up PR to implement the IDE. Your call whether you change it now in preparation or churn this code and refactor later

crates/bins/src/git_history.rs
```diff
@@ -183,6 +183,7 @@ fn scan_all_blobs_for_secrets(
                 GIT_HISTORY_BLOB_PLACEHOLDER,
                 &content,
                 options,
+                secrets_config.ast_filter,
```
robertohuertasm-datadog: **blocking**

I did not run a git-history scan. I traced `find_secrets` with this call site. Passing `secrets_config.ast_filter` here does not run the AST filter. It only looks like it does.

`find_secrets` now decides whether to filter with two things at once (`crates/secrets/src/scanner.rs` lines 69 and 92, not in this file): this bool must be `true`, _and_ `get_language_for_file(filename)` must return a language. This call passes `GIT_HISTORY_BLOB_PLACEHOLDER` (`"<git-history-blob>"`, not in this diff, around line 16). That string has no extension, is not an exact `Dockerfile`/`BUILD` match, and does not use the `Dockerfile` prefix, so there is never a language. The match arm that calls `filter_secrets_for_ast` never runs. `is_filtered_by_ast` stays `false` for every git-history match, even when the config flag is on. Pass 2 then clones those results with `clone_with_path` (around line 390) and never filters either.

The placeholder comment still says the scanner only uses the filename for logs. That was true before this PR. Line 69 in `find_secrets` made the filename load-bearing.

Pass 2 already waits for a real path before applying ignore-file checks (`should_ignore_file_for_secret` around line 384, also not in this diff). The AST filter needs the path in the same way: JS/TS vs not comes from the filename, not from the blob bytes. Git stores content and name separately, so the same blob can be `app.js` in one commit (filter applies) and `notes.txt` in another (filter should not). Pass 1 cannot tell those apart. Pass 2 can, and it fans the same `SecretResult` out with `clone_with_path`, so a single flag on the blob would not even be right for both paths.

Would something like this be the right split?

1. Here in pass 1, pass `false`. Update the placeholder comment so it no longer says the filename is logs-only. That keeps git-history matching pre-PR behavior: the HEAD scan still filters, history does not.
2. Optional, not needed in this PR: if git-history should honor `ast_filter` later, that work would live in pass 2, once `path` is known. Derive the language per occurrence, and if the flag is on, run `filter_secrets_for_ast` on those matches (that step needs the blob content again; pass 2 today only has the `SecretResult`). That would be a new feature, fine as a follow-up. It would put history on the same criterion the HEAD scan already uses: JS/TS matches outside a string or comment get flagged. Git-history never did that, including in #953.

In summary, if git-history is meant to stay unfiltered on purpose, (1) alone would be enough. If the experimental flag should mean the same thing for history as it does for HEAD, (2) would be the follow-up. The part that feels off in this PR is passing `true` at a call site that can never filter.

crates/secrets/src/scanner.rs
robertohuertasm-datadog, line 41: **non-blocking**

On [#953](https://github.com/DataDog/datadog-static-analyzer/pull/953#discussion_r3926876050) I was wondering about `ast_filter_language: Option<Language>` on `find_secrets`. The idea was to keep the SDS byte indexes (`start_index` / `end_index`) on the internal match and set `is_filtered_by_ast` in the same place `is_suppressed` is already set (the `#no-dd-secrets` line check, a few lines below in this function). That way we would not convert SDS offsets to `Position` and later invert them back to bytes for the AST. This PR did that part: the indexes are on the internal result, and the filter runs inside `find_secrets`. That is what I was hoping would come out of that earlier note.

The contract around it went a different way: this bool, plus `get_language_for_file(Path::new(filename))` at line 69. That is the same mechanism as the git-history call at `crates/bins/src/git_history.rs` line 186: that site can pass `true` and still never filter, because language is inferred from the filename here. Two ways to skip filtering, and the second one is silent:

- `should_filter_using_ast == false`
- filename does not map to a language, so line 69 returns no language and the match at line 92 never calls `filter_secrets_for_ast`

Git-history pass 1 is that second path (`crates/bins/src/git_history.rs` line 186): the bool can be `true` and nothing is filtered, because the placeholder name is not a language. `crates/bins/src/lib.rs` still calls `get_language_for_file` at line 314 for `is_test_file`, so on the HEAD scan, when this flag is on, detection runs twice on the same path.

Would `Option<Language>` still be simpler? `find_secrets` would not call `get_language_for_file` at all. It would filter only when the caller passes `Some(language)` and that language is in the allow-list. `None` would mean "do not filter", with no second hidden exit. The filename could go back to being mostly for logs, which is what the git-history placeholder comment still claims.

The lookup would move to the callers that already know (or know they do not know) the path:

- HEAD (`crates/bins/src/lib.rs`): it already computes the language for `is_test_file`. It could do that once and pass `Some(lang)` when `ast_filter` is on, or `None` when it is not.
- Git-history pass 1: no path, so it would pass `None`. That would be visible at the call site instead of `true` plus a name that cannot work.
- Server / IDE: they have a real `request.filename`. They could look up the language and pass `Some`, or pass `None` if this path should stay unfiltered.

The bool looks easier because callers only thread `SecretsConfiguration.ast_filter`. The cost is that `find_secrets` has to guess the language from a string that is not always a path, and when it cannot, the flag does not mean what it looks like. `Option<Language>` is one extra argument at each call site, but it makes "can we filter?" a fact the caller already has, not something the scanner infers.

Happy to be told the bool is intentional so secrets does not need a `Language` in its public signature. The silent skip at line 92 is the part that feels easy to miss, and it is what makes the git-history call look wired.

crates/common/src/model/language.rs
```diff
@@ -104,3 +105,232 @@ impl fmt::Display for Language {
...
+        if fe.0 == *language {
+            let extensions = fe.1.to_vec();
+            return Some(extensions.iter().map(|x| x.to_string()).collect());
+        }
+    }
+    None
+}
+
+// filter the file according to a list of extensions
+fn match_extension(path: &Path, extensions: &[String]) -> bool {
```
robertohuertasm-datadog: **non-blocking**

These three matchers (`match_extension`, `match_exact_filename`, `match_prefix_filename`) are still private in `crates/cli/src/file_utils.rs` around lines 160 to 189 (that copy is not in this diff). `filter_files_for_language` still uses the `cli` copies, while `get_language_for_file` uses these.

Would it be worth keeping one pair here (`pub(crate)` so `file_utils` can call them) and deleting the `cli` copies? I could be missing a reason to keep the two independent. The `FILE_*` tables are `pub` with no readers outside this file, so they could stay private at the same time.

crates/common/src/model/language.rs
```diff
@@ -104,3 +105,232 @@ impl fmt::Display for Language {
...
+pub static FILE_EXACT_MATCH_PER_LANGUAGE_LIST: &[(Language, &[&str])] = &[
+    (Language::Dockerfile, &["Dockerfile"]),
+    (Language::Starlark, &["BUILD", "BUILD.bazel"]),
+];
+
+pub static FILE_PREFIX_PER_LANGUAGE_LIST: &[(Language, &[&str])] =
+    &[(Language::Dockerfile, &["Dockerfile"])];
+
+// get all extensions for a language.
+pub fn get_extensions_for_language(language: &Language) -> Option<Vec<String>> {
```
robertohuertasm-datadog: **non-blocking**

The getters copy the static tables on every lookup: `to_vec()` then `map(to_string)` into a `Vec<String>`, and `match_extension` then does `contains(&e.to_string().to_lowercase())`. The tables are already `&[&str]` (`"js"`, `"jsx"`, …).

Would it be simpler to return `&'static [&'static str]` from the getters and compare with `eq_ignore_ascii_case`, so neither side allocates? Same shape for `get_exact_filename_for_language` and `get_prefix_for_language` just below.

crates/common/src/model/language.rs
```diff
@@ -104,3 +105,232 @@ impl fmt::Display for Language {
...
+        Some(p) => match p.to_str() {
+            Some(s) => prefixes_list.iter().any(|p| s.to_string().starts_with(p)),
+            None => false,
+        },
+        None => false,
+    }
+}
+
+/// Find the language for a given file.
+pub fn get_language_for_file(path: &Path) -> Option<Language> {
```
robertohuertasm-datadog, line 209: **non-blocking**

`get_language_for_file` is first-match across languages and excludes `Dockerfile.*.dockerignore`. `filter_files_for_language` in `crates/cli/src/file_utils.rs` (not in this diff) is a per-language OR and does not apply that exclusion, so "is this a Dockerfile?" can disagree depending on which helper you call.

Would it be worth picking one of those policies now that both helpers sit on the same tables in this file?

crates/secrets/src/lib.rs
```diff
@@ -1,3 +1,14 @@
+#[derive(Clone, Debug, PartialEq)]
```
robertohuertasm-datadog, line 1: **nit**

Carrying `start_index` / `end_index` on the internal result is the shape I was hoping for on [#953](https://github.com/DataDog/datadog-static-analyzer/pull/953#discussion_r3926876050). Parking it at crate root, above the `pub mod` declarations, with the `use`s it needs after those modules (lines 20 to 21), feels like the leftover of sharing it between `scanner` and `ast_filter`.

Would it be cleaner as `pub(crate)` in `ast_filter.rs` (scanner already depends on that module) or in a small private module? Happy to be told the crate-root placement was the smallest way to break the cycle.

robertohuertasm-datadog (commented):
@juli1 A few comments for your consideration.

crates/secrets/src/scanner.rs
juli1: Great point and great catch!

robertohuertasm-datadog (approved):
LGTM! 🚀

PR #959, opened 2026-09-10
[K9CODESEC-6268] Adjust how diff aware hash is computed

jdelgo:
The diff aware hash stopped taking into account `.gitignore` and generated files in`0.9.2` and surfaced again in `0.9.3`. This resulted in:
- Changes to `.gitignore` could leave the config hash unchanged
- Changes to generated file handling could leave the config hash unchanged
- A baseline with different file selections could be selected, when we would need a new baseline run

## What is your solution?
- Populate `path_config` for SAST and Secrets with ignored files after we resolve each's config
- Adjust `select_files()` to now use `path_config` instead of computing its own resolved files

## Alternatives considered
- Building a shared path configuration. This overcomplicated the problem and instead we are now just reverting to old behavior
Adding the missing values directly to the digest:
- This would fix the issue but would make it so different implementation for file selection and hashing could drift again. Sharing this path config keeps the two behaviors in line

- This will invalidate config hashes that are using `.gitignore` and generated file exclusions. This is on purpose but will lead to new baseline runs occurring
- This doesn't change which files are selected and just makes the config hash more accurate

## How did this behavior regress?
Before, we had `cli_configuration` which had function [generate_diff_aware_digest()](https://github.com/DataDog/datadog-static-analyzer/blob/501172b2b14f1c1b892e0b735bf63bb08d3547c6/crates/cli/src/model/cli_configuration.rs#L69) that took into account generated and gitignore'd files. This was populated in [datadog-static-analyzer.rs](https://github.com/DataDog/datadog-static-analyzer/blob/501172b2b14f1c1b892e0b735bf63bb08d3547c6/crates/bins/src/bin/datadog-static-analyzer.rs#L401-L418). This means we were populating this `path_config` and then hashing it as we would want.

After, we do something similar in `sast_configuration` where we have [generate_diff_aware_digest()](https://github.com/DataDog/datadog-static-analyzer/blob/d4bfbd086838183298efe166a8bb18a26e817f8c/crates/cli/src/model/sast_configuration.rs#L53) but the issue is our `path_config` is not populated as we want. `path_config` is now populated in [resolve_sast_config()](https://github.com/DataDog/datadog-static-analyzer/blob/main/crates/bins/src/bin/datadog-static-analyzer.rs#L440-L455) but the issue is when it is [called](https://github.com/DataDog/datadog-static-analyzer/blob/d4bfbd086838183298efe166a8bb18a26e817f8c/crates/bins/src/bin/datadog-static-analyzer.rs#L902), we don't pass in any generated or gitignore'd files. That is resolved later in [select_sast_files()](https://github.com/DataDog/datadog-static-analyzer/blob/d4bfbd086838183298efe166a8bb18a26e817f8c/crates/bins/src/bin/datadog-static-analyzer.rs#L967) where all the eligible SAST files are picked. This means we are correctly selecting the diff'd files, but we are just not accounting for that in the diff aware hash since we are not populating `path_config`

## Testing
Ran the analyzer on the same repo with the same config but changed the `.gitignore`
Before changes:
- Baseline gitignore hash (sha: `24e4c7d79507ad0a7f15d04c5654ed34df947eac`)
`338fe5f43fe965aa027c270679786d40b32361c5ad852697e756533720c1bc5a`
- Adjusted gitignore hash (sha: `50ed41bd635d51acc19e9aa1d5ef427412ec9463`)
`338fe5f43fe965aa027c270679786d40b32361c5ad852697e756533720c1bc5a`
**Notice how they are the same hashes**

After changes:
- Baseline gitignore hash (sha: `24e4c7d79507ad0a7f15d04c5654ed34df947eac`)
`8a2087c0d36c1b4bada6c07865c7661be707c710077fa94e94293129175ad51b`
- Adjusted gitignore hash (sha: `50ed41bd635d51acc19e9aa1d5ef427412ec9463`)
`c73e82e1953c8810b68f3501444ee22a4c273ecb09764f7d019bba88ab48ac52`
**Notice how the hashes are now different**


crates/cli/src/model/sast_configuration.rs
```diff
@@ -48,11 +61,13 @@ impl DiffAware for CliConfigurationSast<'_> {
...
 
+        let path_config =
+            effective_path_config(self.gitignore_patterns, &self.sast.file_selection());
+
         let full_config_string = format!(
             "{}:{}:{}:{}:{}:{}:{}",
-            self.sast.path_config.ignore.join(","),
-            self.sast
-                .path_config
+            path_config.ignore.join(","),
```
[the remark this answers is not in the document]
jdelgo: This was pre-existing

jasonforal (commented):
So if I am reading the description correctly, before 0.9.2, we mutated `ignore` on `PathConfig` by pushing the globs from the `.gitignore` into it. But from 0.9.2, we no longer mutate `ignore` -- rather, we apply the filtering logic separately (via the new `select_sast_files(...)`).

Thus, one part of the system (diff aware digest) that assumed that `.gitignore` contents would be in the `ignore` field had its behavior altered.

---

So given that, while your PR works, it seems a bit over-complicated. Could we just have this PR restore the previous behavior? It's plausible that other parts of the system likewise assume `.gitignore` globs will show up in the `ignore`.

Would it work to instead (if the config requires) mutate the `SastConfiguration`'s `PathConfig` to push all the gitignore globs to the `ignore` field (and likewise do the same for `SecretsConfiguration`)?

crates/cli/src/model/sast_configuration.rs
```diff
@@ -208,4 +218,54 @@ mod tests {
 
         assert_ne!(digest_root, digest_subdir);
     }
+
+    #[test]
+    fn diff_aware_hash_depends_on_gitignore_patterns() {
```
jasonforal: These aren't testing `.gitignore` -- they're testing the path config. You'll need to use `tempfile`. Example: https://github.com/DataDog/datadog-static-analyzer/blob/a894622244b57e5e0e3e1351234eb0cbd4b47b5a/crates/cli/src/file_utils.rs#L383-L387

crates/cli/src/model/sast_configuration.rs
```diff
@@ -208,4 +218,54 @@ mod tests {
...
+        let run = run_configuration(vec![]);
+        let mut includes_generated = sast_configuration();
+        includes_generated.ignore_generated_files = false;
+        let mut excludes_generated = includes_generated.clone();
+        excludes_generated.ignore_generated_files = true;
+        crate::file_utils::extend_path_config_ignores(
+            &mut excludes_generated.path_config,
+            &[],
+            true,
+            true,
```
jasonforal: Hardcoding `true` here defeats the purpose of the test, because it makes
```rs
excludes_generated.ignore_generated_files = true;
```

Do nothing. You should be passing that boolean in:
```rs
crate::file_utils::extend_path_config_ignores(/**/, /**/, /**/, excludes_generated.ignore_generated_files)
```

crates/cli/src/model/sast_configuration.rs
```diff
@@ -208,4 +218,54 @@ mod tests {
...
+            run: &run,
+            sast: &second_sast,
+        }
+        .generate_diff_aware_digest();
+
+        assert_ne!(first_digest, second_digest);
+    }
+
+    #[test]
+    fn diff_aware_hash_depends_on_generated_file_setting() {
```
jasonforal: This isn't really testing that the "diff aware hash depends on generated file setting".

It's primarily testing that `extend_path_config_ignores` pushes the default generated file globs to the `PathConfig` when `ignore_generated_files == true`.

crates/cli/src/file_utils.rs
```diff
@@ -652,24 +805,26 @@ mod tests {
         assert_not_contains_files!(&base_path, files, ["src/b/main.rs", "test/a/main.rs"]);
     }
 
-    // `select_files` respects the gitignore patterns it is given unless `ignore_gitignore` is set,
-    // in which case an ignored file is selected anyway.
     #[test]
     fn select_files_honors_ignore_gitignore() {
```
jasonforal: Similar to above, this has nothing to do with `select_files`, because `select_files` only accepts a `PathConfig` -- the only filtering logic it performs is on file size. This test name is misleading

I would recommend to create a test specifically for `extend_path_config_ignores` behavior since this PR is newly introducing this function, and it has important logic.

You will need to re-assess which tests are necessary now (and which need to be renamed) since you are changing the behavior of `select_files`.

Also, now that I'm reading this test, what is the difference between this and the other tests you added? https://github.com/DataDog/datadog-static-analyzer/pull/959/changes#diff-ae5fdd1120210e45907b6f6668a16c7cf7a21629ab1ef29b6a79ee713aba7bf3R221-R223

Would you please make sure we have adequate coverage of the important logic without any duplicate tests?

jasonforal (approved):

PR #963, opened 2026-09-14
Fix missing fingerprint on secrets found in non-UTF-8 files

MikaYuoadas:
## Summary
- Secrets and SAST findings detected in files with invalid UTF-8 byte sequences (e.g. Latin-1 encoded) were not getting a fingerprint, because fingerprint generation re-read the file strictly as UTF-8 while the scanner itself reads files with a lossy UTF-8 decode.
- This caused a violation to be found during scanning but silently dropped in the pipeline downstream.
- Fingerprint generation now reads the file the same lossy way the scanner does, so it's consistent with what was actually scanned.

## Test plan
- [x] Added a unit test covering a violation on a Latin-1 encoded file, asserting a fingerprint is produced
- [x] `cargo test -p cli` passes
- [x] `cargo clippy -p cli --all-targets` clean on the touched code
- [x] `cargo fmt` clean


crates/cli/src/file_utils.rs
```diff
@@ -264,6 +264,7 @@ pub fn get_fingerprint_for_violation(
         }
         return None;
     };
+    let file_contents = String::from_utf8_lossy(&file_bytes);
```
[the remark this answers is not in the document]
MikaYuoadas: Done

crates/cli/src/file_utils.rs
```diff
@@ -377,6 +378,34 @@ mod tests {
...
+        };
+
+        let fingerprint = get_fingerprint_for_violation(
+            "my_rule".to_string(),
+            &violation,
+            dir.path(),
+            Path::new("latin1.txt"),
+            false,
+        );
+        assert!(fingerprint.is_some());
```
robertohuertasm-datadog: **non-blocking**

This would still pass if we hashed the raw Latin-1 bytes, an empty string, or a different replacement character. I only read the test, I did not run a decoder mismatch against it.

Would it be worth asserting equality against `get_fingerprint_from_contents` on the lossy-decoded bytes, so the test locks the scanner's U+FFFD string rather than "some fingerprint exists"? Totally understand if presence was enough for this PR.

```suggestion
        let decoded = String::from_utf8_lossy(b"line one\ncaf\xe9 secret here\nline three\n");
        let expected = get_fingerprint_from_contents(
            "my_rule".to_string(),
            &violation,
            "latin1.txt",
            &decoded,
        );
        assert_eq!(fingerprint, expected);
```

crates/cli/src/sarif/sarif_utils.rs
```diff
@@ -1749,6 +1749,90 @@ mod tests {
...
+        .expect("generate sarif report");
+
+        let emitted_fingerprint = report.runs[0].results.as_ref().unwrap()[0]
+            .partial_fingerprints
+            .as_ref()
+            .expect("partial fingerprints present")
+            .get(SARIF_PROPERTY_DATADOG_FINGERPRINT)
+            .expect("DATADOG_FINGERPRINT present")
+            .to_string();
+        assert!(
```
robertohuertasm-datadog: **non-blocking**

`test_history_only_secret_gets_fingerprint_from_blob` (around line 1746) compares the emitted fingerprint to `get_fingerprint_for_violation` on the same bytes. This sibling drops that check and only asserts non-empty, so a wrong decode would still pass. I only compared the two tests in source, I did not execute a mismatch.

Would it be simpler to assert `emitted_fingerprint` equals `get_fingerprint_from_contents` on `String::from_utf8_lossy(blob_content)`, or even to skip this ~80-line `generate_sarif_report` clone if `get_fingerprint_for_violation_non_utf8_file` in `file_utils.rs` already covers the decode? Happy to be told the SARIF wiring is worth keeping on its own.

```suggestion
        let decoded = String::from_utf8_lossy(blob_content);
        let violation = Violation {
            start: Position { line: 1, col: 1 },
            end: Position { line: 1, col: 22 },
            message: "some secret".to_string(),
            severity: RuleSeverity::Notice,
            category: RuleCategory::Security,
            fixes: vec![],
            taint_flow: None,
            is_suppressed: false,
        };
        let expected = get_fingerprint_from_contents(
            "secret-rule".to_string(),
            &violation,
            relative_path,
            &decoded,
        )
        .expect("fingerprint for lossy-decoded blob");
        assert_eq!(
            emitted_fingerprint, expected,
            "history fingerprint must match the lossy-decoded blob content"
        );
```

crates/cli/src/file_utils.rs
robertohuertasm-datadog: **question**

`get_fingerprint_from_contents` (not in this diff) still takes `&str`, so every caller has to remember `from_utf8_lossy`. That is how the history path in `crates/cli/src/sarif/sarif_utils.rs` line 788 previously used `str::from_utf8(...).ok()?` and dropped fingerprints.

I'm wondering whether we should pass `&[u8]` into `get_fingerprint_from_contents` and decode there, so the SARIF arm cannot reintroduce a strict UTF-8 `?`. Or is keeping the decode at the two call sites intentional, given `read_file` already owns the scan-side policy?

robertohuertasm-datadog (dismissed):
LGTM! I left a few non-blocking comments for your consideration, though.

crates/cli/src/file_utils.rs
MikaYuoadas: That’s a valid point, we should indeed check the value we expect.
I decided to also check against a hardcoded value to avoid a broken `get_fingerprint_from_contents()` to go unnoticed.

crates/cli/src/sarif/sarif_utils.rs
MikaYuoadas: Agreed on fixing that test to actually check the fingerprint instead of just non-empty.

I still think it has value as git-history secrets follow a different path to calculate fingerprint than regular HEAD-scan.
Here this test is more to check that the SARIF generation don’t skip git-history fingerprints silently (it’s easy to do, Doing this PR I first forgot about it and only realized when doing e2e tests with the built binary against my own test repo).

crates/cli/src/file_utils.rs
MikaYuoadas: Indeed I see no reason not to: that’s simpler and the less places to mess up, the better. 👍

robertohuertasm-datadog (approved):
LGTM! 🚀

PR #965, opened 2026-09-14
Support configuration in secret-scanning server requests

juli1:
## Summary
- Add an optional `configuration_base64` field to `SecretScanRequest` (`/scan-secrets`) and `ScanSecretsRequest` (`/ide/v1/secrets/scan`), mirroring `AnalysisRequest`.
- Decode the configuration the same way `process_analysis_request` does, and derive `should_filter_using_ast` from `secrets.experimental-ast-filter` instead of always passing `false` to `secrets::scanner::find_secrets`.
- Apply the same treatment to both the main server endpoint (`process_secret_scan_request`) and the IDE endpoint (`scan`).

## Test plan
- [x] `cargo build --workspace`
- [x] `cargo test -p datadog-static-analyzer -p static-analysis-server`
- [x] New tests cover: no configuration (unchanged behavior), invalid base64/YAML configuration (error), and a configuration with
  ```yaml
  schema-version: v1.6
  secrets:
    experimental-ast-filter: true
  ```
  verifying a secret outside a string literal is flagged (`is_filtered_by_ast: true`) while the same secret inside a string literal is not.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


crates/bins/src/bin/datadog_static_analyzer_server/endpoints.rs
```diff
@@ -177,6 +178,24 @@ fn process_secret_scan_request(
         return Err("Invalid filename: path traversal detected".to_string());
     }
 
+    // Decode the configuration, if present.
+    let configuration = if let Some(config_b64) = request.configuration_base64.clone() {
```
robertohuertasm-datadog, line 182: **non-blocking**

This decode block is the same 16 lines as `ide/secrets/endpoints.rs` around line 61, including the `.clone()` on `configuration_base64` (the owned request is not used afterward) and the same error strings. The two scan pipelines are different (rule JSON vs `SecretRuleApiType`, `disable_validation`), so unifying those feels like the wrong grain.

Would it be worth a single `decode_secrets_configuration(Option<String>) -> Result<Option<ConfigFile>, _>` in `static-analysis-server`? That crate already owns `SecretScanRequest` and the SAST config-decode constants. A `ConfigFile::experimental_ast_filter()` would also save callers from walking `secrets().map(|s| s.experimental_ast_filter).unwrap_or(false)` (the CLI does the same unwrap today). Returning `Option<ConfigFile>` would make a later secrets field easier to honor once, instead of pasting this again. Totally understand if that is out of scope for this PR.

robertohuertasm-datadog (commented):
The Copilot notes on `parse_yaml` vs `parse_any_schema_yaml` are right about the parser: a legacy (schema-version: v1) or schema-less blob will fail here, unlike `/analyze`.

That is fine for this PR. The IDE will not send legacy configs on `/ide/v1/secrets/scan`. We either omit configuration or only send Code Security v1.x. Those legacy files cannot set experimental-ast-filter anyway, so there is nothing to honor. Copilot’s two inline comments can be resolved.

robertohuertasm-datadog (dismissed):
LGTM! 🚀

robertohuertasm-datadog (approved):

